PR #27, opened 2015-12-01
trying all the dockerRequirements options

portah:


portah:
I think it is good to try all the available options for Docker if no dockerImageId then pull but if pull did'n succeed and dockerFile is provided try to create a container.

mr-c:
@portah Thanks for this PR; if it is still needed can you open a new one that's up to date?

PR #29, opened 2016-01-02
Collection of small fixes from bcbio testing

chapmanb:
This collects several small changes run into when testing nested complex
workflows generated by bcbio:

- draft2tool.py -- Correctly handles more cases of files: single returned
  file objects, a None value for a file, and nested lists of files.
- job.py -- Stringify command line arguments to avoid confusing errors
  when a command line contains integers or other non-string bits.
- workflow.py
   - compare nested types explicitly, instead of using `==`. If you add
     inputBindings into nested types they won't be the same with an
     equality test (since one has inputBindings and the other doesn't)
     but actually have the same type.
   - catch an additional case where we should avoid removing a temporary
     directory if `rm_tmpdir` is False.


chapmanb:
I merged this with the draft-3 updates so it should apply cleanly again. Thanks much.

PR #42, opened 2016-02-22
Improving expressiveness of parser --help

smoe:
The patch adds a one-line description as it would appear behind
the name of a yet non-existing man page. This is meant to eventually
help auto-creating it/them.


mr-c:
Thanks, @smoe !

Issue #45, opened 2016-02-24
"No such file or directory" in very simple example


Keyeoh:
Hi everybody,

First of all I would like to congratulate you all for the amazing work you 
are doing with the CWL. It really seems a great idea to try to standardize 
practices which we are actually reimplementing almost continuously in our 
work (I am a bioinformatician). 

I am beginning with the tools, just trying to replicate some of the simple 
examples (Hello world!) and finding out if I can do something similar with 
a command line R script, which is something I am currently very interested 
in. I already have the following dummy R script:

```
#!/usr/bin/env Rscript

args = commandArgs(trailingOnly = TRUE)

result = sum(as.numeric(args))

cat(result, '\n')
```

And I have created the following CWL specification:

```
- id: "#sumcmd"
  class: CommandLineTool
  inputs:
    - id: "#a"
      type: int
      inputBinding: {}
    - id: "#b"
      type: int
      inputBinding: {}
    - id: "#output-file"
      type: string
  outputs:
    - id: "#filename"
      type: File
      outputBinding: 
        glob: $(inputs['output-file'])
  baseCommand: sum.r
  stdout: $(inputs['output-file'])
```

If I try to run the tool, it seems fine at first:

```
vagrant@debian-jessie:~$ cwl-runner sum.cwl#sumcmd -h
/usr/local/bin/cwl-runner 1.0.20160222205901
usage: sum.cwl#sumcmd [-h] -a A -b B --output-file OUTPUT_FILE [job_order]

positional arguments:
  job_order             Job input json file

optional arguments:
  -h, --help            show this help message and exit
  -a A
  -b B
  --output-file OUTPUT_FILE
```

But when I try to run it with actual arguments:

```
vagrant@debian-jessie:~$ cwl-runner sum.cwl#sumcmd -a 50 -b 49 
--output-file foobar.txt
/usr/local/bin/cwl-runner 1.0.20160222205901
Got workflow error
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/cwltool/main.py", line 160, 
in single_job_executor
    for r in jobiter:
  File "/usr/local/lib/python2.7/dist-packages/cwltool/draft2tool.py", line 
131, in job
    j.stdout = builder.do_eval(self.tool["stdout"])
  File "/usr/local/lib/python2.7/dist-packages/cwltool/builder.py", line 164
, in do_eval
    context=context, pull_image=pull_image)
[...]

Issue #39, opened 2016-02-10
Custom nodejs engine?


brainstorm:
Hello CWL folks!

Any reason for shipping a node runtime/engine in the reference implementation?:

https://hub.docker.com/r/commonworkflowlanguage/nodejs-engine/

```
cwltool romanvg$ grep -ri node-engine* *
cwltool/schemas/draft-2/draft-2/bwa-mem-tool.cwl:  - import: node-engine.cwl
cwltool/schemas/draft-2/draft-2/bwa-mem-tool.cwl:      engine: "node-engine.cwl"
cwltool/schemas/draft-2/draft-2/count-lines2-wf.cwl:  - import: node-engine.cwl
cwltool/schemas/draft-2/draft-2/count-lines2-wf.cwl:        engine: node-engine.cwl
cwltool/schemas/draft-2/draft-2/parseInt-tool.cwl:  - import: node-engine.cwl
cwltool/schemas/draft-2/draft-2/parseInt-tool.cwl:  engine: node-engine.cwl
cwltool/schemas/draft-2/draft-2/wc2-tool.cwl:  - import: node-engine.cwl
cwltool/schemas/draft-2/draft-2/wc2-tool.cwl:            engine: node-engine.cwl
cwltool/schemas/draft-2/draft-2/wc3-tool.cwl:  - import: node-engine.cwl
cwltool/schemas/draft-2/draft-2/wc3-tool.cwl:            engine: node-engine.cwl
```

Why not use the upstream one instead (possibly more optimized via alpine linux)?

@chapmanb @mr-c

tetron:
Fixed in 6ce6370

PR #46, opened 2016-02-25
Stop looking for nodejs when found

chapmanb:
A small fix for 6ce637045 to stop searching for nodejs
implementations when the first is found. This prevents an error
when you have a local nodejs but not the node:slim docker image.
(ref #45 and #39).


tetron:
This is one pull request I can confidently review and merge on my phone :-)

PR #47, opened 2016-03-03
Override setting --user option for docker run

agduncan94:
The flag **--no-user** will stop cwltools from passing **--user euid** to docker run and will instead pass the environment variable **HOSTUSER=euid**.

This is to deal with situations where the EUID of the host does not match the EUID of the user in the container (or the EUID of the host does not exist in the container).  Such a situation could cause unexpected issues with running the container. Instead what creators of the containers can do is chown the output directory to the given HOSTUSER environment variable (if needed).  This deals with the assumption that the container can be run as the user with EUID the same as the host and that the user with the given EUID exists.

### DIFF
--- a/cwltool/job.py
+++ b/cwltool/job.py
@@ -69,7 +69,10 @@ class CommandLineJob(object):
             runtime.append("--read-only=true")
             runtime.append("--net=none")
             euid = docker_vm_uid() or os.geteuid()
-            runtime.append("--user=%s" % (euid))
+            if kwargs.get("no_user") is True:
+                runtime.append("--env=HOSTUSER=%s" % (euid))
+            else:
+                runtime.append("--user=%s" % (euid))
 
             if rm_container:
                 runtime.append("--rm")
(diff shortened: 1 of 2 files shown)


denis-yuen:
:+1: @tetron

tetron:
Unfortunately I don't want to add HOSTUSER=euid to the default environment without adding it to the CWL spec, which requires more discussion.  It also requires some cooperation from tool running in the container and I'd prefer to avoid that.

Another possible approach to dealing with file permissions is to use the set group id bit on the output directory.  This should result in the files written to the output directory having the gid of the group of the host users, which means the host user should be able to manipulate the files after the container has finished.  However I haven't tried it yet, but would probably accept an alternate pull request if you want to do it.

mr-c:
Thank you @agduncan94 for your pull request. There has been a lot of changes to the spec and this implementation since March so I'm going to close this for now.  You are very welcome to open an issue to discuss this further.

PR #49, opened 2016-03-07
feat(docker-net): add option to allow using default network

philloooo:
This allows user to opt to enable network for docker container. 
r? @tetron


denis-yuen:
:+1:

PR #60, opened 2016-04-04
Fix tmp prefix usage

alexbarrera:
Modifications to actually use --tmpdir-prefix and --tmp-outdir-prefix paths when cwltool is run with the --no-container flag (addressing issue #3 opened by @schultzmattd). Please note that the execution halt has been removed when tmpdir-prefix or tmp-outdir-prefix do not exist, since prefixes are used to create temporary directories instead of as paths to existing directories.

### DIFF
--- a/cwltool/main.py
+++ b/cwltool/main.py
@@ -539,19 +539,8 @@ def main(args=None,
     if type(t) == int:
         return t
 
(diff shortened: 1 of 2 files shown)


tetron:
I believe this is redundant with #167 and #244

PR #64, opened 2016-04-13
feat(docker-net-option): Allows user to define the docker network

kmhernan:
Instead of `--enable_net` being a boolean, I added the ability to explicitly define the docker network id. If I wanted to use, say, the `host` docker network:

```
cwltool.py --enable-net host ...
```

If you don't use the option is acts just as before.

### DIFF
--- a/cwltool/job.py
+++ b/cwltool/job.py
@@ -67,8 +67,10 @@ class CommandLineJob(object):
             runtime.append("--volume=%s:%s:rw" % (os.path.abspath(self.tmpdir), "/tmp"))
             runtime.append("--workdir=%s" % ("/var/spool/cwl"))
             runtime.append("--read-only=true")
-            if kwargs.get("enable_net") is not True:
+            if kwargs.get("enable_net", None) is None: 
                 runtime.append("--net=none")
+            else:
+                runtime.append("--net={0}".format(kwargs.get("enable_net")))
 
             if self.stdout:
                 runtime.append("--log-driver=none")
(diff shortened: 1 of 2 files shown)


mr-c:
add to testlist

mr-c:
Hey Kyle, it is nice to see you here.

I think you want to move from our current two outcome support of `--net=none` or nothing specified to a three outcome situation: `--net=none`, `--net=custom-thingy`, or nothing specified. However this implementation is lacking support for the default networking option.

Can you add that back in?

kmhernan:
Thanks :-). Hmm.. I think understand what you are saying. But I'm not sure what the best implementation would be? I could keep `--enable-net` working like it has been, and add another option to specify the network name for my case, but then it would probably need to a be a mutually exclusive option? If, I try something like:

```
parser.add_argument("--enable-net", const=True, nargs="?", default=False, help="Blah blah..")
```

You could theoretically use `--enable_net` with no value and it would act as it did before, and if you pass a value it could add my functionality; however, I'm not sure if that plays nicely with the positionals in the argparser?? Am I misunderstanding? Thoughts?

mr-c:
Does https://github.com/common-workflow-language/cwltool/commit/a42ef0604e1957c8c91fde26a2f939444df5009b work for you?

kmhernan:
Awesome! Yup. Thanks

mr-c:
@kmhernan Feel free to pull that commit into this PR. I'd like @tetron to review this before we merge.

kmhernan:
@mr-c I believe I just did. Thanks!

mr-c:
Great, thanks @kmhernan !

kmhernan:
@mr-c it looks like there is one line of code missing [#L607](https://github.com/common-workflow-language/cwltool/blob/master/cwltool/main.py#L607) that does not pass the `custom_net` argument to the kwargs of the executor!

PR #66, opened 2016-04-17
add convenient entry point for cwltool

mr-c:
@tetron I would suggest using this (`cwltool.py`) instead of `cwltool/main.py` to solve the issues Curoverse was having since the move to relative imports.

As a reminder, here is the other workaround:  `setup.py install` or `setup.py develop` and use the `cwltool` that gets install to the appropriate `bin/` directory.

### DIFF
--- a/cwltool.py
+++ b/cwltool.py
@@ -1,0 +1,8 @@
+#!/usr/bin/env python
+"""Convienance entry point for cwltool."""
+
+import sys
+from cwltool import main
+
+if __name__ == "__main__":
+    sys.exit(main.main(sys.argv[1:]))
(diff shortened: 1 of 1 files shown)


prismofeverything:
What was the motivation for relative imports? It breaks being able to run anything unless you are running inside the package. What is the benefit side of this tradeoff?

mr-c:
Hello @prismofeverything,

Relative imports are required for Python 3 compatibility, and the use of any Python 3 based tools, such as the [MyPy static type checker](http://mypy.readthedocs.org/en/latest/introduction.html)

This PR fixes the ability to run `cwltool` without otherwise installing or turning on `develop` mode. I'm going to go ahead and merge this as I see no downsides.

Is there something I missed?

prismofeverything:
Ah, thanks for explaining! I forgot python 3 even existed. Yes, with your addition things run again, thanks : )

mr-c:
Glad to hear everything is working :-) Thank you for reminding me that I needed to merge this.

PR #68, opened 2016-04-26
Avoid input binding of optional arguments when they are not specified.

alexbarrera:
It seems to be an oversight. Currently, this is only a problem when valueFrom is specified and a expression uses _self_, which will be null when doing the eval.

### DIFF
--- a/cwltool/builder.py
+++ b/cwltool/builder.py
@@ -66,7 +66,7 @@ class Builder(object):
 
             if schema["type"] == "record":
                 for f in schema["fields"]:
-                    if f["name"] in datum:
+                    if f["name"] in datum and datum[f["name"]]:
                         bindings.extend(self.bind_input(f, datum[f["name"]], lead_pos=lead_pos, tail_pos=f["name"]))
                     else:
                         datum[f["name"]] = f.get("default")


mr-c:
okay to test

mr-c:
ok to test

mr-c:
Hello @alexbarrera 

Thank you for your pull request.

Can you make a minimal test case demonstrating the problem you are trying to fix? It sounds reasonable to me, but it would be nice to see it in action.

Note that your fix does break [one of the conformance tests](https://github.com/common-workflow-language/common-workflow-language/blob/master/draft-3/draft-3/null-expression1-tool.cwl): 

```
Test [13/61] Test failed: /var/lib/jenkins/jobs/cwltool-pr-conformance/workspace/cwltool-venv/bin/cwltool --outdir=/tmp/tmp3vZuRi --quiet draft-3/null-expression1-tool.cwl draft-3/null-expression1-job.json
Test explicitly passing null to Any type inputs with default values.
output expected {
    "output": 2
}
 got {
    "output": 1
}
Compare failure 2 != 1
```

alexbarrera:
Hi @mr-c 

Here is a simple tool currently breaking:
```
#!/usr/bin/env cwl-runner

cwlVersion: 'cwl:draft-3.dev3'
class: CommandLineTool

requirements:

inputs:
  - id: foo
    type:
      - 'null'
      - File
    description: foo
    inputBinding:
      position: 1
      loadContents: true
      valueFrom: $(self.contents.replace(/\n+/g, ' ').replace(/ $/, ''))
outputs:
  - id: '#bar'
    type: File
    outputBinding:
      glob: bar
stdout: "bar"
baseCommand: 
 - echo
 - baz
```
If executed without the --foo argument, it breaks. 

I would argue that `default` should not be taken into account when not specified in an optional argument (since it is not going to be bind). Still, if you think it should, I can try to add the proper logic.

mr-c:
@alexbarrera I agree, this is an error with `cwltool`.

Here's an updated minimal test case in v1.0 syntax that you can add to the conformance tests:

```
#!/usr/bin/env cwl-runner

cwlVersion: v1.0
class: CommandLineTool

requirements:

inputs:
  foo:
    type: File?
    inputBinding:
      loadContents: true
      valueFrom: $(self.contents.replace(/\n+/g, ' ').replace(/ $/, ''))
outputs:
  bar:
    type: stdout
baseCommand: 
 - echo
 - baz
```

mr-c:
Jenkins, test this please

mr-c:
In the conformance test that is failing, type `Any` combined with a default value means that it will always be evaluated; that is the correct and intended behavior.

tetron:
This requires a spec clarification.  Right now, the `cwltool` behavior is to always evaluate `valueFrom`.  We should update the spec with an errata that clarifies which behavior is correct (always evalute `valueFrom` or only evaluate `valueFrom` on non-null value).

anton-khodak:
This issue was closed in https://github.com/common-workflow-language/cwltool/pull/630. Alex, thank you for your effort!

PR #74, opened 2016-05-04
skip null from items when testing same type

alexbarrera:
This should fix the issue encountered when a step in a workflow feeds optional arguments to a tool. Example:

__Foo tool:__ 
```
#!/usr/bin/env cwl-runner

cwlVersion: 'cwl:draft-3'
class: CommandLineTool

requirements:

inputs:
  - id: foo
    type:
      - 'null'
      - File
    description: foo
outputs:
  - id: '#bar'
    type: File
    outputBinding:
      glob: bar_out
stdout: "bar_out"
baseCommand: 
 - echo
```

__Workflow using Foo tool:__
```
#!/usr/bin/env cwl-runner

class: Workflow
requirements:
  - class: ScatterFeatureRequirement
inputs:
  - id: foo_files
    type:
      type: array
      items: File

outputs:
  - id: '#bar'
    source: "#footool.bar"
    type:
      type: array
      items: File
steps:
  - id: "#footool"
    run: {import: "tool-test.cwl" }
    scatter: "#footool.foo"
    inputs:
      - id: "#footool.foo"
        source: "#foo_files"
    outputs:
      - id: "#footool.bar"
```

Currently, an exception is thrown because the items types are not the same (the sink is an array with _null_ and _File_, while the source is just _File_) :
```
WorkflowException: Type mismatch between source 'file:///Users/abarrera/workspace/cwltool/tests/workflow-test.cwl#foo_files' ({'items': 'File', 'type': 'array'}) and sink 'file:///Users/abarrera/workspace/cwltool/tests/workflow-test.cwl#footool.foo' ({'items': ['null', 'File'], 'type': 'array'})
```

### DIFF
--- a/cwltool/workflow.py
+++ b/cwltool/workflow.py
@@ -94,6 +94,8 @@ def are_same_type(src, sink):  # type: (Any, Any) -> bool
     """
     if isinstance(src, dict) and isinstance(sink, dict):
         if src["type"] == "array" and sink["type"] == "array":
+            if 'null' in sink["items"]:
+                return are_same_type([src["items"]], [it for it in sink["items"] if it != 'null'])
             return are_same_type(src["items"], sink["items"])
         elif src["type"] == sink["type"]:
             return True


mr-c:
okay to test

mr-c:
ok to test

mr-c:
@alexbarrera This is great, thanks! Can you turn your example into a conformance test?

We'll need to see if the standard needs updating as well

alexbarrera:
@mr-c I'm looking into turning this example into a conformance test. Do you mean updating the conformance_test_draft3.yaml file to add this test (with the tool and workflow files) in the schemas/draft-3 folder and creating another PR to merge those changes? I've done that locally and the run_test.sh passes correctly.

mr-c:
@alexbarrera Indeed, except you'd update the source of the schemas folder and make a PR there https://github.com/common-workflow-language/common-workflow-language

@tetron Can I get confirmation from you that this is an acceptable behavior according to the standard?

tetron:
The spec doesn't lay out any type checking rules.  It probably should.  The rule should probably be if the intersection between two type fields (which which my be unions) is nonempty, then it passes static type checking.

tetron:
In other words since the intersection of [File] and [null, File] is [File] then that should pass type checking.

mr-c:
Thank you @alexbarrera , we will clarify the spec as well.

PR #75, opened 2016-05-05
Draft 4 support

tetron:
Refactor tool loading:

* Disentangle command line operations like --print-pre and --print-rdf from the core loading functionality
* Refactor document fetch, document validate, and make tool into separate functions
* For draft-3 and above, validate using the stated schema before applying updates.  This avoids prevents users from writing invalid documents that (for example) declare draft-3 but actually use draft-4 features.
* Support for draft-2 and draft-3.dev versions are now deprecated and will not load by default, can still be enabled with a feature flag --enable-dev


mr-c:
LGTM

PR #76, opened 2016-05-05
Avoid keyword type argument when adding boolean inputs

alexbarrera:
Currently, boolean types do not translate well into argparse arguments. This is very general, but here is an example to test:

```
#!/usr/bin/env cwl-runner
cwlVersion: 'cwl:draft-3'
class: CommandLineTool
inputs:
  - id: bdg
    type: "boolean"
outputs:
  - id: output
    type: File
    outputBinding:
      glob: foo
baseCommand:
  - echo
  - "ff"
stdout: foo
```

### DIFF
--- a/cwltool/main.py
+++ b/cwltool/main.py
@@ -285,8 +285,10 @@ def generate_parser(toolparser, tool, namemap):
             _logger.debug(u"Can't make command line argument from %s", inptype)
             return None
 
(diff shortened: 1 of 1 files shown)


mr-c:
add to testlist

mr-c:
Adapted in #77 ; thanks!

PR #84, opened 2016-05-13
Restore global logger state modified in main.py.

jmchilton:
This allows rerunning of ``main()`` within a single Python application... an application that rhymes with flanemo.


mr-c:
Sure thing

PR #85, opened 2016-05-17
are_same_type return true if intersection of src and sink not empty

alexbarrera:
In line with what @tetron commented in #74 the `are_same_type` function should return `True` if the intersection of `src` and `sink` is not empty. I found out that the solution in #74 was only partial (fixing the problem when the _input_ param is optional, but not when the _output_ param is). This should cover both cases.

### DIFF
--- a/cwltool/utils.py
+++ b/cwltool/utils.py
@@ -16,3 +16,7 @@ def get_feature(self, feature):  # type: (Any, Any) -> Tuple[Any, bool]
         if t["class"] == feature:
             return (t, False)
     return (None, None)
+
+class HashableDict(dict):
+    def __hash__(self):
+        return hash(tuple(sorted(self.items())))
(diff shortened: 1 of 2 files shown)


cnjr2:
I seem to be having the same issue.

```
 Type mismatch between source 'file:///Users/gdebaets/Coding/maker_pipeline/cwl/workflow.cwl#vsearch_fastq_filter/fastqout_discarded' ({'items': ['null', 'File'], 'type': 'array'}) and sink 'file:///Users/gdebaets/Coding/maker_pipeline/cwl/workflow.cwl#fastqout_discarded' (['null', {'items': 'File', 'type': 'array'}])
```

tetron:
Sorry about sitting on this PR for so long.  The solution in this PR isn't quite right, the general case is to determine if there is any intersection of types between a source and sink.  This fix is in PR #106 along with some test cases.  Thanks again for raising this issue.

PR #91, opened 2016-06-02
when secondaryFiles is in an array, pass to each item

jeremiahsavage:


jeremiahsavage:
The fix only works if cwl is draft-2, and causes crash if draft-3. So don't merge.

jeremiahsavage:
Tested successfully with draft-3. I had a bug in cwl test in which I used old style `$self` instead of newer `self`.

jeremiahsavage:
creating new PR which fixes failed check and introduces new test.

mr-c:
New PR is https://github.com/common-workflow-language/cwltool/pull/170

PR #92, opened 2016-06-03
Pathmapper now has list of file objects and not just file names.

tetron:
job.py now stages files so that each file is in a separate directory (along with any
secondary files).  Building block for Directory type and filename munging.
Most but not all conformance tests pass.


tetron:
I'm now using this branch for implementing the Directory type as well.

mr-c:
- [ ] replace non-unicode aware filename validation regex with one that doesn't exclude 80% of the world's population

Issue #101, opened 2016-06-24
CWL enum requires a name field


hmenager:
When specifying an enum parameter for a CWL file, the validation of the type will fail, unless you specify an additional 'name' parameter, required in avro but not in cwl.
a workaround is to specify the name field.

https://gist.github.com/hmenager/5e4ceee602b667cf77bbea86fbb31202

PR #104, opened 2016-06-27
implement enum type in cwltool

hmenager:
this is a fix to the problem described in #101 , with additional modifications to generate the parser correctly if parameters include an enum type.


mr-c:
@hmenager This is great, thanks! I appreciate you also fixing the parser generation; I got caught by that at the GCC hackathon yesterday.

mr-c:
Jenkins, test this please

PR #106, opened 2016-06-28
Fix type checking that a source port is assignable to a sink port

tetron:
Now properly accomodates unions.  Alternate solution for #85


mr-c:
Thanks!

PR #123, opened 2016-07-13
Fix implementation of writable flag

tetron:
Was not correctly propagated from 'entry' to path mapper, and job.py was overwriting the modified output file by accident.


tetron:
See https://github.com/common-workflow-language/common-workflow-language/pull/263

PR #127, opened 2016-07-15
cwltool: Add a compute_checksum optional argument to compute_checksum…

rchippada:
… with default value of True. When False, skip checksum computing which can be expensive.

### DIFF
--- a/cwltool/draft2tool.py
+++ b/cwltool/draft2tool.py
@@ -374,7 +374,7 @@ class CommandLineTool(Process):
         except validate.ValidationException as e:
             raise WorkflowException("Error validating output record, " + str(e) + "\n in " + json.dumps(ret, indent=4))
 
-    def collect_output(self, schema, builder, outdir):
+    def collect_output(self, schema, builder, outdir, compute_checksum=True):
         # type: (Dict[str,Any], Builder, str) -> Union[Dict[unicode, Any], List[Union[Dict[unicode, Any], unicode]]]
         r = []  # type: List[Any]
         if "outputBinding" in schema:
@@ -414,10 +414,12 @@ class CommandLineTool(Process):
                                 files["contents"] = contents
                             filesize = 0
                             while contents != "":
-                                checksum.update(contents)
+                                if compute_checksum:
+                                    checksum.update(contents)
                                 filesize += len(contents)
                                 contents = f.read(1024*1024)
-                        files["checksum"] = "sha1$%s" % checksum.hexdigest()
+                        if compute_checksum:
+                            files["checksum"] = "sha1$%s" % checksum.hexdigest()
                         files["size"] = filesize
                         if "format" in schema:
                             files["format"] = builder.do_eval(schema["format"], context=files)
--- a/cwltool/draft2tool.py
+++ b/cwltool/draft2tool.py
@@ -107,13 +107,10 @@ class CallbackJob(object):
 
     def run(self, **kwargs):
         # type: (**Any) -> None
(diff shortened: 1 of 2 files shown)


mr-c:
Jenkins, okay to test

mr-c:
Thanks @rchippada ! Can you add a command-line option to `cwltool` so users of the reference implementation can also benefit?

rchippada:
@mr-c Michael, added the command line argument and updated references to pass the argument. Please take another look. Thanks.

mr-c:
Thank you @rchippada However the tests are failing: https://ci.commonwl.org/job/cwltool-pr-conformance/128/console

rchippada:
@mr-c Michael, sorry about it. I had mistakenly left a print statement, which seems to have caused the failure. The tests (seem to have) passed with the latest. Please recheck. Thanks.

cwltool/draft2tool.py
```diff
@@ -414,10 +416,12 @@ def collect_output(self, schema, builder, outdir):
                                 files["contents"] = contents
                             filesize = 0
                             while contents != "":
-                                checksum.update(contents)
+                                if compute_checksum:
+                                    checksum.update(contents)
                                 filesize += len(contents)
                                 contents = f.read(1024*1024)
```
tomclegg: If we're not bothering to compute a checksum, this whole "read all the data" loop seems like a really expensive way to determine file size. Is there a more efficient way to get file size?

rchippada:
It appears that we can get file_size using seek when both binding.get("loadContents") and compute_checksum are not set and avoid reading file contents altogether. I will make an update to this effect shortly. Thanks.

rchippada:
@mr-c Hi Michael, I tried to suppress reading the file when compute_checksum is false and use seek to get the file size in that case. However, the tests are failing. Can you please check and let me know what could be going wrong?

I am running unit tests, but do not know how to run the perf tests locally. Please let me know if it is possible to run these tests locally without pushing the branch (and waiting for jenkins to run tests). 

If need be, please roll back to commit:96f088e5d154b064f9341086f74abee297a85538, which was not doing the smart file size handling and tests were passing. Thanks.

rchippada:
@mr-c I reverted back to 96f088e5d154b064f9341086f74abee297a85538 since. It appears that I can get the file size without reading contents using “ f.seek(0, 2) “. However, it appears that the test fail if files[“contents”] do not match. Please suggest how, if at all, we can compute checksum optionally without reading the file contents when needed. Thanks.

rchippada:
With tomclegg's help, learned how to run the conformance tests using common-workflow-language repo. All tests are now passing locally, except one issue. 

After each test run, the file "v1.0/v1.0/whale.txt" is still modified and the subsequent test run is failing. I am manually checking out this file for each test run and now the tests are passing. 

I will push another update to the PR shortly with the latest that also implements the logic to read the file contents only when needed.

rchippada:
@mr-c All conformance tests are passing locally to me at commit:89585b65 

Please check why the tests are failing in cwl ci environment. As I said above, when testing locally, I noticed that the  "v1.0/v1.0/whale.txt" file updated during the test and is not reverted back after the test ends. I am wondering this is what is the case in the ci environment and if this is why the tests are failing. Thanks.

mr-c:
Jenkins, test this please

mr-c:
Jenkins, test this please

mr-c:
Jenkins, test this please

mr-c:
Jenkins, test this please

@rchippada I can confirm that the conformance test run correctly outside of the CI server and I am working on a fix.

Issue #133, opened 2016-07-20
Folder cwl-runner mentioned in README does not exist


peterjc:
The README file currently says to install from source:

```
git clone https://github.com/common-workflow-language/cwltool.git
cd cwltool && python setup.py install
cd cwl-runner && python setup.py install
```

However, (sub)folder ``cwl-runner`` does not exist. Should this say ``cwlref-runner`` instead?

mr-c:
Fixed in #134

PR #134, opened 2016-07-21
Fix README new directory location cwlref-runner

johnbradley:
Looks like a directory name has changed and README instructions for install from source didn't keep up.


mr-c:
Fixes #133 

Thanks!

PR #142, opened 2016-08-02
Copy 3 attributes from array to array items

pvanheus:
As discussed on Gitter with @tetron, for arrays of File objects, the secondaryFiles, format and streamable attributes need to be copied from the array down to the File items that are contained in the array. This change requires the schema change: https://github.com/common-workflow-language/common-workflow-language/pull/281


tetron:
Incorporated this into #211

mr-c:
Thank you @pvanheus , given #211 I'm going to close this PR.

PR #143, opened 2016-08-02
Add check that workflow step 'out' parameter

tetron:
Must match one of the outputs of the underlying process.


tetron:
Jenkins, test this please

PR #144, opened 2016-08-02
Fix command line inputBinding of Directory input parameter.

tetron:


tetron:
Jenkins, test this please

PR #145, opened 2016-08-03
Normalize symlinks in out and temp directories

chapmanb:
When running on systems where temporary directories have symlinks like
(/scratch -> /myfilesystem/scratch) outdir will end up as /scratch/tmp12345
while the actual output files are in /myfilesystem/scratch/tmp12345
resulting in failures in draft2tool
(https://github.com/common-workflow-language/cwltool/blob/e23bf6ae7b7838cf599445830ab3a69648ed8833/cwltool/draft2tool.py#L98). This normalizes these so paths match correctly.


tetron:
Jenkins, test this please

mr-c:
Thanks!

PR #150, opened 2016-08-05
Fix up command line generation, get rid of do_eval and use valueFrom …

tetron:
…correctly.


tetron:
Jenkins, test this please

PR #155, opened 2016-08-10
Refactor fs_access abstraction

tetron:
Pass factory object that takes relative reference basedir instead of actual FsAccess object.


mr-c:
Looks like this will affect Toil and Galaxy; what's the plan?

tetron:
Made some changes that should reduce the chance of backwards compatibility issues with Toil.  Not sure whether/how it will affect Galaxy.  I need this fix merged urgently to address an Arvados bug.

mr-c:
Good enough for now, okay.

PR #156, opened 2016-08-10
Update docker build scripts.

tetron:


mr-c:
According to https://github.com/jpetazzo/dind#a-word-of-warning `wrapdocker` and `dind` is obsolete

undu:
Since #619 and #690 take care of this and have a better approach, let's close this PR.

PR #162, opened 2016-08-16
Fix/docker outdir

jeremiahsavage:
Only check that symlink exists before removing staged link and applying source link.

### DIFF
--- a/cwltool/job.py
+++ b/cwltool/job.py
@@ -244,7 +244,7 @@ class CommandLineJob(object):
                 def linkoutdir(src, tgt):
                     # Need to make the link to the staged file (may be inside
                     # the container)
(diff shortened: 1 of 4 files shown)


thomasyu888:
+1

tetron:
@jeremiahsavage thanks reporting the bug and providing a test case.  For future reference, additions to the test suite should go to the https://github.com/common-workflow-language/common-workflow-language repository, the test suite in `cwltool/schemas` is just a `git subtree` of the primary spec repository.

tetron:
Committed be0484b6d50c8171d778f15080eef90f015489fc to master, and added the corresponding test to the tests in the common-workflow-language spec repository.

PR #164, opened 2016-08-17
when complete, remove tmp-outdirs while retaining cachedirs

jeremiahsavage:
Currently, when a workflow completes, cwltool retains both tmp-outdirs and cachedirs, accumulating storage for each run. This PR creates a set of cachedirs, and passes this set to cleanIntermediate(), so that any tmp-outdir, which is not a cachedir, is removed.


jeremiahsavage:
this approach does not work with subworkflows. Will redesign.

PR #165, opened 2016-08-19
set CommandLineTool outdir to be a subdir of --tmpdir-prefix, if set

jeremiahsavage:
If the user sets --tmpdir-prefix, then have the outdir use that as a prefix.


mr-c:
Jenkins, test this please

tetron:
Fixed in #244

tetron:
Thanks for the contribution, and my apologies for sitting on the bug for so long.

PR #170, opened 2016-08-22
Fix/docker array secondaryfiles

jeremiahsavage:
PR to allow this test to pass
https://github.com/common-workflow-language/common-workflow-language/pull/292

### DIFF
--- a/cwltool/builder.py
+++ b/cwltool/builder.py
@@ -85,10 +85,17 @@ class Builder(object):
                     if binding:
                         b2 = copy.deepcopy(binding)
                         b2["datum"] = item
-                    bindings.extend(
-                        self.bind_input(
-                            {"type": schema["items"], "inputBinding": b2},
-                            item, lead_pos=n, tail_pos=tail_pos))
+                    if "secondaryFiles" in schema:
+                        bindings.extend(
+                            self.bind_input(
+                                {"type": schema["items"], "inputBinding": b2,
+                                 "secondaryFiles" : schema["secondaryFiles"]},
+                                item, lead_pos=n, tail_pos=tail_pos))
+                    else:
+                        bindings.extend(
+                            self.bind_input(
+                                {"type": schema["items"], "inputBinding": b2},
+                                item, lead_pos=n, tail_pos=tail_pos))
                 binding = None
 
             if schema["type"] == "File":


kmhernan:
@jeremiahsavage @mr-c hmm so with more investigation, it seems like for v1.0 at least within the `if schema["type"] == "array"` block there will be no secondary files passed to this would not work. I did get it to work by doing this around line 70:

```        
   elif isinstance(schema["type"], dict):
            st = copy.deepcopy(schema["type"])
            if binding and "inputBinding" not in st and st["type"] == "array" and "itemSeparator" not in binding:
                st["inputBinding"] = {}
            #-- KMH FIX
            if 'secondaryFiles' in schema: st['secondaryFiles'] = schema['secondaryFiles']
            bindings.extend(self.bind_input(st, datum, lead_pos=lead_pos, tail_pos=tail_pos))
```

then back in the block where the commits are placed in this PR, I have:

```
            if schema["type"] == "array":
                for n, item in enumerate(datum):
                    b2 = None
                    if binding:
                        b2 = copy.deepcopy(binding)
                        b2["datum"] = item
                    #-- KMH FIX
                    if 'secondaryFiles' in schema:
                        bindings.extend(
                            self.bind_input(
                                {"type": schema["items"], "inputBinding": b2, "secondaryFiles": schema['secondaryFiles']},
                                item, lead_pos=n, tail_pos=tail_pos))
                    else:
                        bindings.extend(
                            self.bind_input(
                                {"type": schema["items"], "inputBinding": b2},
                                item, lead_pos=n, tail_pos=tail_pos))
                binding = None
```

Now, the `secondaryFiles` will be seen for v1.0. But I don't know if this is a bad approach? I have tested this using the `1.0.20160913171024` build of cwltool.

jeremiahsavage:
@kmhernan I'll try testing your patch for `v1.0`

One problem there might be for `v1.0`, is that the spec removed mention of `secondaryFiles` in `InputArraySchema`, while it was present in the `draft-3` spec.

See `secondaryFile` in `draft-3`:
`5.4.1.1.2 InputArraySchema` http://www.commonwl.org/draft-3/CommandLineTool.html
absent in `v1.0`:
`5.4.2.2 InputArraySchema` http://www.commonwl.org/v1.0/CommandLineTool.html

tetron:
Thanks for looking into this.  I have a fix pending in #211

tetron:
Thanks, I believe this is fixed so I'm closing the PR.

PR #192, opened 2016-09-15
Correctly propagate requirements/hints on Workflow step.

tetron:


tetron:
Jenkins, test this please

PR #194, opened 2016-09-19
Scandeps only picks up file objects when "location" in urlfileds.

tetron:


mr-c:
Can we get a test case?

tetron:
Turns out there was a test case for `scandeps` but it didn't actually assert anything.  Whoops.

PR #207, opened 2016-10-11
Allow the disabling of the network inside a docker container, and add the ability to provide a default docker container if one is not specified

prismofeverything:
This pull request addresses two security issues we have faced when using CWL.

The first is that the `--enable-net` flag never actually worked, in that the network in a docker container was always enabled, even if this flag was not present. It claimed the default behavior was to disable the network, but this was not the case.

In order to not surprise everyone who expects the network to be enabled without passing in a flag, this was changed to `--disable-net`, which disables the network inside a docker container if present.

The other issue was that we want to enforce the use of a docker container (in order to prevent network calls) even if the tool itself doesn't specify one. For this a `--default-container` flag was added that, if present, sets the docker container to run the command in if a docker container is not specified.

Let me know any questions, thanks!

### DIFF
--- a/cwltool/job.py
+++ b/cwltool/job.py
@@ -87,6 +87,9 @@ class CommandLineJob(object):
         if docker_req and kwargs.get("use_container") is not False:
             env = os.environ
             img_id = docker.get_from_requirements(docker_req, docker_is_req, pull_image)
+        elif kwargs.get("default_container", None) is not None:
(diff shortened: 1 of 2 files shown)


mr-c:
FYI, you can specify overriding requirements in the input object document with `cwl:requirements`, this would be another way to ensure all command ran inside a docker container.

PR #208, opened 2016-10-11
Refactor toward allowing extensions required in #93.

jmchilton:


tetron:
@jmchilton can you update this branch?

PR #215, opened 2016-10-28
Add --output-relative-location

tetron:
Specifies that paths in output object should be relative to outdir.


mr-c:
@tetron Is this still relevant?

tetron:
I think I had a use case but clearly I worked around it.  We can close it.

PR #222, opened 2016-11-08
Allow disabling matching the container user to the running user

denis-yuen:
We're running into an issue where cwltool sets the uid to match the running user overriding the `USER` command https://docs.docker.com/engine/reference/builder/#/user

This pull request proposes an option to disable this setting.

This is further documented at https://github.com/ga4gh/dockstore/issues/475

### DIFF
--- a/cwltool/job.py
+++ b/cwltool/job.py
@@ -172,7 +172,9 @@ class CommandLineJob(object):
                 runtime.append("--log-driver=none")
 
             euid = docker_vm_uid() or os.geteuid()
-            runtime.append(u"--user=%s" % (euid))
+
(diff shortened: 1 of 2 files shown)


mr-c:
Thanks @denis-yuen Does this PR fix it for @brainstorm et al. or is the gid sticky bit trick also needed?

brainstorm:
So far it does! I'm running the pipeline from ga4gh/dockstore#475 and it does not seem to crash as immediately as it did before on the cwltool step.

+1 on merging this I would say.

brainstorm:
No sticky bits or other additions needed AFAICT right now... so ++1 for merge, imho.

mr-c:
Thanks, all!

PR #237, opened 2016-11-23
Make package dependencies more specific about versions.

tetron:


setup.py
```diff
@@ -42,17 +42,16 @@
...
-          'setuptools >= 18.5',  # needed by html5lib
-          'requests',
+          'setuptools',
+          'requests>=1.0',
           'ruamel.yaml == 0.12.4',
-          'rdflib >= 4.1.0',
-          'rdflib-jsonld >= 0.3.0',
+          'rdflib >= 4.2.0, < 4.3.0',
+          'rdflib-jsonld == 0.3.0',
+          'html5lib >=0.90, <= 0.9999999',
```
jmchilton: This is what planemo uses - I think it is the better approach if it is attempting to solve the same problem with setuptools.

```
html5lib>=0.9999999,!=0.99999999,!=0.999999999,!=1.0b10
```

PR #260, opened 2016-12-23
Add ability to override logger handler so caller can provide own logger.

tetron:


zymergen-mbarton:
Is there a way to do this with the cwltool Factory? I am looking through the code and it looks like it gets passed in `executor(..., logger=_logger)`. Is there a way to pass to the factory so that it gets passed all the way through?

PR #261, opened 2016-12-23
Add yamlcopy to work around bug in ruamel.ordereddict

tetron:
Failing to preserve node attributes.  See https://bitbucket.org/ruamel/yaml/issues/84/deepcopy-does-not-copy-flow-style


tetron:
May not be necessary, copying bug is fixed in ruamel.yaml 0.13.5, see #262

PR #287, opened 2017-02-08
improve python3 support

gijzelaerr:


mr-c:
"This branch is out-of-date with the base branch" Should I hit the 'update branch' button, or do you want to do that locally?

cwltool/expression.py
```diff
@@ -24,7 +25,7 @@ def jshead(engineConfig, rootvars):
 segment_re = re.compile(segments, flags=re.UNICODE)
 param_re = re.compile(r"\((%s)%s*\)$" % (seg_symbol, segments), flags=re.UNICODE)
 
-JSON = Union[Dict[Any, Any], List[Any], Text, int, long, float, bool, None]
+JSON = Union[Dict[Any, Any], List[Any], Text, int, int, float, bool, None]
```
mr-c: hmm...

cwltool/sandboxjs.py
```diff
@@ -18,7 +17,7 @@ class JavascriptException(Exception):
 
 _logger = logging.getLogger("cwltool")
 
-JSON = Union[Dict[Text, Any], List[Any], Text, int, long, float, bool, None]
+JSON = Union[Dict[Text, Any], List[Any], Text, int, int, float, bool, None]
```
mr-c: Oh, didn't know we duplicated this with https://github.com/common-workflow-language/cwltool/pull/287/files#diff-cb6110fc9c5f825722db0f9c57cb1335R28

cwltool/sandboxjs.py
```diff
@@ -1,12 +1,11 @@
-import cStringIO
```
mr-c: This is needed on line 107

cwltool/expression.py
```diff
@@ -126,7 +127,7 @@ def next_seg(remain, obj):  # type: (Text, Any) -> Any
             try:
                 key = int(m.group(0)[1:-1])
             except ValueError as v:
-                raise WorkflowException(unicode(v))
+                raise WorkflowException(u(v))
```
mr-c:
```
cwltool/expression.py: note: In function "next_seg":
cwltool/expression.py:130: error: Argument 1 to "u" has incompatible type "ValueError"; expected "str"
```

cwltool/draft2tool.py
```diff
@@ -406,7 +407,7 @@ def collect_output_ports(self, ports, builder, outdir, compute_checksum=True):
                                 % shortname(port["id"]), exc_info=True)
                             raise WorkflowException(
                                 u"Error collecting output for parameter '%s':\n%s"
-                                % (shortname(port["id"]), indent(unicode(e))))
+                                % (shortname(port["id"]), indent(u(e))))
```
mr-c:
```
cwltool/draft2tool.py: note: In member "collect_output_ports" of class "CommandLineTool":
cwltool/draft2tool.py:410: error: Argument 1 to "u" has incompatible type "Exception"; expected "str"
```

cwltool/sandboxjs.py
gijzelaerr: ah yes. clearly this is not covered by the test suite :)

cwltool/expression.py
gijzelaerr: where did you find these errors?

cwltool/sandboxjs.py
gijzelaerr: well, it is not required to run the code, only to validate the types. But there are many more type violations due to missing imports in the code... so what should we do?

cwltool/expression.py
mr-c: `make mypy` and https://travis-ci.org/common-workflow-language/cwltool/jobs/199538581#L185

cwltool/sandboxjs.py
mr-c: The lines I commented on are the only type violations I'm aware of.

gijzelaerr: pycharm finds a lot, but that is mostly things like `typing.List` and `typing.Union`, which are probably imported by default by mypy.

mr-c: Oh, those should be added in, yes.

gijzelaerr: it looks impossible to make that type annotation work properly with Python3, the stringio stuff works quite different now.

cwltool/sandboxjs.py
```diff
@@ -101,9 +100,9 @@ def term():
...
     tm.start()
 
-    stdin_buf = StringIO(json.dumps(fn) + "\n")
-    stdout_buf = StringIO()
-    stderr_buf = StringIO()
+    stdin_buf = BytesIO(json.dumps(fn) + "\n")
+    stdout_buf = BytesIO()
+    stderr_buf = BytesIO()
 
     completed = []  # type: List[Union[cStringIO.InputType, cStringIO.OutputType]]
```
mr-c: I'd try `completed = []  # type: List[BytesIO]` here ...

mr-c:
and add 3.4 & 3.6 to `.travis.yml' to turn those tests on

cwltool/sandboxjs.py
gijzelaerr: yep, but that is python3 only.

mr-c: I updated `schema_salad` and `cwltool` for `mypy` 0.470 (the project was renamed from `mypy-lang`). With that in place `type: List[BytesIO]` works for me. If you'll merge in from master then I think we'll be done with this PR.

PR #293, opened 2017-02-17
Remove ez_setup from conformance test Docker

chapmanb:
Follow up from #292 Thanks to @mr-c and @tetron


mr-c:
Confirmed, thanks!

PR #299, opened 2017-02-22
Hashes in filenames

tetron:
Quote hashes in filenames except in special cases, in order to distinguish between hash used as fragment identifier and as part of filename.


mr-c:
Oh no, I can just imagine how you ran into this!

tetron:
A certain very popular sequencing instrument produces filenames like `h5lmgalxx_2016-12-19T14-02-09 Read #1 Cycle #1.log`.

PR #306, opened 2017-02-27
py3

mr-c:
Continuation of #287


mr-c:
This doesn't add full py3 support, but it incorporates the contributions of @gijzelaerr and @tjelvar-olsson which get us closer.

An outline of what is left to do is at https://github.com/common-workflow-language/cwltool/issues/310

PR #311, opened 2017-02-28
Optional directory listing

tetron:
Change handling of directory objects so "listing" isn't assumed.

This is an implementation of https://github.com/common-workflow-language/common-workflow-language/issues/348#issuecomment-266808096


tetron:
Jenkins, test this please

cwltool/process.py
```diff
@@ -44,7 +45,8 @@
                                 "ShellCommandRequirement",
                                 "StepInputExpressionRequirement",
                                 "ResourceRequirement",
-                                "InitialWorkDirRequirement"]
+                                "InitialWorkDirRequirement",
+                                "LoadListingRequirement"]
```
mr-c: Can we have a test for using `LoadListingRequirement` without passing the `--enable-ext` flag? Should behave the same way `cwltool` does w/o this PR

cwltool/pathmapper.py
```diff
@@ -106,6 +106,39 @@ def mark(d):
...
+                get_listing(fs_access, ent, recursive)
+            listing.append(ent)
+        else:
+            listing.append({"class": "File", "location": ld, "basename": bn})
+    rec["listing"] = listing
+
+def trim_listing(obj):
+    """Remove 'listing' field from Directory objects that are file references.
+
+    When Directory objects represent Keep references, it redundant and
```
mr-c: Can you remove the reference to `Keep`?

cwltool/process.py
tetron: Added test_ext.TestListing which tests the various cases.

tetron:
This is blocked on https://github.com/common-workflow-language/common-workflow-language/pull/400

PR #319, opened 2017-03-03
Showing usage message when no arguments are passed 

kapilkd13:
When we run cwltool without any argument we get a "try -h" message. Instead of that directly show Usage message.

### DIFF
--- a/cwltool/main.py
+++ b/cwltool/main.py
@@ -432,6 +432,7 @@ def load_job_order(args, t, stdin, print_input_deps=False, relative_deps=False,
         if toolparser:
             if args.tool_help:
                 toolparser.print_help()
+                print("this is print help")
                 return 0
             cmd_line = vars(toolparser.parse_args(args.job_order))
             for record_name in records:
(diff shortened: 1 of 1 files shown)


cwltool/main.py
```diff
@@ -630,7 +631,8 @@ def main(argsl=None,  # type: List[str]
                 setattr(args, "workflow", "CWLFile")
             else:
                 _logger.error("")
-                _logger.error("CWL document required, try --help for details")
+                _logger.error("CWL document required, No input file was passed\n")
```
mr-c: Is the `\n` really needed?

I would say "CWL document required, no input file was provided"

mr-c:
Jenkins, okay to test

cwltool/main.py
kapilkd13: Not really, should I remove it

cwltool/main.py
```diff
@@ -432,6 +432,7 @@ def load_job_order(args, t, stdin, print_input_deps=False, relative_deps=False,
         if toolparser:
             if args.tool_help:
                 toolparser.print_help()
+
```
mr-c: `cwltool/main.py:435:1: W293 blank line contains whitespace`
from https://travis-ci.org/common-workflow-language/cwltool/jobs/207374774#L164

kapilkd13:
updated message, removed space

mr-c:
Thanks!

PR #321, opened 2017-03-06
10827 static checker

lijiayong:
Add static checker for a workflow before run time.

Modify can_assign_src_to_sink with a "strict" option, which is used to distinguish "pass" from "warning" when comparing types.


mr-c:
Jenkins, okay to test

mr-c:
@lijiayong, thank you so much for your contribution!

I will review this tomorrow, unless another team member gets to this first.

lijiayong:
@mr-c, no problem. Actually I'm working with @tetron and we've been talking about this for a while. I think he's looking at this. Thanks.

cwltool/workflow.py
```diff
@@ -436,7 +479,79 @@ def __init__(self, toolpath_object, **kwargs):
...
+                        if check_types(src["type"], sink["type"], linkMerge, valueFrom) == "warning":
+                            validation["warning"].append(SrcSink(src, sink, linkMerge))
+                        elif check_types(src["type"], sink["type"], linkMerge, valueFrom) == "exception":
+                            validation["exception"].append(SrcSink(src, sink, linkMerge))
+            return validation
+
+        warnings = check_all_types(step_inputs, "source")["warning"] + \
+                    check_all_types(workflow_outputs, "outputSource")["warning"]
+        exceptions = check_all_types(step_inputs, "source")["exception"] + \
+                    check_all_types(workflow_outputs, "outputSource")["exception"]
```
tetron: Please only call `check_all_types(step_inputs, "source")` and  `check_all_types(step_inputs, "source")` once and assign the results to variables, then you can separate the warnings and exceptions.

cwltool/workflow.py
```diff
@@ -436,7 +479,79 @@ def __init__(self, toolpath_object, **kwargs):
...
+            for sink in sinks:
+                if sourceField in sink:
+                    valueFrom = sink.get("valueFrom")
+                    if isinstance(sink[sourceField], list):
+                        srcs_of_sink = [src_dict[parm_id] for parm_id in sink[sourceField]]
+                        linkMerge = sink.get("linkMerge", "merge_nested")
+                    else:
+                        parm_id = sink[sourceField]
+                        srcs_of_sink = [src_dict[parm_id]]
+                        linkMerge = sink.get("linkMerge")
```
tetron: If `source` is not a list with > 1 element, `linkMerge` has no effect.

cwltool/workflow.py
```diff
@@ -436,7 +479,79 @@ def __init__(self, toolpath_object, **kwargs):
...
+                    for src in srcs_of_sink:
+                        if check_types(src["type"], sink["type"], linkMerge, valueFrom) == "warning":
+                            validation["warning"].append(SrcSink(src, sink, linkMerge))
+                        elif check_types(src["type"], sink["type"], linkMerge, valueFrom) == "exception":
+                            validation["exception"].append(SrcSink(src, sink, linkMerge))
```
tetron: Don't call `check_types` twice, assign the return value to a variable and then determine if it is a warning or exception.

cwltool/workflow.py
```diff
@@ -93,31 +94,73 @@ def match_types(sinktype, src, iid, inputobj, linkMerge, valueFrom):
...
+
     if sink == "Any":
         return True
     if isinstance(src, dict) and isinstance(sink, dict):
         if src["type"] == "array" and sink["type"] == "array":
-            return can_assign_src_to_sink(src["items"], sink["items"])
+            return can_assign_src_to_sink(src["items"], sink["items"], strict)
         elif src["type"] == "record" and sink["type"] == "record":
-            return _compare_records(src, sink)
+            return _compare_records(src, sink, strict)
```
tetron: Because you removed "return False" at the bottom of the function, you need to add it here, otherwise if src["type"] != sink["type"] it'll fall through and return None instead of False.

cwltool/workflow.py
```diff
@@ -436,7 +479,79 @@ def __init__(self, toolpath_object, **kwargs):
...
+                    check_all_types(workflow_outputs, "outputSource")["warning"]
+        exceptions = check_all_types(step_inputs, "source")["exception"] + \
+                    check_all_types(workflow_outputs, "outputSource")["exception"]
+
+        warning_msgs = []; exception_msgs = []
+        for warning in warnings:
+            src = warning.src; sink = warning.sink; linkMerge = warning.linkMerge
+            msg = ("Warning: potential type mismatch between source '%s' (%s) and "
+                    "sink '%s' (%s)" %
+                    (src["id"], yaml.safe_load(json.dumps(src["type"])),
```
lijiayong: Yes, that's the reason, and json.dumps() works like a charm. Thanks for the tip!

cwltool/workflow.py
lijiayong: Good call.

cwltool/workflow.py
lijiayong: Good call.

cwltool/workflow.py
lijiayong: I've taken that into account. In the original code, it's possible to fall through the condition "isinstance(src, list)" and "isinstance(sink, list)". But in the new code, they all have a return value at the end.

cwltool/workflow.py
```diff
@@ -436,7 +479,79 @@ def __init__(self, toolpath_object, **kwargs):
         self.steps = [WorkflowStep(step, n, **kwargs) for n, step in enumerate(self.tool.get("steps", []))]
         random.shuffle(self.steps)
 
-        # TODO: statically validate data links instead of doing it at runtime.
+        # statically validate data links instead of doing it at runtime.
```
lijiayong: Sounds good.

cwltool/workflow.py
lijiayong: Sounds good.

cwltool/workflow.py
```diff
@@ -436,7 +479,79 @@ def __init__(self, toolpath_object, **kwargs):
...
+
+        # make a dictionary of source parameters, indexed by the "id" field
+        src_parms = workflow_inputs + step_outputs
+        src_dict = {}
+        for parm in src_parms:
+            src_dict[parm["id"]] = parm
+
+        SrcSink = namedtuple("SrcSink", ["src", "sink", "linkMerge"])
+
+        def check_all_types(sinks, sourceField):
```
lijiayong: Would you clarify this suggestion? I can make this a function under the class Workflow, but I don't think it makes sense for it to be a method of a Workflow object, like check_all_types(self, sinks, sourceField).

lijiayong:
@tetron, please review the new changes.

cwltool/workflow.py
tetron: Still missing a return False on the `if isinstance(src, dict) and isinstance(sink, dict):` clause?

cwltool/workflow.py
```diff
@@ -459,6 +511,84 @@ def visit(self, op):
...
+                (src["id"], json.dumps(src["type"]),
+                sink["id"], json.dumps(sink["type"]))
+                )
+        if linkMerge:
+            msg += ", with source linkMerge method being %s" % linkMerge
+        exception_msgs.append(msg)
+    all_warning_msg = "\n".join(warning_msgs); all_exception_msg = "\n".join(exception_msgs)
+
+    if warnings:
+        print all_warning_msg
```
tetron: Should be `_logger.warn()` not print

cwltool/workflow.py
```diff
@@ -459,6 +511,84 @@ def visit(self, op):
...
+    step_inputs_val = check_all_types(src_dict, step_inputs, "source")
+    workflow_outputs_val = check_all_types(src_dict, workflow_outputs, "outputSource")
+
+    warnings = step_inputs_val["warning"] + workflow_outputs_val["warning"]
+    exceptions = step_inputs_val["exception"] + workflow_outputs_val["exception"]
+
+    warning_msgs = []; exception_msgs = []
+    for warning in warnings:
+        src = warning.src; sink = warning.sink; linkMerge = warning.linkMerge
+        msg = ("Warning: potential type mismatch between source '%s' (%s) and "
```
tetron: You want to use `SourceLine(src).makeError()` and SourceLine(sink).makeError().  Also use `shortname(src["id"])`.

cwltool/workflow.py
```diff
@@ -459,6 +511,84 @@ def visit(self, op):
...
+        if linkMerge:
+            msg += ", with source linkMerge method being %s" % linkMerge
+        warning_msgs.append(msg)
+    for exception in exceptions:
+        src = exception.src; sink = exception.sink; linkMerge = exception.linkMerge
+        msg = ("Type mismatch between source '%s' (%s) and "
+                "sink '%s' (%s)" %
+                (src["id"], json.dumps(src["type"]),
+                sink["id"], json.dumps(sink["type"]))
+                )
```
tetron: Same comment as above.

tetron:
Need to fix linting & typechecking errors as well.

cwltool/workflow.py
lijiayong: Under that clause is a recursive call, which amounts to one of the following three cases.

The correctness of this function is tested under `test_typecomparestrict` of test_examples.py.

cwltool/workflow.py
lijiayong: Cool.

cwltool/workflow.py
lijiayong: Under that clause is a recursive call, which amounts to one of the following three cases.

The correctness of this function is tested under `test_typecomparestrict` of test_examples.py.

cwltool/workflow.py
lijiayong: Could you be more explicit about the use of `SourceLine(src).makeError()` please? Where should I place it?

tetron:
```
msg = SourceLine(src).makeError(
    "Warning: potential type mismatch between source '%s' (%s) and sink '%s' (%s)" %
                (shortname(src["id"]), json.dumps(src["type"]),
                shortname(sink["id"]), json.dumps(sink["type"]))

File: cwltool/workflow.py (hunk 4 above)
Comment (tetron) [inline cwltool/workflow.py]: What happens in this case:

```
src["type"] = "array"
sink["type"] = "record"
```

Comment (lijiayong) [inline cwltool/workflow.py]: It goes to the `else` clause and execute `return src == sink`, which is False. That's what we expected, isn't it? I mean the old code would return the same result.

File: cwltool/workflow.py (hunk 9 above)
Comment (lijiayong) [inline cwltool/workflow.py]: How do we include `SourceLine(sink).makeError()` in it as well? Maybe `SourceLine(src).makeError(SourceLine(sink).makeError(...))`?

Comment (tetron) [inline cwltool/workflow.py]: Yea, actually we probably want to split the error message across two lines.

```
msg = SourceLine(src).makeError("Warning: source %s with type %s may be incompatible"  % (...)) + "\n" +
SourceLine(sink).makeError("with sink %s with type %s"  % (...))"
```

File: cwltool/workflow.py (hunk 4 above)
Comment (tetron) [inline cwltool/workflow.py]: That's not how it works:

```
if True:
    if False:
        print "one"
else:
    print "two"
```

This won't print anything, because the `else` clause is attached to the first `if`, not the second one.  If have an if-elif-else chain, if it goes into any of the branches of the chain, it exits the entire block.

cwltool/workflow.py
lijiayong: Got it.

cwltool/workflow.py
lijiayong: I see, sorry that I missed your point earlier. I'll update the code and add that to a test case. Thanks so much!

lijiayong:
@tetron travis passed, but jenkins failed for unknown reason.

See the following output of static checker on tests/checker_wf/broken-wf.cwl. I've noticed two things:
1. On the third line, the line source 1:1: is wrong, it should be referring to the `#echo_w/echo_in`. All the other line sources are correct. I think this is a bug of CommentedMap.
2.  Note that the second to last line, 'txt' is referring to `#cat/txt`, and it doesn't have line source. This is because step output is stored as dict and not CommentedMap.
<pre>
WARNING:cwltool:
tests/checker_wf/broken-wf.cwl:9:5: Source 'letters0' with type ["string", "int"] may be incompatible
tests/checker_wf/broken-wf.cwl:1:1: with sink 'echo_in' with type ["string", {"type": "array", "items": "string"}]
tests/checker_wf/broken-wf.cwl:15:5: Source 'letters2' with type ["string", "int"] may be incompatible
tests/checker_wf/broken-wf.cwl:43:9: with sink 'echo_in' with type {"items": ["string", {"type": "array", "items": "string"}], "type": "array"}, with source linkMerge method being merge_nested
  File "<stdin>", line 1, in <module>
  File "cwltool/factory.py", line 50, in make
	load = load_tool.load_tool(cwl, self.makeTool)
  File "cwltool/load_tool.py", line 258, in load_tool
	makeTool, kwargs if kwargs else {})
  File "cwltool/load_tool.py", line 231, in make_tool
	tool = makeTool(processobj, **kwargs)
  File "cwltool/workflow.py", line 37, in defaultMakeTool
	return Workflow(toolpath_object, **kwargs)
  File "cwltool/workflow.py", line 491, in __init__
	static_checker(workflow_inputs, workflow_outputs, step_inputs, step_outputs)
  File "cwltool/workflow.py", line 560, in static_checker
	raise validate.ValidationException(all_exception_msg)
schema_salad.validate.ValidationException: 
tests/checker_wf/broken-wf.cwl:21:5: Source 'letters4' with type "int" is incompatible
tests/checker_wf/broken-wf.cwl:51:9: with sink 'echo_in' with type {"items": ["string", {"type": "array", "items": "string"}], "type": "array"}, with source linkMerge method being merge_flattened
tests/checker_wf/broken-wf.cwl:9:5: Source 'letters0' with type ["string", "int"] is incompatible
tests/checker_wf/broken-wf.cwl:65:9: with sink 'cat_in' with type {"type": "array", "items": "File"}, with source linkMerge method being merge_flattened
Source 'txt' with type "File" is incompatible
tests/checker_wf/broken-wf.cwl:29:5: with sink 'all' with type {"type": "array", "items": "File"}
</pre>

tetron:
@lijiayong another thing the static checker should check for: if a step has a required input (i.e. "null" is not a valid input, and it does not have a default or an expression), it needs to have a source, or it is an error.

lijiayong:
cwltool produces the raises the following exception when a workflow input is missing.

> Workflow error, try again with --debug for more information:
> Invalid job input record:
> functional-wf.cwl:8:3: Missing required input parameter `letters0`

@tetron What sort of exception would you like to raise when a step input is missing? Workflow exception or validation exception? May I use the above message as a template?

tetron:
@lijiayong It should be a validation exception.  The message should be something like "required parameter X on step Y does not have source, default or valueFrom expression".

lijiayong:
@tetron I've added the check for missing parameters. The travis-ci failed but I don't think it has to do with my changes.

mr-c:
This is a huge feature, thank you very much @lijiayong @tetron !!!

PR #322, opened 2017-03-07
Setting job from stdin. Bug fix: dict -> CommentedMap

michael-kotliar:
When setting job from stdin `yaml.load` function is supposed to return `CommentedMap`, but returns `dict`. To return `CommentedMap` `RoundTripLoader` should be used.

### DIFF
--- a/cwltool/main.py
+++ b/cwltool/main.py
@@ -414,7 +414,7 @@ def load_job_order(args, t, stdin, print_input_deps=False, relative_deps=False,
     if len(args.job_order) == 1 and args.job_order[0][0] != "-":
         job_order_file = args.job_order[0]
     elif len(args.job_order) == 1 and args.job_order[0] == "-":
-        job_order_object = yaml.round_trip_load(stdin)
+        job_order_object = yaml.load(stdin, yaml.RoundTripLoader)
         job_order_object, _ = loader.resolve_all(job_order_object, file_uri(os.getcwd()) + "/")
     else:
         job_order_file = None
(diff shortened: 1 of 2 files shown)


mr-c:
Jenkins, add to test list

cwltool/main.py
```diff
@@ -414,8 +414,8 @@ def load_job_order(args, t, stdin, print_input_deps=False, relative_deps=False,
     if len(args.job_order) == 1 and args.job_order[0][0] != "-":
         job_order_file = args.job_order[0]
     elif len(args.job_order) == 1 and args.job_order[0] == "-":
-        job_order_object = yaml.load(stdin)
-        job_order_object, _ = loader.resolve_all(job_order_object, "")
+        job_order_object = yaml.load(stdin, yaml.RoundTripLoader)
```
tetron: How about `yaml.round_trip_load()` instead

cwltool/main.py
```diff
@@ -414,8 +414,8 @@ def load_job_order(args, t, stdin, print_input_deps=False, relative_deps=False,
...
+        job_order_object, _ = loader.resolve_all(job_order_object, 'file://%s' % os.path.abspath('.'))
```
tetron: For the second parameter, use: `file_uri(os.getcwd()) + "/"`

cwltool/main.py
michael-kotliar: Done

cwltool/main.py
michael-kotliar: Done

cwltool/main.py
michael-kotliar: Should I change it back to `yaml.load(stdin, yaml.RoundTripLoader)`?
`yaml.round_trip_load()` causes an error in mypy check on Travis

mr-c: @michael-kotliar Yeah, that's fine to switch it back for now.

tetron:
Will merge this in #391

PR #326, opened 2017-03-09
Add extensions.yml to manifest and exclude backup files.

tetron:


tetron:
The annoying this is that I did add it to setup.py package_data, I just didn't realize it needed to be in MANIFEST.in too.  Oof.

mr-c:
@tetron Yeah, guess it is time to have Jenkins run the [release-test.sh](https://github.com/common-workflow-language/cwltool/blob/master/release-test.sh) automatically like [we do for schema_salad](http://ci.commonwl.org/view/All/job/schema_salad_release/).

PR #336, opened 2017-03-16
Adjust behavior for "default"

tetron:
Adjust behavior so thta "default" applies when source value is both " missing and "null".  Conforms to behavior clarifications in upcoming v1.0.1.


tetron:
Incorporated into #485

Issue #268, opened 2017-01-11
nameroot field is not generated for File in input


psafont:
When defining input files only by path the nameroot field does not get generated by the cwl to be consumed by the steps when using the requirement `StepInputExpressionRequirement`.

I made a little repo where it can be easily reproduced: https://gitlab.com/unduthegun/cwl-test

OS: Linux
Python: 2.7.13
cwltool: 1.0.20170105144051

kapilkd13:
Hi, To solve this bug, what I am doing is checking(in steps.input ) if `nameroot` exists, if it doesn't then generate `nameroot` using `basename` (if it exist). But I am not sure where to put this condition.
I am thinking of putting it in https://github.com/common-workflow-language/cwltool/blob/master/cwltool/workflow.py#L148 . Is this approach correct.

kapilkd13:
I guess I was wrong. Better thing to do would be to do the changes in `job_order_object`  while it is being resolved, before it is passed to `workflow`. My changes are here https://github.com/common-workflow-language/cwltool/pull/338

PR #338, opened 2017-03-19
generating nameroot when not provided explicitly from basename Issue#268

kapilkd13:
When a file is provided in input along with its path and nameroot is not provided, cwltool throws exception. In this PR we are generating nameroot if it is not provide explicitly using basename. Issue #268

### DIFF
--- a/cwltool/pathmapper.py
+++ b/cwltool/pathmapper.py
@@ -75,6 +75,9 @@ def normalizeFilesDirs(job):
             parse = urllib.parse.urlparse(d["location"])
             d["basename"] = os.path.basename(urllib.request.url2pathname(parse.path))
 
+        if "nameroot" not in d:
+            d["nameroot"] = ('.').join(d["basename"].split('.')[:-1])  # for basenames like a.b.xyz
+
     adjustFileObjs(job, addLocation)
     adjustDirObjs(job, addLocation)
 


kapilkd13:
Hi, This PR is failing on a unit test. It is generating nameroot field from basename, when a File is specified. But this behavior is expected. After getting your review on it, I would update testcase. Thanks

mr-c:
Thank you @kapilkd13 -- the other place we set the `nameroot` is at https://github.com/common-workflow-language/cwltool/blob/master/cwltool/draft2tool.py#L154

We need to decide if the duplication is okay, or if this change eliminates the need for the other setting of `nameroot`.

Any thoughts, @tetron ?

undu:
When I execute a workflow with a dir, I get this warning, even though I do not define the "nameroot" field

```
jobs/textures-job.yml:2:3: invalid field `nameroot`, expected one of: 'class', 'location', 'path', 'basename', 'listing'
```

But I have to say the workflow that was blocked by #268 now works. 👍

undu:
The current patch is a bit iffy, when the filename doesn't have an extension it's considered that it has no nameroot.

This is a bit better, but still not good:
```
>>> paths = ['lalala', 'lala.la', 'la.la.la', '.lalala']
>>> [path.rsplit('.', 1) for path in paths]
[['lalala'], ['lala', 'la'], ['la.la', 'la'], ['', 'lalala']]
```

kapilkd13:
Closing this PR. See this [PR](https://github.com/common-workflow-language/cwltool/pull/420) by @undu

PR #348, opened 2017-03-24
Add callback to provide custom schema.

tetron:


mr-c:
@tetron Can you share an example `custom_schema_callback()`?

Issue #2, opened 2015-10-16
cwltool uses both default and job_order paths


denis-yuen:
### Long story short:

It appears as though cwltool validates that both the default and job_order paths exist. 
Ideally, it would only validate the job_order paths if provided, the default paths if not provided. 
### Long story:

A toy command and file tree to demonstrate this follows
Required files: https://gist.github.com/denis-yuen/a1d36345ecc4bdc1580a

```
.
├── 8e888694-9c56-4529-a750-d6bfbd4a74e7.txt -> inputs/0bf6ab9f-83a8-4b72-ad67-460eb696bd64/ref_file_2
├── b4cdad91-676a-446c-a635-57453f17617a.txt -> inputs/d2816fb8-7773-4b2a-bd31-ee4a7efa8e04/ref_file_1
├── configs
├── foo.json
├── image-descriptor.cwl
├── inputs
│   ├── 0bf6ab9f-83a8-4b72-ad67-460eb696bd64
│   │   └── ref_file_2
│   ├── 0e129840-fa77-4cec-a885-19b1e538507a
│   │   └── hello_input
│   └── d2816fb8-7773-4b2a-bd31-ee4a7efa8e04
│       └── ref_file_1
├── logs
├── node-engine.cwl
├── outputs
│   └── hello-output.txt
└── working
```

When I run the following command, it executes successfully. ( `cwltool --outdir /home/dyuen/consonance/consonance-arch/./datastore/launcher-03ad4441-c074-4893-b5bd-6d14aabb4995/outputs/ image-descriptor.cwl foo.json` )

However, note the symbolic links in the root. These correspond to the default paths in image-descriptor.cwl which should be overridden by the entries in foo.json.

But when I delete those two links, cwltool dies with the following validation error.

```
/usr/local/bin/cwltool 1.0.20151013173827
Tool definition failed validation:
While checking field `inputs`
  While checking object `file:///home/dyuen/consonance/consonance-arch/datastore/launcher-03ad4441-c074-4893-b5bd-6d14aabb4995/image-descriptor.cwl#ref_file_1`
    While checking field `default`
      Field `path` contains undefined reference to `file:///home/dyuen/consonance/consonance-arch/datastore/launcher-03ad4441-c074-4893-b5bd-6d14aabb4995/8e888694-9c56-4529-a750-d6bfbd4a74e7.txt`
[...]

PR #360, opened 2017-04-10
adding test: give warning instead of error when default path is not present  cwltool Issue #2

kapilkd13:
adding a test for `default path` which if specified but not present at its location should throw a warning instead of error and move forward to check user specified location. Currently this test should give error and would work after https://github.com/common-workflow-language/common-workflow-language/pull/408 . Issue on which it is based is https://github.com/common-workflow-language/cwltool/issues/2


mr-c:
Jenkins, add to testlist

PR #380, opened 2017-05-16
Removed unused stepname in check_adjust 

tetron:


tetron:
Yes, that's a good idea.  That's going to take a little longer, though.

kapilkd13:
Hi, I am trying to create test for `check_adjust`, and I am little confused on how to obtain input to be passed to this function. I have created a base model for unit test here. https://gist.github.com/kapilkd13/70e53c79703611e678f2ab29f7189d5e Please have a look and tell if I am going in right direction. Thanks

mr-c:
@kapilkd13 I think the bigger issue is that there is no test that involved `--cachedir` -- even a simple invocation would have caught this bug

kapilkd13:
You are right. We should actually create a test case  for `--cachedir` and other arguments.

mr-c:
I can confirm this fix, thank you @tetron

PR #386, opened 2017-05-19
No DockerCommandLineJob when use_container=False

chapmanb:
Avoids using a Docker command line when `--no-container` passed.
This avoids creating an invalid Docker command line (with `None` as the
image). I believe this is a regression from here:

https://github.com/common-workflow-language/cwltool/commit/a91c16c343edf9ff7f52e3374ec1dfefe3d23799#diff-29ce315cc581b507c2832e379ce71ad6R175

cc @tetron


mr-c:
Thanks @chapmanb ! Can you include a test to confirm the fix?

mr-c:
P.S. I added a commit to fix the type signature

tetron:
Yea looks like I screwed this up a bit.

The intended behavior is that if the DockerRequirement is a hint, then --no-container won't use it, however if DockerRequirement is an actual requirement, then --no-container would be an error.

chapmanb:
Thanks Peter and Michael. If you have a better way to fix this happy to close and rely on your initial intention, I was only trying to guess. I also had a dig through tests and it wasn't totally clear where one for this should go. If you want to keep this and need a test, any pointers about how to contribute it would be welcome.

PR #387, opened 2017-05-19
Add option to avoid eliminating symbolic links

chapmanb:
Currently the code resolves symbolic links back to the original file,
which can cause issues when files get symlinked with different
extensions. This makes it optional so we can avoid symlinking from Toil.

Specifically, I'm trying to fix the issue with CWL Toil where it copies
all files (BD2KGenomics/toil#1627). Switching to symlinks instead of
copying exposes the issue because files with standard extensions
(`should_work.bam`) are symlinked to internal Toil names
(`tmp12345.tmp`). Resolving the symlinks exposes the internal names to
the program, breaking any code that relies on standard file extensions.

I'm not sure about the origin or motivation for dereferencing (and also
not using os.path.realpath, which would have made it easier to find). It
looks like @tetron has been doing work with this for a while:

common-workflow-language@992e3b1#diff-014f1453ccc1711230a19f9e2893cfacR63

### DIFF
--- a/cwltool/draft2tool.py
+++ b/cwltool/draft2tool.py
@@ -181,7 +181,8 @@ class CommandLineTool(Process):
 
     def makePathMapper(self, reffiles, stagedir, **kwargs):
         # type: (List[Any], Text, **Any) -> PathMapper
-        return PathMapper(reffiles, kwargs["basedir"], stagedir)
+        return PathMapper(reffiles, kwargs["basedir"], stagedir,
+                          eliminate_symlinks=kwargs.get("eliminate_symlinks", True))
 
     def job(self,
             job_order,  # type: Dict[Text, Text]
--- a/cwltool/pathmapper.py
+++ b/cwltool/pathmapper.py
@@ -168,11 +168,13 @@ class PathMapper(object):
 
     """
 
-    def __init__(self, referenced_files, basedir, stagedir, separateDirs=True):
+    def __init__(self, referenced_files, basedir, stagedir, separateDirs=True,
+                 eliminate_symlinks=True):
         # type: (List[Any], Text, Text, bool) -> None
         self._pathmap = {}  # type: Dict[Text, MapperEnt]
         self.stagedir = stagedir
         self.separateDirs = separateDirs
+        self.eliminate_symlinks = eliminate_symlinks
         self.setup(dedup(referenced_files), basedir)
 
     def visitlisting(self, listing, stagedir, basedir, copy=False, staged=False):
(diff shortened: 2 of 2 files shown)


mr-c:
@chapmanb If a tool requires a particular file extension the CWL descriptions should be enforcing them via `InitialWorkDirRequirement`s

chapmanb:
Michael;
How is that done? Is there a way to specify you should always pass files with the original extension? I'm using InitialWorkDirRequirement now to pass in the inputs as a json file, but this resolution of symlinks occurs prior to that. What I'm generalizing here always evaluates symlinks back to the original file right now during the job initialization process.

tetron:
@chapmanb the intended way to preserve filenames is to set the "basename" field explicitly.  Maybe Toil isn't honoring that?

mr-c:
@chapmanb Use IWDR & a dirent, setting 'entryname' to a name that meets the required naming convention: http://www.commonwl.org/v1.0/CommandLineTool.html#Dirent

mr-c:
What @tetron said works, but that is under the user's control and not the CLT describer -- if a particular name or naming convention is required for a tool then that should be modeled in the CWL description; in v1.0 that is done with the InitialWorkDirRequirement

chapmanb:
Peter -- thanks for the pointer. I'll see if I can fix that directly in Toil without needing this change.

Michael -- thanks for the explanation. From a user perspective this seems excessive since the majority of tools will want to honor the original file names. To get that you'd have to double specify everything in the inputs and in InitialWorkDirRequirement? Only the Galaxy team think people want all their files named `tmp12345.txt`. I'd rather have Toil do the right thing here than pushing this requirement on the users.

chapmanb:
Peter -- I've been struggling to resolve this issue without changes to cwltool. I can get Toil to keep references to the non-resolved symlink targets (instead of the original files) but in calls to cwltool it keeps reverting them. Specifically `check_adjust` when running a job converts the `path` and `basename` over to the resolved original file instead of the actual name. Then the inputs to my CWL runs are a mix of non-resolved references (in `location`) and resolved references (in `path` and friends).

So, it ends up as a mess I'm not sure how to get out of. Is the solution I proposed here to be able to avoid resolving symlinks through the API unworkable? What cases does it break? Do you have any other tips/ideas about how best to resolve? Sorry, struggling here trying to get the cwltoil implementation to stop copying files so it can scale and this is my main path forward.

https://github.com/common-workflow-language/cwltool/blob/master/cwltool/draft2tool.py#L151

chapmanb:
Peter -- thanks for the in-person discussion on this. I now have a PR in to Toil (BD2KGenomics/toil#1687) to resolve this without needing cwltool changes, so closing. All of the help and discussion is much appreciated.

PR #393, opened 2017-05-24
feat: prevent python3-pip installation until py3 is supported

manu-chroma:
Closes https://github.com/common-workflow-language/cwltool/issues/392


mr-c:
This is great, thank you @manu-chroma !

PR #402, opened 2017-05-31
Add --print-supported-versions option that prints supported CWL versions

tom-tan:
Currently the only way to check whether the certain version of CWL file is supported by `cwltool` is to run `cwltool` with CWL files and to check its output.

This request adds `--print-supported-versions` that prints supported CWL versions.
We can check the CWL versions supported by `cwltool` by using this option.

Here is the output of `--print-supported-versions` option:
```console
$ ./cwltool.py --print-supported-versions
./cwltool.py 1.0.20170516234254
draft-2
draft-3
draft-3.dev1
draft-3.dev2
draft-3.dev3
draft-3.dev4
draft-3.dev5
draft-4.dev1
draft-4.dev2
draft-4.dev3
v1.0
v1.0.dev4
v1.1.0-dev1
```

### DIFF
--- a/cwltool/main.py
+++ b/cwltool/main.py
@@ -30,7 +30,7 @@ from .process import (shortname, Process, relocateOutputs, cleanIntermediate,
 from .resolver import tool_resolver, ga4gh_tool_registries
 from .stdfsaccess import StdFsAccess
 from .mutation import MutationManager
-from .update import ALLUPDATES
+from .update import UPDATES, ALLUPDATES
 
 _logger = logging.getLogger("cwltool")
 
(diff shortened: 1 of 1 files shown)


mr-c:
Jenkins, okay to test

mr-c:
Thank you @tom-tan for this PR!
FYI, many of those versions require the `--enable-dev` flag -- perhaps it would be best to only print the full list if `--enable-dev` was also passed?

tom-tan:
Thank you for your advice.

I will update this PR to take it into account.

tom-tan:
Done!

I update `supportedCWLversions` to take `enable_dev` flag.
- If `enable_dev` is true, it returns all supported CWL versions (include devel versions).
- Otherwise it returns all draft and released versions without devel versions.

mr-c:
Great, thanks again!

PR #403, opened 2017-05-31
Build docker images with tags if it is released version

tom-tan:
Currently [Docker Hub for cwltool](https://hub.docker.com/r/commonworkflowlanguage/cwltool/tags/) only provides `latest` tag and does not provide other versions like `1.0.20170525215327`, `1.0.20170510165748` and others.
It would be helpful to provide other images with released versions.

This request is to fix `build-cwl-docker.sh` to generate docker images for `cwltool` and `cwltool_module` with tags if the corresponding commit is tagged.

I leave the code to push the tagged images to Docker Hub. To do that, we have to modify job command for Jenkins CI server.

Implementation details:
- I assume that the released tag is like `1.0.20170525215327`. That is, it consists of numbers and dots (`.`) and does not contain other characters.
- I checked whether the commit is tagged release or not by using `git describe --tags` command. I just check the existence of `-` in the output.
  - If it is tagged, it outputs the same as the tag name like `1.0.20170525215327`.
  - Otherwise it outputs like `${latest-tag}-${num-commits}-${hash}` where `${latest-tag}` is the latest tag, `${num-commits}` is the number of commits after `${latest-tag}` and `${hash}` is the hash for current commit.


mr-c:
Jenkins, okay to test

mr-c:
Thank you @tom-tan !

A challenge is that we often don't tag until after this script is triggered. 

But this brings up the point that we don't have a formal release process.

@tetron What do you think about having a separate job on Jenkins that is manually triggered for releases? It would make the wheels, Docker images, tag everything, etc..

tetron:
@mr-c yes, that's a good idea.  Right now I'm doing releases manually (python setup.py sdist && twine upload && add a tag on github) so formalizing with a script would be a good idea.

tom-tan:
I wonder how about using Travis CI to automatically upload pip package and docker image?

If it is automatically tested and packaged by using Travis CI,
it makes easier to see how it works because it will be described in `.travis.yml` and
makes easier to improve the testing and release processes by sending pull requests.

If it is OK for you, I will make another request to automatically build and push a tagged docker image by using Travis CI.

---

Anyway, after merging this request, we can make a tagged docker image by using the following commands:

```console:
$ git tag -a 1.0.20170707 -m "Version 1.0.20170707"    # add tag
$ git push origin 1.0.20170707     # push a tag to github
$ ./build-cwl-docker.sh     # build docker image `commonworkflowlanguage/cwltool:1.0.20170707`
```

tom-tan:
ping.

Should I update `jenkins.bash` to work with Jenkins server?

undu:
Superseded by https://github.com/common-workflow-language/cwltool/pull/690

build-cwl-docker.sh
```diff
@@ -2,3 +2,10 @@
 set -e
 docker build --file=cwltool.Dockerfile --tag=commonworkflowlanguage/cwltool-module --target module .
 docker build --file=cwltool.Dockerfile --tag=commonworkflowlanguage/cwltool .
+
+version=$(git describe --tags)
+echo $version | grep -vq '\-' >& /dev/null
+if [ $? -eq 0 ];then
+    docker tag commonworkflowlanguage/cwltool_module commonworkflowlanguage/cwltool_module:$version
```
mr-c: Here we have `cwltool_module` but above it is `cwltool-module` (underscore vs. dash); this will likely fail

PR #414, opened 2017-06-05
[WIP] Windows Compatibility for Cwltool

kapilkd13:


kapilkd13:
Hi @anton-khodak It is actually my mistake. We will go for an alternative way for symlinks but only for windows OS. I was testing it for windows OS and mistakenly pushed here. I have shifted my work to local branch in my repo and would amend this branch later. Sorry for this mess up.

mr-c:
Replaced by #419

PR #419, opened 2017-06-09
Windows Compatibility: ensuring Cwltool works on windows OS

kapilkd13:
This is the branch I use for testing new changes across all systems when working towards windows compatibility for cwltool. Once when this branch looks good,I would make clean(official) commits on [main windows branch](https://github.com/common-workflow-language/cwltool/pull/414). This branch is for review and suggestions :smile:


kapilkd13:
Hi, I am working on windows compatibility for cwltool. I am getting file not found error in TestFactory's `test_partial_scatter` and `test_partial_output` see here https://ci.appveyor.com/project/mr-c/cwltool/build/.7-master/job/usc0vga9xhaj7e39#L945 There are few requirements in cwl files of these tests like `StepInputExpressionRequirement` and `ScatterFeatureRequirement` see [here](https://github.com/common-workflow-language/cwltool/blob/master/tests/wf/scatterfail.cwl) and [here](https://github.com/common-workflow-language/cwltool/blob/master/tests/wf/wffail.cwl). 
I want to know if either of these testcases depends on symlinks or any unix dependent feature(except unix tools like echo,touch etc). Once I know this I can completely focus on debugging it. Thanks

appveyor.yml
```diff
@@ -0,0 +1,26 @@
...
+  - "python --version"
+  - "python -c \"import struct; print(struct.calcsize('P') * 8)\""
+  - "pip install --disable-pip-version-check --user --upgrade pip"
+
+build_script:
+  - "%CMD_IN_ENV% python setup.py install"
+
+
+test_script:
+  - "%CMD_IN_ENV% python setup.py test"
```
manu-chroma: you can add a command before calling ``setup.py test`` to install requirements from ``pip``. 

Rather than modifying ``setup.py``, replace ``schema_salad`` in ``requirements.txt`` to 
``https://github.com/kapilkd13/schema_salad/archive/windows#egg=schema_salad-2.4.20170612171942``

manu-chroma: this will be enough. you can remove ``dependency_link`` from ``setup.py`` if you use this approach.

kapilkd13: Currently its just for testing, until schema-salad PR get merged. Its working fine right now but if any problem occur I will go this way. I liked the idea of tweaking requirements.txt rather than setup file.

manu-chroma: yeah, I feel the same. just specify version in ``setup.py`` :smile:

tetron:
StepInputExpressionRequirement and ScatterFeatureRequirement do not depend on symlinks or any unix features.  They are internal operations on the workflow graph.

InplaceUpdate is an extension and doesn't need to be supported on Windows.

kapilkd13:
Thanks @tetron It helped 👍  Now we are passing all tests on appveyor. 😺

kapilkd13:
Hi @manu-chroma As you are working on python3 compatibiltiy, I think you can help me with this https://travis-ci.org/common-workflow-language/cwltool/jobs/243363735 travis failing test. It is saying `too few args for lchmod` but lchmod takes 2 args and are provided. Maybe I am missing something. Thanks

manu-chroma:
replace the line with the following 
``os.lchmod(d, mode)  #type: ignore``

this will fix the ``mypy`` error for the time being 😄

kapilkd13:
On testing conformance tests on windows OS. I am getting 
```

E:\GitHub\gsoc\common-workflow-language\v1.0\v1.0>cwltool --debug null-expressio
n1-tool.cwl empty.json
E:\GitHub\gsoc\cwltool\cwltool.py 1.0.20170621183908
Resolved 'null-expression1-tool.cwl' to 'file:///E:/GitHub/gsoc/common-workflow-
language/v1.0/v1.0/null-expression1-tool.cwl'
inside docker pull
3
Pulled Docker image node:slim slim: Pulling from library/node
Digest: sha256:64a2f7356e6336d59a52f580034a2c38262f1852aaa12ed1318208706523f22e
Status: Image is up to date for node:slim

Failed to evaluate expression:
Expression evaluation error:
(10038, 'An operation was attempted on something that is not a socket')
  File "E:\GitHub\gsoc\cwltool\cwltool\draft2tool.py", line 59, in run
    ev = self.builder.do_eval(self.script)
  File "E:\GitHub\gsoc\cwltool\cwltool\builder.py", line 228, in do_eval
    debug=self.debug)
  File "E:\GitHub\gsoc\cwltool\cwltool\expression.py", line 217, in do_eval
    raise WorkflowException("Expression evaluation error:\n%s" % e)
WorkflowException: Expression evaluation error:
(10038, 'An operation was attempted on something that is not a socket')
Removing intermediate output directory c:\users\kapilk~1\appdata\local\temp\tmpx
qoasj
{}
Final process status is permanentFail
```
The reason behind this is https://github.com/common-workflow-language/cwltool/blob/master/cwltool/sandboxjs.py#L149 In this line we are passing `stdin, stderr` to `select.select` method. 
Select's documentation suggests 

> Note File objects on Windows are not acceptable, but sockets are. On Windows, the underlying select() function is provided by the WinSock library, and does not handle file descriptors that don’t originate from WinSock.

Some of the solutions mentioned on web are quite ugly while others modify codebase a lot. 
Also I want ti know what is this code actually doing. I am assuming here we are waiting for I/O events from `stdin,stderr`.
I want to know if anyone has dealt with such problem and if a simple fix exist for it before I try methods mentioned on the web.
Thanks

tetron:
I can tell you the purpose of this code: to safely interact with the external nodejs process without deadlocks.  We need to write the code on stdin and read the feedback from stdout and stderr.  We don't use Popen.communicate() because that waits until the process terminates, whereas this code is reusing a single nodejs process for multiple expression evaluations for performance reasons (starting nodejs each time is very expensive relative to the amount of work being done, especially if it is itself run in a docker container.)

kapilkd13:
Hi @tetron I looked on the web to find solution to our problem. I tried modules: pywin32, msvcrt and gevent. But all of them do not support stdin/stdout for non-blocking read write operation. They expect socket instead of a file descriptor. So finally, based on the information you provided I implemented the read/write operations through threads. Here is the commit https://github.com/common-workflow-language/cwltool/pull/419/commits/69fc0fb9790d020bbd9f079b86a36781ae9f3e87 It's working good on all the tests I have. Would you like to have a look and suggest any changes. Thanks 😃

kapilkd13:
Time for next issue 😄  I am getting `bin/sh not found` on running some conformance tests on windows.
There is `ShellCommandRequirement` in all such tests that try to access shell. 
docs says
> Modify the behavior of CommandLineTool to generate a single string containing a shell command line.
What I understand with this that this allow us to access shell and its tools in our `CommandLineTool`. am I right.
When I run same test in a linux docker it passes.
here is the error 
```
Test [10/87] 
Test failed: cwltool '--outdir=c:\users\appveyor\appdata\local\temp\1\tmpzsr8x0' --quiet v1.0/stderr.cwl v1.0/empty.json
Test command line with stderr redirection
Returned non-zero
'/bin/sh' not found
[job stderr.cwl] completed permanentFail
Final process status is permanentFail
```
So I guess we need to run such tests with `ShellCommandRequirement` to run under docker only. Or is there some other way. 
Thanks

kapilkd13:
when cwltool try to run a command inside a docker container on windows. I got 
```
 docker \
    run \
    -i \
    --volume=/c/users/kapilk~1/appdata/local/temp/tmpjfaqun:/E/var/spool/cwl:rw \
    --volume=/c/users/kapilk~1/appdata/local/temp/tmpv4a3lj:/tmp:rw \
    --volume=/GitHub/gsoc/common-workflow-language/v1.0/v1.0/example_human_Illumina.pe_2.fastq:/E/var/lib/cwl/stgc58606b0-fabf-4bcb-970b-210cd47982db/example_human_Illumina.pe_2.fastq:ro \
    --volume=/E/GitHub/gsoc/common-workflow-language/v1.0/v1.0/args.py:/E/var/lib/cwl/stg2e70dbbe-6467-4241-9c2a-5043fcaa7a53/args.py:ro \
    --volume=/GitHub/gsoc/common-workflow-language/v1.0/v1.0/chr20.fa:/E/var/lib/cwl/stg96452ec4-94e9-4217-8975-4d158ba8d08d/chr20.fa:ro \
    --volume=/GitHub/gsoc/common-workflow-language/v1.0/v1.0/example_human_Illumina.pe_1.fastq:/E/var/lib/cwl/stg70cc2be2-4df6-440d-aa0b-563369047d53/example_human_Illumina.pe_1.fastq:ro \
    --workdir=/E/var/spool/cwl \
    --read-only=true \
    --log-driver=none \
    --user=1000 \
    --rm \
    --env=TMPDIR=/tmp \
    --env=HOME=/E/var/spool/cwl \
    python:2-slim \
    python \
    /E/var/lib/cwl/stg2e70dbbe-6467-4241-9c2a-5043fcaa7a53/args.py \
    bwa \
    mem \
    -t \
    4 \
    -I \
    1,2,3,4 \
    -m \
    3 \
    /E/var/lib/cwl/stg96452ec4-94e9-4217-8975-4d158ba8d08d/chr20.fa \
    /E/var/lib/cwl/stg70cc2be2-4df6-440d-aa0b-563369047d53/example_human_Illumina.pe_1.fastq \
    /E/var/lib/cwl/stgc58606b0-fabf-4bcb-970b-210cd47982db/example_human_Illumina.pe_2.fastq > c:\users\kapilk~1\appdata\local\temp\tmpjfaqun\output.sam
python: can't open file 'E:\var\lib\cwl\stg5a19a8b0-4afd-4ed6-bbdd-9bd546562ce5\args.py': [Errno 2] No such file or directory
unrecognized extension field `http://commonwl.org/cwltool#generation`.  Did you include a $schemas section?
[job bwa-mem-tool.cwl] completed permanentFail

```
Error shows `python cannot find the file`. The reason for this is that the docker run command is run by docker-engine which has a linux base and therefore it accepts unix like path. Also for some reason when mounting volume on docker, docker uses `:` to separate paths. In order to overcome this I am passing volume paths to a function `docker_windows_path_adjust` as you can see paths like `/E/var/lib in the error`. This works fine. But if the command to be executed inside docker also contain windows path(`as in this error message we have python E:\........`) then we get error as as it is a linux based container. We can solve this problem by applying the same function to path passed in command to docker. But the problem is how should we determine which `part/argument` of command passed is a path and should be modified accordingly.
Thanks

mr-c:
@kapilkd13 at least one source of problem is the `python \ /E/var/lib/cwl/stg2e70dbbe-6467-4241-9c2a-5043fcaa7a53/args.py` in the `docker run` command. Should be `python /var/lib/cwl/stg2e70dbbe-6467-4241-9c2a-5043fcaa7a53/args.py` (without the Windows style `/E` at the front)

You should play with docker by hand to see how to get directories and files from Windows mounted inside a Linux container.

cwltool/draft2tool.py
```diff
@@ -133,6 +133,12 @@ def revmap_file(builder, outdir, f):
 
     raise WorkflowException(u"Output File object is missing both `location` and `path` fields: %s" % f)
 
+# On windows os.path.join would use backslash to join path, since we would use these paths in Docker we would convert it to /
+def pathFix(path):  # type: (Text) -> (Text)
```
mr-c: can we have a better name than `pathFix`?

cwltool/job.py
```diff
@@ -275,13 +282,16 @@ def run(self, pull_image=True, rm_container=True,
         if vars_to_preserve is not None:
             for key, value in os.environ.items():
                 if key in vars_to_preserve and key not in env:
-                    env[key] = value
-        env["HOME"] = self.outdir
-        env["TMPDIR"] = self.tmpdir
-
-        stageFiles(self.pathmapper, os.symlink, ignoreWritable=True)
+                    # On Windows, subprocess env can't handle unicode.
```
mr-c: Then lets only call `str` if `onWindows()`

cwltool/job.py
```diff
@@ -303,25 +313,26 @@ def add_volumes(self, pathmapper, runtime, stage_output):
                 containertgt = vol.target
             if vol.type in ("File", "Directory"):
                 if not vol.resolved.startswith("_:"):
-                    runtime.append(u"--volume=%s:%s:ro" % (vol.resolved, containertgt))
+                    runtime.append(u"--volume=%s:%s:ro" % (docker_windows_path_adjust(vol.resolved), docker_windows_path_adjust(containertgt)))
```
mr-c: Please wrap lines longer than 80 characters for easier review :-)

cwltool/load_tool.py
```diff
@@ -48,7 +48,7 @@ def fetch_document(argsworkflow,  # type: Union[Text, Dict[Text, Any]]
     workflowobj = None  # type: CommentedMap
     if isinstance(argsworkflow, string_types):
         split = urllib.parse.urlsplit(argsworkflow)
-        if split.scheme:
+        if split.scheme and split.scheme in [u'http',u'https',u'file']:
```
mr-c: Which scheme's does this code change skip?

cwltool/main.py
```diff
@@ -695,6 +697,18 @@ def main(argsl=None,  # type: List[str]
                 argsl = sys.argv[1:]
             args = arg_parser().parse_args(argsl)
 
+        # If On windows platform, A default Docker Container is Used if not explicitely provided by user
+        if onWindows() and not args.default_container:
+            args.default_container = "ubuntu"
+
+        # Making default Container as lowest priority, that should be used if nothing overrides it
+        if args.default_container:
```
mr-c: Why is this code needed? This behaviour should have been present before this PR.

cwltool/pathmapper.py
```diff
@@ -246,3 +248,9 @@ def reversemap(self, target):  # type: (Text) -> Tuple[Text, Text]
             if v[1] == target:
                 return (k, v[0])
         return None
+
+    # On windows os.path.join would use backslash to join path, since we would use these paths in Docker we would convert it to /
+    def pathFix(self,path):  # type: (Text) -> (Text)
```
mr-c: Why is this a class method? `self` is not used

Should have a different name than https://github.com/common-workflow-language/cwltool/pull/419/files#diff-29ce315cc581b507c2832e379ce71ad6R137

cwltool/process.py
```diff
@@ -205,15 +205,26 @@ def adjustFilesWithSecondary(rec, op, primary=None):
             adjustFilesWithSecondary(d, op, primary)
 
 
-def stageFiles(pm, stageFunc, ignoreWritable=False):
-    # type: (PathMapper, Callable[..., Any], bool) -> None
+def stageFiles(pm, stageFunc=None, ignoreWritable=False, symFunc=True):
+    # type: (PathMapper, Callable[..., Any], bool, bool) -> None
```
mr-c: `symFunc` is type `bool`  but the name implies that it is a function. Would `symlink` be a better name?

cwltool/stdfsaccess.py
```diff
@@ -44,3 +49,9 @@ def join(self, path, *paths):  # type: (Text, *Text) -> Text
 
     def realpath(self, path):  # type: (Text) -> Text
         return os.path.realpath(path)
+
+    # On windows os.path.realpath appends unecessary Drive, here we would avoid that
+    def docker_compatible_realpath(self, path):  # type: (Text) -> Text
+        if onWindows():
+            return path
```
mr-c: Why is `realpath` being called in the first place? Seems dangerous to do nothing if `onWindows()`..

cwltool/utils.py
```diff
@@ -19,9 +24,93 @@ def get_feature(self, feature):  # type: (Any, Any) -> Tuple[Any, bool]
             return (t, True)
     for t in reversed(self.hints):
         if t["class"] == feature:
+            if "requirement" in t and t["requirement"] is True:
+                return(t, True)
```
mr-c: This is a big behavior change. I'm guessing that this is about enforcing priority of the dockerRequirements, but now this method works rather differently than before. Perhaps a rename is called for?

tests/test_examples.py
```diff
@@ -126,9 +126,9 @@ def test_partial_scatter(self):
         try:
             fail()
         except cwltool.factory.WorkflowStatus as e:
-            self.assertEquals('sha1$e5fa44f2b31c1fb553b6021e7360d07d5d91ff5e', e.out["out"][0]["checksum"])
+            self.assertEquals('sha1$356a192b7913b04c54574d18c28d46e6395428ab', e.out["out"][0]["checksum"])
```
mr-c: Why the change in checksums?

tests/test_relax_path_checks.py
```diff
@@ -25,10 +25,12 @@ class ToolArgparse(unittest.TestCase):
 '''
 
     def test_spaces_in_input_files(self):
-        with NamedTemporaryFile(mode='w') as f:
+        with NamedTemporaryFile(mode='w', delete=False) as f:
```
mr-c: why? Not nice to leave temporary files around :-)

kapilkd13: Actually we are opening this file to write and later in the execution reading it  without closing. On linux its possible but on windows it gave errors so I did this.

tests/test_examples.py
kapilkd13: Hmm. Earlier we were writing \n and that gave checksum error So I removed \n and added this new checksum. But now since we have changed all `w` mode to `wb`. It would be interesting to check if it works. I think it should work without this change in checksum. I will check

cwltool/job.py
kapilkd13: 👍  i will take care.

cwltool/load_tool.py
kapilkd13: This is interesting. On windows path Drive letters get considered as scheme. ex `E:\foo` here on urlsplit E becomes scheme. I complained about this to urllib people, but it turns out that is there implementation as use `:` to find scheme

cwltool/main.py
kapilkd13: Actually the previously default container code was placed before job. So it was not propogating to workflow steps. I removed that and changed the position. Strangely can't find that old implementation. I guess some PR changed it. I would see it again.

cwltool/pathmapper.py
kapilkd13: hmm. I would change it. I prefer to move it to utils, as we are using it at one more file. what do you say. Yes a new name is needed. I will do that.

cwltool/process.py
kapilkd13: Yes, symlink sound much better. I would change it 😄

cwltool/stdfsaccess.py
kapilkd13: Actually, on windows Drive letter gets appended. The E drive in docker path, this was the cause of it. One thing We can do is if we are on windows calculate os.path.realpath then trim the drive letter and and again change path separater to `/` as this path goes in docker path. Should I do that

cwltool/job.py
kapilkd13: Okay. we can do that

cwltool/draft2tool.py
kapilkd13: Yes, will change it. better move it to utils.

cwltool/utils.py
kapilkd13: Yes, Originally requirements were compulsory checked while hints were not. But as we have added `default-container` on windows, We are adding it as lowest priority Hints. So by default, it isn't checked. So to enforce checking even if a container is in hints(i.e default container) I added this flag. would `get_feature_and_requirement` work as new name.

cwltool/pathmapper.py
mr-c: :+1: to consolidation & renaming in `utils.py`

cwltool/load_tool.py
mr-c: Cool, can you link to that discussion in the code and add what you said as a comment?

cwltool/main.py
mr-c: Thanks for checking

cwltool/stdfsaccess.py
mr-c: Yes, I think removing this function is best

cwltool/utils.py
mr-c: I think this got fixed in another PR, try it without :-)

cwltool/main.py
```diff
@@ -695,6 +697,18 @@ def main(argsl=None,  # type: List[str]
                 argsl = sys.argv[1:]
             args = arg_parser().parse_args(argsl)
 
+        # If On windows platform, A default Docker Container is Used if not explicitely provided by user
+        if onWindows() and not args.default_container:
+            args.default_container = "ubuntu"
```
mr-c: I thought we were going with a more basic container?

kapilkd13:
Yes, we need to change that. I couldn't decide which one. So for now I
sticked to ubuntu. Some possible image can be alpine or busybox.

mr-c:
Hey, the tests are passing -- woohoo!

cwltool/main.py
kapilkd13: I tested conformance tests on Alpine(size 5mb) docker image. 2 types of test are failing 1. test that use `bin/bash` 2. Test that use `rev` command is giving wrong output(Need to find reason for it)
I found another image https://github.com/frol/docker-alpine-bash `alpine-bash` its size is 6 mb and it comes with bash. So after using this 2 tests are failing that use `rev` reverse command, others are passing. Should we add external containers in tests that use bin/bash or keep `alpine-bash` as default image. And if so what would be a good bash image, `alpine-bash` is light.

kapilkd13:
Yes, All tests are passing including conformance tests on windows. 😄  We can merge this . Currently I am doing some cleanup and checking the default-container code, as a previous pr also changed it.

mr-c:
@kapilkd13 Can you resolve the conflicts?

manu-chroma:
@kapilkd13 It's the python 3 builds which are failing. _The ``Queue`` module has been renamed to ``queue`` in Python 3._ this should help: 
``` python
try:
    import queue #python3
except ImportError:
    import Queue as queue 
```

manu-chroma:
@kapilkd13, since you're using Queues with threads, I think it's better to use https://github.com/common-workflow-language/cwltool/pull/419#issuecomment-315697669 rather than multiprocessing Queues. 

You might find this useful: https://stackoverflow.com/a/925241/1180321

kapilkd13:
Okay. I would do that. Thanks for help help 😄

kapilkd13:
Thanks @manu-chroma  👍

manu-chroma:
@mr-c Would it make sense to do a release now that Python 3 is supported before merging this one?
If not on PyPI, a GitHub tag and release would make sense IMO. Thoughts?

cwltool/job.py
```diff
@@ -379,15 +390,18 @@ def run(self, pull_image=True, rm_container=True,
...
 
-        if kwargs.get("no_match_user", None) is False:
+        if kwargs.get("no_match_user", None) is False and euid is not None:
             runtime.append(u"--user=%s" % (euid))
 
         if rm_container:
             runtime.append(u"--rm")
 
-        runtime.append(u"--env=TMPDIR=/tmp")
+        runtime.append(u"--env=TMPDIR=%s" % self.builder.tmpdir)
```
mr-c: no, the tmpdir inside the docker container shouldn't change

cwltool/job.py
```diff
@@ -436,7 +450,7 @@ def _job_popen(
 
         sp = subprocess.Popen(commands,
                               shell=False,
-                              close_fds=True,
+                              close_fds=False,
```
mr-c: That's a big change :-) How about `close_fds=not onWindows()`?

cwltool/job.py
kapilkd13: Okay. I thought we hardcoded tmp by mistake. I will undo it.

cwltool/main.py
mr-c: Did you test the result of removing this?

mr-c:
Hey @kapilkd13 

From the list below, what versions of Windows + docker have been tested?
```
Operating System :: Microsoft :: Windows :: Windows 10
Operating System :: Microsoft :: Windows :: Windows 3.1 or Earlier
Operating System :: Microsoft :: Windows :: Windows 7
Operating System :: Microsoft :: Windows :: Windows 8
Operating System :: Microsoft :: Windows :: Windows 8.1
Operating System :: Microsoft :: Windows :: Windows 95/98/2000
Operating System :: Microsoft :: Windows :: Windows CE
Operating System :: Microsoft :: Windows :: Windows NT/2000
Operating System :: Microsoft :: Windows :: Windows Server 2003
Operating System :: Microsoft :: Windows :: Windows Server 2008
Operating System :: Microsoft :: Windows :: Windows Vista
Operating System :: Microsoft :: Windows :: Windows XP
```

kapilkd13:
@mr-c  Windows 10+native docker for windows and Windows 8.1+ docker toolbox have been tested. I won't be fully sure, but as it works with Docker Toolbox, I would say it should also support Win 8 and win 7.

mr-c:
@manu-chroma I'm making a separate release first, yep :-) See https://github.com/common-workflow-language/cwltool/pull/472

cwltool/main.py
kapilkd13: Yes, A little strange behaviour was shown. For a tool, docker was called, but our path changes that we do under a condition that docker is being called, didn't took place. My first guess is we extract info about docker requirement from `get_feature` function but since now we are not adding default container as hints this  condition return false. A new find_default_container function has been added in main file by some pr, but it wasn't called. I tried to trace it but couln't find a place from where it can be called(It is passed as arg to maketool). So for now I am sticking with this code.

kapilkd13: One more round of refactoring will be needed for this. I would create a `get_docker_requirement` function that should include result from default_container also if result from hints and requirement is false.

mr-c:
```
mcrusoe@mrcdev:~/cwltool$ git grep find_default_container
cwltool/builder.py:        self.find_default_container = None  # type: Callable[[], Text]
cwltool/draft2tool.py:        self.find_default_container = kwargs.get("find_default_container", None)
cwltool/draft2tool.py:            default_container = self.find_default_container(self)
cwltool/job.py:                find_default_container = self.builder.find_default_container
cwltool/job.py:                default_container = find_default_container and find_default_container()
cwltool/main.py:                     'find_default_container': None,
cwltool/main.py:            make_tool_kwds["find_default_container"] = functools.partial(find_default_container, args)
cwltool/main.py:def find_default_container(args, builder):
```

mr-c: so --> https://github.com/common-workflow-language/cwltool/blob/master/cwltool/job.py#L343

cwltool/main.py
mr-c: Lets add `alpine-bash` to those conformance tests, yep!

kapilkd13:
@mr-c  I think we are ready to go. That container issue is resolved. I removed additional code.

tests/util.py
```diff
@@ -15,3 +17,8 @@ def get_data(filename):
     if not filepath or not os.path.isfile(filepath):
         filepath = os.path.join(os.path.dirname(__file__), os.pardir, filename)
     return filepath
+
+# Check if we are on windows OS
+def onWindows():
```
mr-c: why not use the version in `utils.py`?

kapilkd13: You mean windows platform version for win 8 and 10. something like sys.platform==win32 etc?

mr-c: https://github.com/common-workflow-language/cwltool/pull/419/files#diff-172ce10c584f6a72521c6305db028c9bR107

kapilkd13: Oh right 😄  I thought we put all test utilities in util file inside test folder. Earlier i was using this only. i will switch.

cwltool/main.py
kapilkd13: Ok. One problem that we still have is that test 52 and 53 uses `rev` command. It is not posix compliant so ubuntu and alpine have different implmentation thus it gives wrong output. I thinking of adding a ubuntu container for these tests.

kapilkd13: wrong means different output. It appears `Alpine` and `Busybox` `rev` tool can't handle input larger than certain character and some error occur in their buffer causing output to print multiple times. I checked it manually in a `alpine` docker. We had same behaviour.

mr-c: Interesting! Feel free to propose whichever container is needed for those tests

PR #420, opened 2017-06-09
Generate nameroot and nameext from basename

undu:
Here's the cleaner PR to fix https://github.com/common-workflow-language/cwltool/issues/268
Obsoletes https://github.com/common-workflow-language/cwltool/pull/338

### DIFF
--- a/cwltool/pathmapper.py
+++ b/cwltool/pathmapper.py
@@ -78,7 +78,7 @@ def normalizeFilesDirs(job):
             d["basename"] = os.path.basename(urllib.request.url2pathname(path))
 
         if d["class"] == "File":
-            if "nameroot" not in d:
+            if "nameroot" or "nameext" not in d:
                 split = d["basename"].rsplit(".", 1)
                 if split[0] == "" or len(split) == 1 or split[1] == "":
                     name_root = d["basename"]
@@ -86,6 +86,7 @@ def normalizeFilesDirs(job):
                 else:
                     name_root = split[0]
                     name_ext = split[1]
+
                 d["nameroot"] = name_root
                 d["nameext"] = name_ext
 
--- a/cwltool/pathmapper.py
+++ b/cwltool/pathmapper.py
@@ -77,6 +77,9 @@ def normalizeFilesDirs(job):
         if "basename" not in d:
             d["basename"] = os.path.basename(urllib.request.url2pathname(path))
 
+        if "nameroot" not in d:
+            d["nameroot"] = ('.').join(d["basename"].split('.')[:-1])  # for basenames like a.b.xyz
+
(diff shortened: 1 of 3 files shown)


undu:
I'm unsure on what the behaviour should be when only `nameroot` or `nameext` is set, not both.

Should it generate the value from `basename` disregarding the other one; should it try to play nice with the existing value of the other one; should it overwite the existing value because the algorithm knows what's the 'correct' value for both fields?

undu:
For clarification, right now the code (over)writes both from the calculated values to ensure that there are no inconsistencies between both.

tetron:
@undu yes, that's correct, the `nameroot` and `nameext` fields are supposed to be generated from basename, no the other way.

cwltool/pathmapper.py
```diff
@@ -77,6 +77,19 @@ def addLocation(d):
         if "basename" not in d:
             d["basename"] = os.path.basename(urllib.request.url2pathname(path))
 
+        if d["class"] == "File":
+            if "nameroot" or "nameext" not in d:
```
tetron: You can replace this with the line from draft2tool.py:

```f["nameroot"], f["nameext"] = os.path.splitext(f["basename"])```

File: tests/test_examples.py [hunk 2]
```diff
@@ -209,12 +209,16 @@ def loadref(base, p):
 
         self.assertEquals([{
             "basename": "bar.cwl",
+            "nameroot": "bar",
             "class": "File",
+            "nameext": "cwl",
```
tetron: From the spec:

```The basename root such that nameroot + nameext == basename, and nameext is empty or begins with a period and contains at most one period. For the purposess of path splitting leading periods on the basename are ignored; a basename of .cshrc will have a nameroot of .cshrc.```

So nameext in this example should be `.cwl`

File: cwltool/pathmapper.py (hunk 1 above)
Comment (undu) [inline cwltool/pathmapper.py]: Didn't know there was a function that did it out of the box, it's changed now.

File: tests/test_examples.py (hunk 2 above)
Comment (undu) [inline tests/test_examples.py]: All the tests have been changed to conform to `os.path.splitext`.

File: cwltool/pathmapper.py (hunk 1 above)
Comment (tetron) [inline cwltool/pathmapper.py]: You can also remove the conditional.  As mentioned earlier we can set `nameroot` and `nameext` unconditionally.

FYI that construction doesn't do what you want:

```
if "x" or "y" in "abc":
  print "yes"
```

This will always print "yes" because the expression is actually `bool("x") or bool("y" in "abc")` and non-empty strings are `True`.

undu: Thanks for the tip, I removed the condition.

tetron:
Jenkins, test this please

PR #425, opened 2017-06-14
Use consistent job loader context between load_tool and load_job_order.

tetron:


cwltool/load_tool.py
```diff
@@ -143,6 +150,7 @@ def validate_document(document_loader,  # type: Loader
     jobobj = None
     if "cwl:tool" in workflowobj:
         jobobj, _ = document_loader.resolve_all(workflowobj, uri)
+        print "ZZZZZ", jobobj
```
mr-c: Uhhhh....

tetron: Oh, darnit.  Thanks for noticing that.  Will fix.

PR #438, opened 2017-06-22
Adds --js-console option to facilitate debugging

jrandall:
The `--js-console` option adds a `console.log()`
and `console.error()` function to the javascript context
during expression evaluation (so it should NOT be used
in production since it changes the evaluation context)
and prints the resulting console messages to stdout
prefixed by `[log]` and `[err]` to distinguish between
the two types of console logs.


mr-c:
This is great, thank you @jrandall ! Quick ask, how about a test so we can find out when someone else's PR breaks this in the future?

mr-c:
@jrandall I added a commit to fix https://travis-ci.org/common-workflow-language/cwltool/jobs/246125202#L194

mr-c:
@jrandall 
> cwltool/sandboxjs.py:202: error: Type argument 1 of "findall" has incompatible value "object"
https://travis-ci.org/common-workflow-language/cwltool/jobs/254410930#L305

mr-c:
@kapilkd13 Can you help us out with the Windows builld? I get the feeling that `js_console` isn't being passed correctly

kapilkd13:
@mr-c I am looking at this issue.

kapilkd13:
I tested `js-console` flag on another inlineJS based test, its passing there and 
`Running with support for javascript console in expressions (DO NOT USE IN PRODUCTION)` does appers on the screen. So I guess error is in the test case. Error shown on windows is ``No JSON object could be decoded`. I think its some encoding issue while reading file, but I will confirm.

kapilkd13:
Hi @mr-c  @jrandall  I can verify that `js-console` flag is being passed correctly. Although there are two problems on windows
1. Since in the unit test we are using a call to `main` function, docker is required on windows by default, but docker is not available to us on appveyor. I will change or create a new test case to check this flag on windows without using a Docker call, once below given issue is solved.
2. This unit test is returning us with `No JSON object to decode` error on windows whenever we use a `console.log/error` expression. On further analysis, this issue arises because of quotes in console.log https://github.com/common-workflow-language/cwltool/pull/438/files#diff-328b352a26fe1adb3b70c8664b0de3e6R3 
So if we change `console.error("[log] "+require("util").format.apply(this, arguments).split("\n").join("\n[log] "));` line in `cwlNodeEngineJSConsole.js` with
`
var a=10;
console.log(a);
` 10 get printed to console correctly. So I guess its a problem with quotes. I tried few things like escaping quotes and `json.dumps` the file code in Nodecode variable but didn't reach the solution. I am not much experienced in JS.
Do you guys have some suggestion on this?

mr-c:
Based upon https://ci.appveyor.com/project/mr-c/cwltool/build/.347-master/job/lexnbmb6qfc18thv it looks like this feature isn't running under Docker when `cwltool` is run on Microsoft Windows. @kapilkd13 Is this a correct diagnosis?

ThomasHickman:
Hi @mr-c, I added `--no-container` to the test, as some of the windows tests weren't running in a docker container and I used `["python", "-c", "True"]` as the command (which should run on all platforms)

ThomasHickman:
@kapilkd13 had a bit more of a look at the error, `No JSON object to decode` was appearing as the code in it's current state on windows was only reading one chunk of either stderr or stdout - so the the stdout was never actually getting picked up. I've redone the code to now consider the process completed when both stderr and stdout output a string that indicates the process has completed

kapilkd13:
Hi @mr-c  @jrandall I have tested conformance tests which involves `nodejs` process with this PR on windows and they are working fine. We got error without `--no-container` on appveyor because appveyor do not support docker on windows and by default we are using `alpine` docker image when running cwltool on windows. Good news is that I tested the given unit test with a `python:2-slim` docker image locally and it worked fine. 
Since with the latest changes we are using `python -c .....` command and our default `alpine` docker do not have python installed, we do get `python exec not found in path`  when using default image.
 I am thinking of changing default image from `alpine` to `python:alpine` on windows. I haven't tested it yet but according to docs its still a very small image(5 mb) with minimal python libs. here is a little more info with pros and cons on that https://hub.docker.com/_/python/ 

@jrandall I really liked the idea of sending dummy end string, cool 😄

mr-c:
@jrandall @kapilkd13 This turned out to be quite the work, thank you both!

Shall I merge?

ThomasHickman:
@mr-c, I was taking over this pull request from @jrandall. I just fixed something as reported above, as I noticed that javascript error messages were not getting reported, however this should be now OK to be merged

mr-c:
Pardon me, @thomashickman and thank you for your contributions!

PR #445, opened 2017-06-27
Fix Python 3 incompatibility in PYTHON_RUN_SCRIPT.

jxtx:
`file` is no longer a built-in with the new `io` framework. Was causing jobs to fail depending on the python in the conda env.


mr-c:
Jenkins, add to testlist

mr-c:
Jenkins, add to testlist

PR #450, opened 2017-06-29
setup.py: pin ruamel.yaml package version

manu-chroma:
set upper bound for ruamel.yaml version, 
same as in schema_salad: https://github.com/common-workflow-language/schema_salad/blob/master/setup.py#L33

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ setup(name='cwltool',
       install_requires=[
           'setuptools',
           'requests >= 1.0',
-          'ruamel.yaml >= 0.12.4',
+          'ruamel.yaml >= 0.12.4, < 0.15',
           'rdflib >= 4.2.2, < 4.3.0',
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -54,7 +54,7 @@
       install_requires=[
           'setuptools',
           'requests >= 1.0',
-          'ruamel.yaml >= 0.12.4',
+          'ruamel.yaml >= 0.12.4, <= 0.15.2',
```
mr-c: Shouldn't they both be `< 0.15` due to the API changes & warnings?

manu-chroma: I was using ``0.15.2`` due to some mypy updates which were released in this version, but yeah they didn't help much. I will change to ``ruamel.yaml<=0.15``

mr-c: just `<` please -- unless you need something from `0.15`

PR #456, opened 2017-07-07
Tweak SoftwareRequirement handling for reuse in Toil.

jmchilton:
With these changes I can get the requirement handling from #214 to work properly in simple toil tests (after a couple tweaks to toil itself). ~~I'll mark this as WIP until #214 is merged - though comments on the last few commits are more than welcome.~~

There are three commits here:
 - aa1b7c996040026101e095526a06b088f88c0c36 - the main refactor which seems uniformly positive.
 - 5069580853d6f2287a590a4d9b87d8793ebb0470 - which fixes a bug introduced at the last second in #214.
 - 33d1e96b90054c94ba2d3f93fda9a6f6e4d2190a - contains a small tweak which might not be so great. It restores the ``builder`` parameter in ``single_job_executor`` on jobs before executing if it is supplied to the method. I'm guessing I'm somehow depending on ``builder`` later in the job lifecycle than the previous code paths did and I'm not sure if this is a bad thing?

### DIFF
--- a/cwltool/builder.py
+++ b/cwltool/builder.py
@@ -42,7 +42,6 @@ class Builder(object):
         self.pathmapper = None  # type: PathMapper
         self.stagedir = None  # type: Text
         self.make_fs_access = None  # type: Type[StdFsAccess]
-        self.build_job_script = None  # type: Callable[[List[str]], Text]
         self.debug = False  # type: bool
         self.mutation_manager = None  # type: MutationManager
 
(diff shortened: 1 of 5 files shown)


jmchilton:
Matching Toil changes are at https://github.com/jmchilton/toil/tree/cwl_software_requirements.

jmchilton:
Did some more hacking on this and I have updated the description and taken this out of WIP.

I did some more testing on the Toil branch as well and updated the commit (https://github.com/jmchilton/toil/commit/4788dc719c9df71ca173ef04b14ce2664753bf4d) - it looks like now all the options including BioContainer support that work in cwltool also work in cwltoil. Descriptions and examples included in that commit message.

mr-c:
Thanks!

Issue #384, opened 2017-05-18
Convert cwl to json input


dridk:
How can I generate my job.json file ? 
would be greate to have : 

cwltools make jobtemplate pipleline.cwl  > pipeline-job.json

mr-c:
Hello @dridk , this is a great proposal -- I agree that would be very useful!

denis-yuen:
@dridk FYI, this isn't part of cwltool although it is built on top of it. 
We have a similar function as part of Dockstore (see the "make a runtime json" part of https://dockstore.org/containers/quay.io/wtsicgp/dockstore-cgpmap ) from this part of the code https://github.com/common-workflow-language/cwlavro/blob/2.0.0/cwlavro-tools/src/main/java/io/cwl/avro/CWL.java#L122

bolton-lab:
@dridk I am confused about how to setup the tool?  The documentation is good but it's still hard to follow the requirements.  I have a CWL file here https://github.com/kbolton-lab/analysis-workflows/blob/master/definitions/tools/bcftools_filter.cwl and added `CWL Path: /definitions/tools/bcftools_filter.cwl` and it uses a docker image that I did not create so I don't have the Dockerfile in my git repo.  Should we just be able to add that docker image as `Docker Image: hub.docker.com/mgibio/bcftools-cwl` add the 1.9 tag and run:

`dockstore tool convert entry2json --entry hub.docker.com/mgibio/bcftools-cwl:1.9`

mr-c:
For `cwltool` it would be `cwltool --make-template https://github.com/kbolton-lab/analysis-workflows/raw/master/definitions/tools/bcftools_filter.cwl`

@bolton-lab  For asking questions about Dockstore, probably better to do that in https://github.com/dockstore/dockstore/issues

``` yaml
vcf:  # type "File"
    class: File
    path: a/file/path
output_vcf_name: "bcftools_filter.vcf.gz"  # default value of type "string". (optional)
[...]

Issue #455, opened 2017-07-06
regression: incorrect warning in the presence of namespaced extension fields


mr-c:
See https://github.com/common-workflow-language/workflows/issues/137 etc..

mr-c:
Fixed in https://github.com/common-workflow-language/cwltool/pull/656

mr-c:
Conversation has moved to https://github.com/common-workflow-language/cwltool/issues/623 but the issue is not yet fixed.

PR #462, opened 2017-07-12
[wip] simple input object generation

mr-c:
Inspired by Dockstore's equivalent feature 

Will close https://github.com/common-workflow-language/cwltool/issues/384

### DIFF
--- a/cwltool/main.py
+++ b/cwltool/main.py
@@ -444,6 +444,7 @@ def generate_parser(toolparser, tool, namemap, records):
     return toolparser
 
 def generate_example_input(inptype):
+    # type: (Union[Text, Dict[Text, Any]]) -> Any
     defaults = { 'null': 'null',
                  'Any': 'null',
                  'boolean': False,
@@ -458,7 +459,7 @@ def generate_example_input(inptype):
                                 'path': 'default/directory/path' } }
     if (not isinstance(inptype, str) and
         not isinstance(inptype, collections.Mapping)
-        and isinstance(inptype, collections.Iterable)):
+        and isinstance(inptype, collections.MutableSet)):
         if len(inptype) == 2 and 'null' in inptype:
             inptype.remove('null')
             return generate_example_input(inptype[0])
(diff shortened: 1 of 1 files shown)


ctb:
Note that 

```
cwltool --make-template https://github.com/CancerCollaboratory/dockstore-tool-bamstats/raw/develop/Dockstore.cwl > params.yaml
```

succeeds but shows the following error message that made me think it did not succeed:

```
/home/tx160085/env/bin/cwltool 1.0
https://github.com/CancerCollaboratory/dockstore-tool-bamstats/raw/develop/Dockstore.cwl:3:1: unrecognized extension field `http://purl.org/dc/terms/creator`.  Did you include a $schemas section?
```

Any way to change this to a warning, or be more verbose about the fact that params.yml is legit?

ctb:
Also - the dockstore output includes a section for the output file, e.g. `bamstats_report` in the example that https://github.com/ngs-docs/2017-cloud-workflows-misc/pull/1 is modifying. This doesn't seem to be the case with the `--make-template` command. Is that what you mean in [this comment](https://github.com/ngs-docs/2017-cloud-workflows-misc/pull/1#issuecomment-314771135) by `uses the standardized --outdir option`? thx!

mr-c:
@ctb thank you for the verification. This is incomplete but I'll release it for later improvement :-)

Re: output file name in the input object -- yes, that is a dockstore specific extension (and one that I never personally understood). Use the standardized `--outdir` if you'd like to put your outputs anywhere besides the current working directory. If you'd like to discover the filename of a particular output you can parse the output JSON to find out (some workflows the file names vary).

mr-c:
re: the warning, this is a known regression https://github.com/common-workflow-language/cwltool/issues/455

denis-yuen:
@ctb @mr-c  FYI, the output file name is for specifying an output location https://dockstore.org/docs/advanced-features#file-provisioning 

Note that putting it in `/tmp` works but is probably a less interesting example than putting the output into S3, Synapse, etc.

mr-c:
@denis-yuen Ah, that makes more sense. It would be more portable if you added these as a `dockstore` specific directives, otherwise it looks like another input which is confusing to CWL runners and humans alike:

Example:
``` yaml
bam_input:
    class: File
    path: /tmp/NA12878.chrom20.ILLUMINA.bwa.CEU.low_coverage.20121211.bam
dockstore:outputs:
    bamstats_report:
        class: File
        path: /tmp/bamstats_report.zip
```

This should be proposed as a standardized optional feature, then `dockstore:outputs` could become `cwl:outputs`

denis-yuen:
@mr-c I think that would be a reasonable improvement. It would simplify our code too.

PR #467, opened 2017-07-13
Retain "outdir" in WorkflowJob.job.

jxtx:
Otherwise, the `--tmp-outdir-prefix` option passed on the command line is not respected, and outputs are instead written to the system temp directory.


mr-c:
Jenkins, add to test list

mr-c:
Jenkins, add to testlist

tetron:
I think the reason it deletes outdir from kwargs because it sets `self.outdir` in the constructor.  That is supposed to respect `tmp_outdir_prefix`, if it doesn't then that is a bug.  If kwargs["outdir"] isn't removed, it would be passed to the workflow steps, and then each step would use the same value of outputdir instead of using separate ones for each step.

mr-c:
@jxtx @tetron I believe I found the real bug; `--tmp-outdir-prefix` was only being consulted for containerized steps. Please review and test

cwltool/process.py
```diff
@@ -622,7 +623,8 @@ def _init_job(self, joborder, **kwargs):
             builder.tmpdir = builder.fs_access.docker_compatible_realpath(kwargs.get("docker_tmpdir") or "/tmp")
             builder.stagedir = builder.fs_access.docker_compatible_realpath(kwargs.get("docker_stagedir") or "/var/lib/cwl")
         else:
-            builder.outdir = builder.fs_access.realpath(kwargs.get("outdir") or tempfile.mkdtemp())
+            builder.outdir = builder.fs_access.realpath(kwargs.get("outdir")
+                    or tempfile.mkdtemp(prefix=kwargs["tmp_outdir_prefix"]))
             if self.tool[u"class"] != 'Workflow':
                 builder.tmpdir = builder.fs_access.realpath(kwargs.get("tmpdir") or tempfile.mkdtemp())
```
tetron: Does this need to be fixed as well?

tetron:
@mr-c I think you're right.  There might be a few other places with a similar bug?

mr-c:
Jenkins, test this please

PR #477, opened 2017-07-18
setup.py: remove condition which blocks python 3 install

manu-chroma:
the installation will not be blocked when running ``python setup.py install`` under Python 3 runtime.


mr-c:
This has been released as `cwltool` `1.0.20170718140316`

manu-chroma:
This wasn't breaking any user Python 3 installations. It fixed the case when someone tried to install from source. I use ``pip install -e .`` in dev, that's why it went unnoticed.

PR #478, opened 2017-07-19
changing default container for windows from ubuntu to alpine-bash

kapilkd13:


mr-c:
Why is AppVeyor running twice?

kapilkd13:
Actually I pushed commits to testing-windows branch, which is base branch for windows comaptibility. Since it was merged I created a PR. I searched and found that build is initiated on appveyor when

> Builds are triggered on:

    push to branch
    merge PR to branch
    open PR
    sync PR (when you do push to branch with open PR)

So out of two builds I guess the PR one is for this PR and the branch one is for a new commit pushed to testing-windows branch. But I think it shouldn't do this as the testing-windows branch is already merged. I will look for a solution, if not found I will create new branch for this commit. BTW, is it a bad practice to push commits to a merged branch.

kapilkd13:
I tried creating new PR using above solution but still it shows branch PR . https://github.com/common-workflow-language/cwltool/pull/479 I guess we may need to close branch build toggle. I will see.

kapilkd13:
Hi @mr-c , can you move the `required` badge from appveyor/branch to appveyor/pr in branch protection. I think this would solve our problem.

mr-c:
@kapilkd13 Done

Issue #480, opened 2017-07-20
support secondaryFiles for File in CommandInputRecordField


liuxf09:
I have a input parameter that is an array of records, like the following:

    resources:
      type:
        type: array
        items:
        - type: record
          name: resource
          fields:
            - name: resource
              type: File
              inputBinding:
                prefix: --resource
            - name: resource_param
              type: string
              inputBinding:
                prefix: --resource_param

In the record, resource is File object, and will require `secondaryFiles`. But `secondaryFiles` is only allowed for `CommandInputParameter`, and not allowed in `CommandInputRecordField`. Ideally I want to something like the following to work:

    resources:
      type:
        type: array
        items:
        - type: record
          name: resource
          fields:
            - name: resource
              type: File
              secondaryFiles:
                - ".ext"
              inputBinding:
                prefix: --resource
            - name: resource_param
              type: string
              inputBinding:
                prefix: --resource_param
My reason to use array of records here is to combine resource and resource_param together to produce command line like `--resource --resource_param --resource --resource_param`. 
One fix is to support `secondaryFiles` for File in CommandInputRecordField:
    "In `schemas/v1.0/Process.yml`, there is a `secondaryFiles` field under `Parameter`. Just copy this field to under InputRecordField after label field. "
This was posted as a question on BioStars ( https://www.biostars.org/p/263577/#263643 ).

PR #481, opened 2017-07-20
support secondaryFiles for File in CommandInputRecordField #480

liuxf09:
refer to https://github.com/common-workflow-language/cwltool/issues/480. 
The fix will change the spec for InputRecordField, and the change is made into cwltool/schemas/v1.1.0-dev1/Process.yml.


liuxf09:
What is the status for this ticket?

tetron:
hi @liuxf09 thank you for your contribution.

This fix will be included in v1.1 of the CWL spec, development of which is currently underway.

cwltool support for v1.1 is in progress on this branch and includes this fix:

https://github.com/common-workflow-language/cwltool/pull/694

Use `cwltool --enable-dev` to run files with `cwlVersion: v1.1.0-dev1`

We would appreciate feedback as to whether this solves your problem (or not).

PR #484, opened 2017-07-21
Update "path" of files staged to InitialWorkDirRequirement.

tetron:
To conform to v1.0.1 behavior.


tetron:
Incorporated into #485

Issue #469, opened 2017-07-16
[windows] if the Windows default container is used, print a warning


mr-c:
But if the default container wasn't needed then don't mention it.

Warning text (may need to be refined):

```
"""We are on Microsoft Windows and not all components of this CWL description have a
container specified. The default container is %s.

Note, this could affect portability if this CWL description relies on non-POSIX features
or commands in this container. For best results add the following to your CWL
description's hints section:

hints:
  DockerRequirement:
    dockerPull: %s
""" % windows_default_container, windows_default_container
```

mr-c:
Comments on the wording are appreciated, especially from @jvdzwaan

kapilkd13:
Hi Michael, Can you explain `if default container isn't used don't show this warning`. But on windows all execution is happening under docker container, so I guess default container is always used, isn't it.

jvdzwaan:
Good idea to add a warning! I would make it more explicit that all steps without a docker hint/requirement will be executed in the default container. Something like: 

```
We are on Microsoft Windows and not all components of this CWL description have a
container specified. This means that these steps will be executed in the default container, 
which is %s.
```

Also, I don't understand why you would advise to add a hint for the default container. Shouldn't we advise them to use a more appropriate container if they have non-POSIX features in their CWL?

mr-c:
@kapilkd13 correct, if all components of the Workflow, CommandLineTool, or ExpressionTool have a `DockerRequirement` or inherit a `DockerRequirement` (except from the `--default-container` or the Windows fallback default container) then the warning should be printed under Windows.

mr-c:
@jvdzwaan The default container used by `cwltool` could change over time. Your idea that a note that (if needed) the user should specify another container, is a good idea.

mr-c:
Fixed in #492

PR #492, opened 2017-07-22
adding warning when default docker container is used on Windows

kapilkd13:
referencing issue #469


kapilkd13:
@mr-c I need a little help here. We need a test to check that if default container is used on windows, a warning should be given. Although on local system I can test and verify this behavior but since this behavior is `windows with docker` specific, how should we test it in our CI.

mr-c:
No worries, how about a mock object?

https://en.wikipedia.org/wiki/Mock_object

Pe 25 iul. 2017 10:05 p.m., "Kapil kumar" <notifications@github.com> a
scris:

kapilkd13:
Hi @mr-c  I added a test using mock object as suggested by you. That was something new for me. 😄  Can you review it.

mr-c:
Huzzah -- thanks!

Issue #471, opened 2017-07-17
update Jenkins script to add Python 3 conformance tests


manu-chroma:

manu-chroma:
@mr-c Is ``version`` an env variable when running ``jenkins.bash`` script: https://github.com/common-workflow-language/cwltool/blob/master/jenkins.bash#L37

PR #505, opened 2017-07-29
jenkins.bash: Add support for Python 3 conformance testing

manu-chroma:
Closes https://github.com/common-workflow-language/cwltool/issues/471

### DIFF
--- a/jenkins.bash
+++ b/jenkins.bash
@@ -20,12 +20,13 @@ venv() {
         source "$1"/bin/activate
 }
 
+git clean --force -d -x || /bin/true
+cloneorpull common-workflow-language https://github.com/common-workflow-language/common-workflow-language.git
+docker pull node:slim
+
 # Test for Python 2.7 and Python 3
 for PYTHON_VERSION in 2 3
 do
-	git clean --force -d -x || /bin/true
-	cloneorpull common-workflow-language https://github.com/common-workflow-language/common-workflow-language.git
-	docker pull node:slim
 	venv cwltool-venv
 	export PIP_DOWNLOAD_CACHE=/var/lib/jenkins/pypi-cache/
 	pip install -U setuptools wheel pip
--- a/jenkins.bash
+++ b/jenkins.bash
@@ -25,13 +25,14 @@ cloneorpull common-workflow-language https://github.com/common-workflow-language
 docker pull node:slim
 
 # Test for Python 2.7 and Python 3
-for PYTHON_VERSION in 2 3
+for PYTHON_VERSION in 2.7 3
 do
-	venv cwltool-venv
+	venv cwltool-venv${PYTHON_VERSION}
 	export PIP_DOWNLOAD_CACHE=/var/lib/jenkins/pypi-cache/
-	pip install -U setuptools wheel pip
-	pip install .
-	pip install "cwltest>=1.0.20160825151655"
+	# use pip2.7 and pip3 in separate loop runs
+	pip${PYTHON_VERSION} install -U setuptools wheel pip
+	pip${PYTHON_VERSION} install .
+	pip${PYTHON_VERSION} install "cwltest>=1.0.20160825151655"
 	pushd common-workflow-language
 	git clean --force -d -x || /bin/true
 	# shellcheck disable=SC2154
--- a/jenkins.bash
+++ b/jenkins.bash
@@ -14,29 +14,37 @@ cloneorpull() {
 }
 venv() {
         if ! test -d "$1" ; then
-                virtualenv "$1"
+                virtualenv -p python${PYTHON_VERSION} "$1"
         fi
 	# shellcheck source=/dev/null
         source "$1"/bin/activate
 }
-git clean --force -d -x || /bin/true
-cloneorpull common-workflow-language https://github.com/common-workflow-language/common-workflow-language.git
-venv cwltool-venv
-docker pull node:slim
-export PIP_DOWNLOAD_CACHE=/var/lib/jenkins/pypi-cache/
-pip install -U setuptools wheel pip
-python setup.py install
-pip install "cwltest>=1.0.20160825151655"
-pushd common-workflow-language
-git clean --force -d -x || /bin/true
-# shellcheck disable=SC2154
-if [[ "$version" = *dev* ]]
-then
-	EXTRA="EXTRA=--enable-dev"
-fi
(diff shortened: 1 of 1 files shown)


jenkins.bash
```diff
@@ -14,29 +14,37 @@ cloneorpull() {
...
-	EXTRA="EXTRA=--enable-dev"
-fi
-./run_test.sh --junit-xml=result.xml RUNNER=cwltool -j4 DRAFT="${version}" ${EXTRA}
-CODE=$?
-popd
+
+# Test for Python 2.7 and Python 3
+for PYTHON_VERSION in 2 3
+do
+	git clean --force -d -x || /bin/true
```
mr-c: This deletes the test results on the second go through the loop. Move lines 26-28 outside?

manu-chroma: Thanks! :D

jenkins.bash
```diff
@@ -14,29 +14,38 @@ cloneorpull() {
...
-python setup.py install
-pip install "cwltest>=1.0.20160825151655"
-pushd common-workflow-language
-git clean --force -d -x || /bin/true
-# shellcheck disable=SC2154
-if [[ "$version" = *dev* ]]
-then
-	EXTRA="EXTRA=--enable-dev"
-fi
-./run_test.sh --junit-xml=result.xml RUNNER=cwltool -j4 DRAFT="${version}" ${EXTRA}
```
manu-chroma: @mr-c Is ``version`` an env variable when running the script?

manu-chroma:
I'm afraid it's running tests twice in Python 2 rather than running in Python 3 in the second run of the loop. https://ci.commonwl.org/job/cwltool-pr-conformance-multiver/version=v1.0/659/console

jenkins.bash
```diff
@@ -14,29 +14,38 @@ cloneorpull() {
...
+# Test for Python 2.7 and Python 3
+for PYTHON_VERSION in 2 3
+do
+	venv cwltool-venv
+	export PIP_DOWNLOAD_CACHE=/var/lib/jenkins/pypi-cache/
+	pip install -U setuptools wheel pip
+	pip install .
+	pip install "cwltest>=1.0.20160825151655"
+	pushd common-workflow-language
+	git clean --force -d -x || /bin/true
```
manu-chroma: is it ok to delete results in ``common-workflow-lang`` repo?

manu-chroma:
Builds correctly now for Python 2.7 and Python 3.4 https://ci.commonwl.org/job/cwltool-pr-conformance-multiver/version=v1.0/660/console :)

jenkins.bash
mr-c: yes

mr-c: currently only set to `v1.0` will be `v1.1.0-dev1` later, etcetera

mr-c:
- [ ] @mrc to re-enable test parsing on Jenkins

jenkins.bash
```diff
@@ -14,29 +14,39 @@ cloneorpull() {
...
+for PYTHON_VERSION in 2.7 3
+do
+	venv cwltool-venv${PYTHON_VERSION}
+	export PIP_DOWNLOAD_CACHE=/var/lib/jenkins/pypi-cache/
+	# use pip2.7 and pip3 in separate loop runs
+	pip${PYTHON_VERSION} install -U setuptools wheel pip
+	pip${PYTHON_VERSION} install .
+	pip${PYTHON_VERSION} install "cwltest>=1.0.20160825151655"
+	pushd common-workflow-language
+	git clean --force -d -x || /bin/true
```
manu-chroma: Do we need to call ``git clean`` in both repos: ``cwltool`` and ``common-workflow-language``? @mr-c

jenkins.bash
```diff
@@ -14,29 +14,39 @@ cloneorpull() {
...
+	# shellcheck disable=SC2154
+	if [[ "$version" = *dev* ]]
+	then
+		EXTRA="EXTRA=--enable-dev"
+	fi
+	./run_test.sh --junit-xml=result.xml RUNNER=cwltool -j4 DRAFT=${version}
```
manu-chroma: @mr-c Should we've have different names for the result file from both runtimes?

mr-c: Probably safest, yes.

jenkins.bash
mr-c: Yes, but we have to be careful not to delete the test results from the other Python 2/3 run. May use different checkouts or otherwise change where the results are written to?

manu-chroma:
@mr-c I've addressed both the concerns. 

Test results are now written to: 
```
	v1.0/result2.7.xml
	v1.0/result3.xml
```
inside ``common-workflow-language`` repo

jenkins.bash
```diff
@@ -14,29 +14,40 @@ cloneorpull() {
...
+	pip${PYTHON_VERSION} install "cwltest>=1.0.20160825151655"
+	pushd common-workflow-language
+	# shellcheck disable=SC2154
+	if [[ "$version" = *dev* ]]
+	then
+		EXTRA="EXTRA=--enable-dev"
+	fi
+	./run_test.sh --junit-xml=result${PYTHON_VERSION}.xml RUNNER=cwltool -j4 DRAFT=${version}
+	deactivate
+	CODE=$?
```
mr-c: CODE will be overwritten the second time through

manu-chroma: @mr-c I don't understand the function of ``CODE`` in the script. But it can be modified to ``CODE${PYTHON_VERSION}`` so it does not get overwritten

mr-c: Look at the last line, its used to return the true error code (success or failure) of running the tests. Right now it's capturing the return code of `deactivate` instead of `run_test.sh` and the value is overwritten during the second loop. So move it back up a line and replace it with `CODE=$(($CODE+$?))` so that it accumulates. Remember: exit code `0` means "no problems" :-)

manu-chroma: @mr-c Great explanation, thanks! :D

mr-c:
Sweet, the test results are showing up -- though they are combined into the same report: https://ci.commonwl.org/job/cwltool-pr-conformance-multiver/687/version=v1.0/testReport/conformance_test_v1/0/

manu-chroma:
Yayy! :D 
>  though they are combined into the same report: 

Should we do something about this now or later?

mr-c:
Later, I'm filing an issue with `cwltest`

PR #506, opened 2017-07-30
adding --docker-pull flag to force pull image even if the docker image is already present

### LINKED ISSUE #422: dockerPull field in DockerRequirement should pull image updates before running
I have been running some CWL workflows that utilize the DockerRequirement "dockerPull" field, and when the Docker image in question is not present locally, the field works as intended, pulling the image first before running. However, if the image is updated in the Docker registry sometime later, the "dockerPull" field will continue to run the local image with no actual attempt to do a "docker pull" for updates. This requires me to have to do a manual "docker pull" on the command line before invoking the CWL. Is it possible to have CWL perform an actual "docker pull" every time the workflow is invoked with the "dockerPull" field (to grab the latest image update if available), as opposed to simply doing a "docker run" if the image already exists locally?

kapilkd13:
If this flag is present we strictly pull latest docker image even if the image is present locally

Closes #422

### DIFF
--- a/cwltool/builder.py
+++ b/cwltool/builder.py
@@ -49,7 +49,7 @@ class Builder(object):
         self.make_fs_access = None  # type: Type[StdFsAccess]
         self.debug = False  # type: bool
         self.mutation_manager = None  # type: MutationManager
-        self.strict_docker_pull = False  # type: bool
+        self.force_docker_pull = False  # type: bool
 
         # One of "no_listing", "shallow_listing", "deep_listing"
         # Will be default "no_listing" for CWL v1.1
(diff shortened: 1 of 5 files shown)


cwltool/builder.py
```diff
@@ -49,6 +49,7 @@ def __init__(self):  # type: () -> None
         self.make_fs_access = None  # type: Type[StdFsAccess]
         self.debug = False  # type: bool
         self.mutation_manager = None  # type: MutationManager
+        self.strict_docker_pull = False  # type: bool
```
manu-chroma: ``force_docker_pull`` is a better name IMO

kapilkd13: Yes, that sounds better. I would change it. Thanks

kapilkd13:
@mr-c does is look ready for merge

Issue #466, opened 2017-07-13
downloads remote inputs via HTTP(S)


mr-c:
## Expected Behavior
a URI should be accepted for inputs with `type: File`

http://www.commonwl.org/v1.0/CommandLineTool.html#File
## Actual Behavior
> No such file or directory: '/home/michael/src/2017-cloud-workflows-misc/http://github.com/CancerCollaboratory/dockstore-tool-bamstats/raw/develop/rna.SRR948778.bam'

## Workflow Code
```
cwltool https://github.com/CancerCollaboratory/dockstore-tool-bamstats/raw/develop/Dockstore.cwl \
 --bam_input https://github.com/CancerCollaboratory/dockstore-tool-bamstats/raw/develop/rna.SRR948778.bam
```

## Full Traceback
```pytb
/home/michael/src/2017-cloud-workflows-misc/env/bin/cwltool 1.0.20170712193248
https://github.com/CancerCollaboratory/dockstore-tool-bamstats/raw/develop/Dockstore.cwl:3:1: unrecognized extension field `http://purl.org/dc/terms/creator`.  Did you include a $schemas section?
[job Dockstore.cwl] initializing from https://github.com/CancerCollaboratory/dockstore-tool-bamstats/raw/develop/Dockstore.cwl
[job Dockstore.cwl] {
    "bam_input": {
        "class": "File", 
        "location": "https://github.com/CancerCollaboratory/dockstore-tool-bamstats/raw/develop/rna.SRR948778.bam", 
        "basename": "rna.SRR948778.bam", 
        "nameroot": "rna.SRR948778", 
        "nameext": ".bam"
    }, 
    "mem_gb": 0
}
Got workflow error
Traceback (most recent call last):
  File "/home/michael/src/2017-cloud-workflows-misc/env/local/lib/python2.7/site-packages/cwltool/main.py", line 270, in single_job_executor
    for r in jobiter:
  File "/home/michael/src/2017-cloud-workflows-misc/env/local/lib/python2.7/site-packages/cwltool/draft2tool.py", line 323, in job
    builder.pathmapper = self.makePathMapper(reffiles, builder.stagedir, **make_path_mapper_kwargs)
  File "/home/michael/src/2017-cloud-workflows-misc/env/local/lib/python2.7/site-packages/cwltool/draft2tool.py", line 204, in makePathMapper
    return PathMapper(reffiles, kwargs["basedir"], stagedir)
[...]

PR #507, opened 2017-07-31
[WIP] Allowing Http/Https files as input 

kapilkd13:
To allow remote http/https files as input.  Refer Issue #466

### DIFF
--- a/cwltool/pathmapper.py
+++ b/cwltool/pathmapper.py
@@ -5,6 +5,9 @@ import os
 import stat
 import uuid
 from functools import partial
+from tempfile import NamedTemporaryFile
+
+import requests
 from typing import Any, Callable, Dict, Iterable, List, Set, Text, Tuple, Union
 
 import schema_salad.validate as validate
@@ -139,6 +142,15 @@ def trim_listing(obj):
     if obj.get("location", "").startswith("file://") and "listing" in obj:
         del obj["listing"]
 
+# Download http Files
+def downloadHttpFile(httpurl):
+        r = requests.get(httpurl, stream=True)
+        with NamedTemporaryFile(mode='wb', delete=False) as f:
+            for chunk in r.iter_content(chunk_size=1024):
+                if chunk:  # filter out keep-alive new chunks
+                    f.write(chunk)
+        r.close()
+        return f.name
 
 class PathMapper(object):
     """Mapping of files from relative path provided in the file to a tuple of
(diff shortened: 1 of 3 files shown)


tetron:
I assume this is a work in progress, since it doesn't actually download files yet.

kapilkd13:
Hi @tetron Yes, Currently its work in progress. Right now it solves the problem of validation error that was raised earlier. And the workflow error caused due to os.readlink on a web url.

cwltool/pathmapper.py
```diff
@@ -139,6 +142,15 @@ def trim_listing(obj):
     if obj.get("location", "").startswith("file://") and "listing" in obj:
         del obj["listing"]
 
+# Download http Files
+def downloadHttpFile(httpurl):
+        r = requests.get(httpurl, stream=True)
```
tetron: Can we use CacheControl to manage file caching?

cwltool/pathmapper.py
```diff
@@ -139,6 +142,15 @@ def trim_listing(obj):
...
+        with NamedTemporaryFile(mode='wb', delete=False) as f:
+            for chunk in r.iter_content(chunk_size=1024):
```
tetron: You'll get better throughput with a bigger chunk size (suggest at least 16k).

kapilkd13:
Hi @mr-c Just to confirm, we need a conformance test for this feature right?

mr-c:
@kapilkd13 Until https://github.com/common-workflow-language/common-workflow-language/issues/491 is fixed we can't add a conformance test for this as it is optional (though widely supported). So make it a test within `cwltool` only for now

cwltool/pathmapper.py
```diff
@@ -139,6 +144,30 @@ def trim_listing(obj):
     if obj.get("location", "").startswith("file://") and "listing" in obj:
         del obj["listing"]
 
+# Download http Files
+def downloadHttpFile(httpurl):
+    cache_session = None
+    if "HOME" in os.environ:
```
mr-c: Lets make this compliant with the [XDG Base Directory Specification](https://standards.freedesktop.org/basedir-spec/basedir-spec-0.6.html)'s `$XDG_CACHE_HOME` feature (if absent then the default is `$HOME/.config`) and if HOME isn't specified, then you can use [`os.path.expanduser('~')`](https://docs.python.org/3/library/os.path.html#os.path.expanduser) which means the bits with `TMP` and `/tmp` below aren't needed

kapilkd13: I made these changes, i assumed you meant `$HOME/.cache` when `XDG_CACHE_HOME` is not present

kapilkd13:
@mr-c  @tetron I think this feature is ready. Can you review this PR

cwltool/pathmapper.py
```diff
@@ -139,6 +144,33 @@ def trim_listing(obj):
...
+    elif "HOME" in os.environ:
+        cache_session = CacheControl(
+            requests.Session(),
+            cache=FileCache(
+                os.path.join(os.environ["HOME"], ".cache", "cwltool")))
+    else:
+        cache_session = CacheControl(
+            requests.Session(),
+            cache=FileCache(
+                os.path.join(os.path.expanduser('~'), ".cache", "cwltool")))
```
mr-c: Can you refactor this so the if/elif/else is just determining the directory to put `.cache` in and the call to CacheControl occurs afterwards?

mr-c:
woo-hoo! Thank you @kapilkd13 !

PR #491, opened 2017-07-22
load_tool: when no version or incorrect cwlVersion is found, create fatal error

manu-chroma:
Fixes https://github.com/common-workflow-language/cwltool/issues/341, https://github.com/common-workflow-language/cwltool/issues/140

PR #509, opened 2017-08-02
load_tool.py: Pass to regex before checking for acceptable versions

manu-chroma:
Addresses https://github.com/common-workflow-language/cwltool/commit/2c762b536caa0062aaf26d81df7745e246a6a1cf#commitcomment-23435763, https://github.com/common-workflow-language/cwltool/pull/491#issuecomment-318851330

cc: @jmchilton


jmchilton:
Great - thanks a bunch!

PR #510, opened 2017-08-02
Appveyor py3 fix: Earlier py3 was falling back to py2

kapilkd13:
Due to some bad configuration, unit testing for python 3 was falling back to python 2. This disclosed few issues we need to resolve before achieving python 3 compatibility on windows.


kapilkd13:
these commits are moved to #511

PR #512, opened 2017-08-04
cwltool/cwlrdf.py: Pass str rather than bytes to write function (Python 3 fix)

manu-chroma:
Fixes ``--print-rdf`` flag under ``Python 3``. ``Python 2.7`` behaviour is unchanged

Before the changes, we got the following error: 
``` pytb
/home/manu/github/cwltool/venv3/bin/cwltool 1.0.20170801042110
I'm sorry, I couldn't load this CWL file.  The error was:
  File "/home/manu/github/cwltool/cwltool/main.py", line 840, in main
    printrdf(tool, document_loader.ctx, args.rdf_serializer, stdout)
  File "/home/manu/github/cwltool/cwltool/cwlrdf.py", line 25, in printrdf
    stdout.write(gather(wf, ctx).serialize(format=sr))
TypeError: write() argument must be str, not bytes
```

### DIFF
--- a/tests/test_rdfprint.py
+++ b/tests/test_rdfprint.py
@@ -1,0 +1,55 @@
+from __future__ import absolute_import
+import unittest
+
+from six import StringIO
+from cwltool.main import main
+
+class RDF_Print(unittest.TestCase):
+
+    def test_rdf_print(self):
+        self.maxDiff = None
+
+        expected_output = """@prefix CommandLineBinding: <https://w3id.org/cwl/cwl#CommandLineBinding/> .
+@prefix CommandLineTool: <https://w3id.org/cwl/cwl#CommandLineTool/> .
+@prefix CommandOutputBinding: <https://w3id.org/cwl/cwl#CommandOutputBinding/> .
+@prefix Dirent: <https://w3id.org/cwl/cwl#Dirent/> .
+@prefix DockerRequirement: <https://w3id.org/cwl/cwl#DockerRequirement/> .
+@prefix EnvVarRequirement: <https://w3id.org/cwl/cwl#EnvVarRequirement/> .
+@prefix EnvironmentDef: <https://w3id.org/cwl/cwl#EnvironmentDef/> .
+@prefix ExpressionTool: <https://w3id.org/cwl/cwl#ExpressionTool/> .
+@prefix File: <https://w3id.org/cwl/cwl#File/> .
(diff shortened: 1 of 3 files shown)


mr-c:
Thanks! Can you add a test that proves the fix?

mr-c:
- [ ] refactor the function to make it more testable

manu-chroma:
The command line args for ``printrdf`` function leads to alot of other function calls. It makes sense to simply pass the ``argsl`` in the ``cwltool.main.main``.

manu-chroma:
@mr-c

tests/test_rdfprint.py
```diff
@@ -0,0 +1,55 @@
...
+    cwl:cwlVersion cwl:v1.0 ;
+    cwl:outputs <https://raw.githubusercontent.com/common-workflow-language/workflows/master/workflows/hello/hello.cwl#response> ;
+    ns1:comment "Outputs a message using echo" ;
+    ns1:label "Hello World" .
+
+<https://raw.githubusercontent.com/common-workflow-language/workflows/master/workflows/hello/hello.cwl#response> cwl:outputSource <https://raw.githubusercontent.com/common-workflow-language/workflows/master/workflows/hello/hello.cwl#step0/response> ;
+    sld:type cwl:File .
+"""
+
+        argsl = ['--print-rdf', 'https://raw.githubusercontent.com/common-workflow-language/workflows/master/workflows/hello/hello.cwl']
```
mr-c: It would be best if this tested a local resource and compared against a local file instead of a large embedded string.

A minimal test case that would still be useful is to call the function using a local file and assert that no exceptions get thrown :-)

manu-chroma: Ah, using local resource definitely makes more sense. But the problem is that the generated output has a lot of random strings involved. eg: 
```
<https://raw.githubusercontent.com/common-workflow-language/workflows/master/workflows/hello/hello.cwl#step0> cwl:out <https://raw.githubusercontent.com/common-workflow-language/workflows/master/workflows/hello/hello.cwl#step0/response> ;
    cwl:run <https://raw.githubusercontent.com/common-workflow-language/workflows/master/workflows/hello/hello.cwl#step0/49fc3536-04ae-44a5-a6fd-4fa93dd2e1a6> .
```

I will go with the minimal test case approach :+1:

mr-c:
Thanks!

PR #516, opened 2017-08-09
Refactor secondaryFile handling to avoid basing on primary "location".

tetron:
Instead prefer to use basename or path.  Necessary to support engines such as
Toil with non-filename-preserving storage systems.


tetron:
Oh good, the test finally passed.

Issue #250, opened 2016-12-09
Error "'HTTPResponse' object has no attribute 'chunked'" when running cwltool


muffato:
Hi there,
I got this error on a fresh install of cwltool + the workflows
```
$ cwltool --debug workflows/tools/picard-MergeSamFiles.cwl
/home/matthieu/.local/bin/cwltool 1.0.20161207161158
Resolved 'workflows/tools/picard-MergeSamFiles.cwl' to 'file:///home/matthieu/workspace/hive/cwl/workflows/tools/picard-MergeSamFiles.cwl'
Tool definition failed initialization:
('http://dublincore.org/2012/06/14/dcterms.rdf', AttributeError("'HTTPResponse' object has no attribute 'chunked'",))
Traceback (most recent call last):
  File "/home/matthieu/.local/lib/python2.7/site-packages/cwltool/main.py", line 638, in main
    fetcher_constructor=fetcher_constructor)
  File "/home/matthieu/.local/lib/python2.7/site-packages/cwltool/load_tool.py", line 164, in validate_document
    processobj, metadata = document_loader.resolve_all(workflowobj, fileuri)
  File "/home/matthieu/.local/lib/python2.7/site-packages/schema_salad/ref_resolver.py", line 628, in resolve_all
    newctx.add_schemas(document[u"$schemas"], file_base)
  File "/home/matthieu/.local/lib/python2.7/site-packages/schema_salad/ref_resolver.py", line 269, in add_schemas
    content = self.fetch_text(fetchurl)
  File "/home/matthieu/.local/lib/python2.7/site-packages/schema_salad/ref_resolver.py", line 104, in fetch_text
    raise RuntimeError(url, e)
RuntimeError: ('http://dublincore.org/2012/06/14/dcterms.rdf', AttributeError("'HTTPResponse' object has no attribute 'chunked'",))
```

I have a working internet connection and can download the file with `wget` but also from within python
```
$ python
Python 2.7.6 (default, Oct 26 2016, 20:30:19) 
[GCC 4.8.4] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import urllib
>>> h = urllib.urlopen('http://dublincore.org/2012/06/14/dcterms.rdf')
>>> '<rdf' in h.read()
True
```

muffato:
[...]

PR #517, opened 2017-08-09
changing requests min version to 2.12.4

kapilkd13:
As suggested by Peter, we are moving to `2.12.4` as the min version for requests module

Closes: https://github.com/common-workflow-language/cwltool/issues/250


mr-c:
Thanks!

PR #518, opened 2017-08-11
Move version constraint on requests back to 2.4.3 (ships with debian 8)

tetron:


tetron:
See https://github.com/common-workflow-language/cwltool/issues/250#issuecomment-321899030

Issue #401, opened 2017-05-30
update README's test execution instructions


mr-c:
basically:
`pip install tox; tox`

psafont:
How is it different from running the current command? (`python setup.py test`)

manu-chroma:
@psafont tox does ``pip install requirements`` before calling ``setup.py test``

manu-chroma:
and another added benefit is it creates fresh virtual environment

manu-chroma:
does it make sense to move the tests instruction in ``CONTRIBUTING.md`` rather than having them in ``README``?

manu-chroma:
@mr-c can we do detox? really speeds up the testing 
```
pip install detox; detox
```

mr-c:
Neat, I've never seen that before. That can go as an additional option.

manu-chroma:
Yeah, just a wrapper around tox :smile: 

https://github.com/tox-dev/detox#what-is-detox

PR #523, opened 2017-08-14
README: Fix linting and minor errors

manu-chroma:
Closes https://github.com/common-workflow-language/cwltool/issues/401

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,26 +1,24 @@
 ==================================================================
-Common workflow language tool description reference implementation
+Common Workflow Language tool description reference implementation
 ==================================================================
 
-CWL Conformance test: |Build Status|
-
-Travis: |Unix Build Status|
+CWL conformance tests: |Build Status| Travis CI: |Unix Build Status|
 
 .. |Unix Build Status| image:: https://img.shields.io/travis/common-workflow-language/cwltool/master.svg?label=unix%20build
    :target: https://travis-ci.org/common-workflow-language/cwltool
 
 This is the reference implementation of the Common Workflow Language.  It is
-intended to be feature complete and provide comprehensive validation of CWL
+intended to feature complete and provide comprehensive validation of CWL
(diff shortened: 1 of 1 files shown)


manu-chroma:
@mr-c Can you help me with this error? 
```
running check
warning: Check: :206: (WARNING/2) Inline literal start-string without end-string.

warning: Check: Invalid markup which will not be rendered on PyPI.

error: Please correct your package.
```

manu-chroma:
It is rendering correctly when converted to HTML. Also, the error produced is a warning but still returns a failure code.  

image

manu-chroma:
I've a wild proposition to make: Convert current ``rst`` doc to ``markdown`` and have something like this: https://github.com/mr-karan/webkin/blob/master/setup.py#L5 

I will be happy to do this. Editing ``rst`` docs take up just too much time. Your thoughts @mr-c

mr-c:
@manu-chroma When rendering was there a specific error or warning?

manu-chroma:
@no, ``pandoc``directly gave ``HTML`` output. 
Seeing ``pandoc --help``, there does not seem to be any error checking mechanism in place.

mr-c:
I used https://pypi.python.org/pypi/restview and found the formatting errors, just a couple missing spaces

manu-chroma:
> I used pypi.python.org/pypi/restview and found the formatting errors, just a couple missing spaces

Just tried this. Error messages are same as that of ``tox``. @mr-c Can you help me in the fix?

mr-c:
@manu-chroma see my commit :-P

mr-c:
RST needs a space after the ending double backtick of an inline literal: http://docutils.sourceforge.net/docs/user/rst/quickref.html#inline-markup

manu-chroma:
@mr-c Thanks. Missed it in first glance :P

PR #528, opened 2017-08-15
added no-read-only option

SooLee:
Added `--no-read-only` option to `cwl-runner` which turns off the `--read-only=true` option for `docker run` which sometimes cause a problem for java which attempts to write in root directories. Tested.


tetron:
Thanks!

Could you give an example of a Java program that fails like this?

mr-c:
Jenkins, okay to test

SooLee:
Got a java error as below without the `--no-read-only` option, when running a java program called `juicebox pre` (Hi-C data analysis bioinformatics tool) as a 2nd step inside a multi-step cwl workflow.
```
Aug 11, 2017 9:58:34 PM java.util.prefs.FileSystemPreferences$1 run
WARNING: Couldn't create user preferences directory. User preferences are unusable.
Aug 11, 2017 9:58:34 PM java.util.prefs.FileSystemPreferences$1 run
WARNING: java.io.IOException: No such file or directory
Failed to create user directory!
```

tetron:
Okay, so it is trying to write a preferences file.  cwltool sets $HOME but I wonder if Java is trying to look up the user's home directory in /etc/passwd and failing, which is why it is trying to write to root.  Ideally we should fix that bug instead of providing a workaround.

SooLee:
Great summary of the problem! Thanks! How about adding this for now (which may be useful for other purposes as well - it's a rather general solution to anything that needs to write to root) and working on the java problem as the next item? :)

j1z0:
+1 this would be really helpful to get merged in

SooLee:
This java root access problem only happens when used with cwl-runner. We never had an issue with it when using docker directly or on seven bridges. We're using my forked version anyway for now, but if this doesn't get into the official version, we may as well just move on to use another cwl executor eventually. It's a pretty trivial and harmless feature to have which can make big difference for many cases and give users more control and save them a lot of time. i.e. Try to see this PR as more of a feature than a bug fix. The java issue is pretty unrelated to this feature actually, just a small example. It's your choice in the end, if crowd control matters more to you, though I think it's likely that someone else may end up implementing the same thing in a different PR.

tetron:
Jenkins, test this please

PR #532, opened 2017-08-22
using stdout field in cachekey calculation

### LINKED ISSUE #408: --cachedir needs to take `stdout` field into account when calculating the hash

kapilkd13:
using sha of `workflowobj` for creating constant `stdout` value and using that value for cache key calculation.

closes #408


kapilkd13:
@mr-c  Can we merge this.

PR #536, opened 2017-08-28
Remove 'nameroot' and 'nameext' from output object.

tetron:


cwltool/main.py
```diff
@@ -910,6 +910,9 @@ def main(argsl=None,  # type: List[str]
             if out is not None:
 
                 def locToPath(p):
+                    for field in ("path", "nameext", "nameroot", "dirname"):
```
mr-c: Why?

tetron: Because `nameext` and `nameroot` are redundant with `basename`, `dirname` is redundant with `path`, and `path` is either set on the next line or shouldn't be there at all (if the output isn't on the local filesystem).

Having `nameext` and `nameroot` in the output was introduced by accident, and it affects job caching (the extra fields mean the output does not exactly match up.)

mr-c: Got it

Issue #544, opened 2017-09-25
Issue with copying directories from cache if output already exists


jasper1918:
## Expected Behavior
While copying from cache to the output directory, files and directories should either be copied or compared with existing outputs. Not all functions allow overwriting so removing the object first may be necessary.

## Actual Behavior
Files are being handled correctly but directories cause an error to be thrown.  In this case the destination directory should be removed first.

## Workflow Code
```
Paste the template code (ideally a minimal example) that causes the issue

```

## Full Traceback
```pytb
 
Copying /home/cwl_cache/6075a780b7f8b08884e3b012b0382c7c/mysample2_salmon_quant to /home/CWL_Pipelines/workflows/RNASEQ_Workflow/test_data/out/mysample2_salmon_quant
Unhandled error, try again with --debug for more information:
  [Errno 17] File exists: '/home/CWL_Pipelines/workflows/RNASEQ_Workflow/test_data/out/mysample2_salmon_quant'
Traceback (most recent call last):
  File "/home/anaconda2/lib/python2.7/site-packages/cwltool/main.py", line 907, in main
    **vars(args))
  File "/home/anaconda2/lib/python2.7/site-packages/cwltool/main.py", line 294, in single_job_executor
    kwargs["make_fs_access"](""))
  File "/home/anaconda2/lib/python2.7/site-packages/cwltool/process.py", line 289, in relocateOutputs
    stageFiles(pm, stageFunc=moveIt, symLink=False)
  File "/home/anaconda2/lib/python2.7/site-packages/cwltool/process.py", line 229, in stageFiles
    stageFunc(p.resolved, p.target)
  File "/home/anaconda2/lib/python2.7/site-packages/cwltool/process.py", line 282, in moveIt
    shutil.copytree(src, dst)
  File "/home/anaconda2/lib/python2.7/shutil.py", line 177, in copytree
    os.makedirs(dst)
  File "/home/anaconda2/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 17] File exists: '/home/CWL_Pipelines/workflows/RNASEQ_Workflow/test_data/out/mysample2_salmon_quant'
```

## Your Environment
cwltool 1.0.20170828135420

mr-c:
Thank you @jasper1918 I agree that this would be a good enhancement

jasper1918:
submitted a PR to address this

mr-c:
[...]

PR #545, opened 2017-09-26
remove dir before copytree, use copy2 to keep metadata

jasper1918:
Relates to https://github.com/common-workflow-language/cwltool/issues/544

This PR adds:
 * remove an existing directory before copying to avoid errors
 * replace copy with copy2 for files when copying in order to preserve metadata. copytree uses copy2. This way all metadata is consistent for files and dirs.

### DIFF
--- a/cwltool/process.py
+++ b/cwltool/process.py
@@ -279,7 +279,7 @@ def relocateOutputs(outputObj, outdir, output_dirs, action, fs_access):
         if src != dst:
             _logger.debug("Copying %s to %s", src, dst)
             if os.path.isdir(src):
-                if os.path.exists(dst):
+                if os.path.isdir(dst):
                     shutil.rmtree(dst)
                 shutil.copytree(src, dst)
             else:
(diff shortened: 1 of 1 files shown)


tetron:
Thanks @jasper1918 !

tetron:
Jenkins, test this please

cwltool/process.py
```diff
@@ -279,9 +279,11 @@ def moveIt(src, dst):
         if src != dst:
             _logger.debug("Copying %s to %s", src, dst)
             if os.path.isdir(src):
+                if os.path.exists(dst):
+                    shutil.rmtree(dst)
```
tetron: Could you add a check if `dst` is a directory.  If it isn't a directory, it should use os.unlink() instead.  (If you try to use `shutil.rmtree` on a regular file it will raise an exception).

jasper1918: Thanks @tetron. The isdir function is prob more appropriate in this case anyway. Not sure I understand the scenario where a src dir could equate to a dst file. The copy2 func will allow overwrites so no need to unlink in the else block.

tetron: If `src` is a directory but `dst` is a file, `copytree()` will fail.  I believe the problem you're trying to solve is correctly overwriting files in the final directory.

jasper1918: That's fair. Committed for completeness. ;)

tetron:
Jenkins, test this please

PR #566, opened 2017-11-14
Add parallel execution

### LINKED ISSUE #565: support parallel thread execution
It would be nice to support parallel thread execution, and I would be happy to contribute to that. How big of an effort do you guys think this would be? Suggestions welcome :)

itajaja:
close #565 

Just a WIP, but seems to be working for me

- [x] test code changes with toil-cwl-runner


itajaja:
for some reason there is one test `test_examples.test_partial_output` that fails...

itajaja:
it's actually flaky, but I really don't know what I am missing

itajaja:
it has to do with the order in which the three parallel jobs are called. in some cases, the failing job gets executed before the others, so the out callback gets called without the outputs from the non-failing one. Not sure what's the best course of action here

kapilkd13:
@itajaja as per travis, our tests are passing on python 3.3 but fails on other versions ( strange). As you suggested it may be because of order of execution of jobs. Just out of curiosity I want to know why have we acquired the lock twice, wouldn't it block itself?

itajaja:
there is nothing special with 3.3, it's just that the test is flaky and it passes depending on which steps finishes first, which randomly varies. Acquiring the lock twice allows to "mimick" a channel pattern. it doesn't deadlock because eventually a job will finish and call `release()` [here](https://github.com/common-workflow-language/cwltool/pull/566/files#diff-7e9ba38c5e8dd21caad9558b5d833a79R295)

tetron:
Wow, great turnaround!

You need to put a mutex around the invocation of output_callback:

https://github.com/common-workflow-language/cwltool/blob/74577663b4e8d19275b68953c9824160b7e25aad/cwltool/job.py#L268

This is because `run()` is in its own thread, so it could race with the workflow excution or callbacks of other processes.

Similarly, you need to have the mutex acquired when calling the generator.

tetron:
You'll also want to run the conformance tests.

```
git clone https://github.com/common-workflow-language/common-workflow-language
cd common-workflow-language
./run-tests.sh RUNNER=cwltool
```

itajaja:
> Similarly, you need to have the mutex acquired when calling the generator.

I don't think we need a lock around that because there is only one thread calling the generator.

itajaja:
[here](https://gist.github.com/itajaja/075bfa291676e04a54ed5b1c95e9a672) is the output of the conformance tests. As far as I could test, both 18 and 88 failed against the master branch as well

tetron:
> I don't think we need a lock around that because there is only one thread calling the generator.

When a process ends, it calls an output callback.  This callback goes to the same WorkflowJob object (eg to `receive_output`) that is called when you iterate on the generator.  So if the generator is executing when an output callback happens in a different thread, there will be a race.

itajaja:
I think I understand. the probably means that I have to switch the `for r in jobiter` syntax with something that uses `next(jobiter)`, correct?

tetron:
It's up to you.  Here's how arvados-cwl-runner does it:

```
            self.cond.acquire()
            # Will continue to hold the lock for the duration of this code
            # except when in cond.wait(), at which point on_message can update
            # job state and process output callbacks.

            for runnable in jobiter:
                if self.stop_polling.is_set():
                    break

                if runnable:
                    runnable.run(**kwargs)
                else:
                    if self.processes:
                        self.cond.wait(1)
                    else:
                        logger.error("Workflow is deadlocked, no runnable jobs and not waiting on any pending jobs.")
                        break
```

itajaja:
sorry, but I am lost, I am not sure what I actually need to lock :/ is it the `jobiter` and something else in another file?

tetron:
@itajaja yes, the lock needs to happen in the callback in job.py line 268, I linked it in my previous comment:

https://github.com/common-workflow-language/cwltool/pull/566#issuecomment-344462708

tetron:
@anton-khodak glad to see you working on this!

cwltool/main.py
```diff
@@ -63,7 +65,8 @@ def arg_parser():  # type: () -> argparse.ArgumentParser
     parser.add_argument("--no-container", action="store_false", default=True,
                         help="Do not execute jobs in a Docker container, even when specified by the CommandLineTool",
                         dest="use_container")
-
+    parser.add_argument("--parallel", action="store_true", default=False,
+                        help="Run jobs in parallel")
```
mr-c: "Experimental: Run jobs in parallel. Does not currently take ResourceRequirments into account."

mr-c:
Hey @anton-khodak , here is the result of running this with Toil
```
  File "/home/michael/src/toil/src/toil/worker.py", line 309, in workerScript
    job._runner(jobGraph=jobGraph, jobStore=jobStore, fileStore=fileStore)
  File "/home/michael/src/toil/src/toil/job.py", line 1328, in _runner
    returnValues = self._run(jobGraph, fileStore)
  File "/home/michael/src/toil/src/toil/job.py", line 1273, in _run
    return self.run(fileStore)
  File "/home/michael/src/toil/src/toil/cwl/cwltoil.py", line 472, in run
    (output, status) = cwltool.main.single_job_executor(**opts)
AttributeError: 'module' object has no attribute 'single_job_executor'
ERROR:toil.worker:Exiting the worker because of a failed job on host laptop
```

mr-c:
@anton-khodak That would be the easiest thing to do, yes

cwltool/main.py
anton-khodak: Should we mark it as experimental? Since we warn that it does not take ResourceRequirements into account, is there any other behavior one would expect from this option that it doesn't provide?

mr-c: Since it could leads to dozens or hundreds of simultaneously tool executions on someone's laptop, then it really should be labeled as being experimental or dangerous

Maybe something more explicit like

> Experimental: Run jobs in parallel. Does not currently keep track of ResourceRequirments like the number of cores or memory and could overload this system.

would be better

mr-c:
Jenkins, okay to test

mr-c:
Thank you @itajaja and @anton-khodak !

itajaja:
glad to see this merged :)

PR #569, opened 2017-11-20
One installs from PyPI or GitHub, not both

mr-c:


tetron:
Apparently a new version of `diff-quality` was released which doesn't support `pep8` for documentation quality checking (?) but you're supposed to use `pycodestyle` instead.

PR #601, opened 2018-01-09
Fix typo in dependency-resolvers-conf.yml

khillion:
Hi, I noticed that the example given to use module thanks to the dependency resolver did not work. It was raising a `KeyError: 'module'` so I fixed that in the `dependency-resolvers-conf.yml`. Thanks !


mr-c:
Thank you @khillion !

PR #605, opened 2018-01-11
Don't complain MultipleInputRequirement if source is single-item list.

tetron:


tetron:
Fix so cwltool runs files created by Rabix Composer.

PR #613, opened 2018-01-17
Short circuit google storage schemas during validation.

danbills:
We at the Broad Institute are using cwltool's pre-processing capability.

We have noticed that Google storage links cannot pass the `check_exists` function as they are by nature not publicly accessible.

What I've submitted here is a quick workaround to get over this hurdle.  I've tested it locally and it is working as intended.

One may wonder why we aren't using `schema-salad-tool` directly to do the preprocessing.  We've noticed that `cwltool` seems to handle a larger variety of cases that `schema-salad-tool` does not, however I do not have such comparisons in hand.

We're definitely open to more comprehensive discussion on a parameter to disable this sort of validation (perhaps only applicable during `--print-pre` situations?).

### DIFF
--- a/cwltool/schemas/v1.0/salad/schema_salad/ref_resolver.py
+++ b/cwltool/schemas/v1.0/salad/schema_salad/ref_resolver.py
@@ -155,7 +155,7 @@ class DefaultFetcher(Fetcher):
                 else:
                     raise RuntimeError('Error reading %s: %s' % (url, e))
         else:
-            raise ValueError('Unsupported scheme in url: %s' % url)
+            raise ValueError('Unsupported scheme in url: %s for scheme %s' % (url, scheme))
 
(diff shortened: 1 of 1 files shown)


mr-c:
@danbills If this is part of pre processing, then perhaps the code should be separate from the reference runner?

If we need to refactor to make that more straight forward, then that is fine by me

tetron:
This is patching the wrong thing, it needs to be submitted to https://github.com/common-workflow-language/schema_salad (because the copy of the code here in cwltool is _not_ actually the code used by cwltool)

tetron:
That said, I would be fine with adding an option to schema-salad to specify URI schemes which are opaque but should not generate an error.

danbills:
Closing in favor of a PR to schema salad repo, thanks for your input!

PR #616, opened 2018-01-19
Update README.rst

rmccole:
I added information about additional things the user may want to have installed, as listed at http://www.commonwl.org/user_guide/setup/. In my hands, running the basic tests without node.js installed resulted in a few failures.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -131,9 +131,11 @@ Remember, if co-installing multiple CWL implementations then you need to
 maintain which implementation ``cwl-runner`` points to via a symbolic file
 system link or `another facility <https://wiki.debian.org/DebianAlternatives>`_.
 
-=====
-Usage
-=====
+You may also want to have the following installed: 
+node.js
+Docker, udocker, or Singularity (optional)
+
+Without these, some examples in the CWL tutorials at http://www.commonwl.org/user_guide/ may not work. 
 
 Run on the command line
 -----------------------


README.rst
```diff
@@ -57,6 +57,13 @@ Remember, if co-installing multiple CWL implementations then you need to
 maintain which implementation ``cwl-runner`` points to via a symbolic file
 system link or `another facility <https://wiki.debian.org/DebianAlternatives>`_.
 
+You may also want to have the following installed: 
+node.js
+Java compiler
+Docker (optional)
```
mr-c: Hmm.. These are already listed at http://www.commonwl.org/user_guide/setup/, so I don't know if we need to duplicate them here?

README.rst
```diff
@@ -57,6 +57,13 @@ Remember, if co-installing multiple CWL implementations then you need to
...
+
+Without these, some examples in the CWL tutorials at http://www.commonwl.org/user_guide/ may not work.
```
mr-c: Is this PR about running the CWL tutorials or the built in tests?

README.rst
rmccole: It's up to you. Most of the installation steps are already duplicated between  http://www.commonwl.org/user_guide/setup/ and the readme. These extras are the only information not already duplicated in both places.

Plus, I was under the impression that some of the tests were failing without node.js, so a new user might look for any other things to install that might be necessary and allow the tests to complete. I was trying to make it easier for that user to find such information. If there was a separate reason for the tests to fail, then this would not be relevant.

Issue #553, opened 2017-10-24
unsupported operand type(s) for +: 'NoneType' and 'str'


gijzelaerr:
```
$ .virtualenv/bin/cwltool ndppp_prep_cal.cwl --msin data/L591513_SB000_uv.MS
.virtualenv/bin/cwltool 1.0.20171017195544
Resolved 'ndppp_prep_cal.cwl' to 'file:///home/gijs/prefactor-cwl/ndppp_prep_cal.cwl'
Got workflow error
Traceback (most recent call last):
start of ndppp_prep_cal step
  File "/home/gijs/prefactor-cwl/.virtualenv/local/lib/python2.7/site-packages/cwltool/main.py", line 283, in single_job_executor
    for r in jobiter:
  File "/home/gijs/prefactor-cwl/.virtualenv/local/lib/python2.7/site-packages/cwltool/draft2tool.py", line 496, in job
    j.command_line = flatten(list(map(builder.generate_arg, builder.bindings)))
  File "/home/gijs/prefactor-cwl/.virtualenv/local/lib/python2.7/site-packages/cwltool/builder.py", line 243, in generate_arg
    args.append(prefix + self.tostr(j))
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
Workflow error, try again with --debug for more information:
unsupported operand type(s) for +: 'NoneType' and 'str'
```

CWL script:

https://github.com/gijzelaerr/prefactor-cwl/blob/393a99922fe84e161df1973bb30607295c0f9672/ndppp_prep_cal.cwl

mr-c:
While the CWL description is wrong, the error experienced is worse and should be fixed. Thank you @gijzelaerr for the report

Issue #555, opened 2017-10-30
AttributeError: 'NoneType' object has no attribute 'get'


gijzelaerr:
```
$ .virtualenv/bin/cwltool --debug ndppp_prep_cal.cwl --msin data/L591513_SB000_uv.MS
.virtualenv/bin/cwltool 1.0.20171017195544
Resolved 'ndppp_prep_cal.cwl' to 'file:///home/gijs/prefactor-cwl/ndppp_prep_cal.cwl'
I'm sorry, I couldn't load this CWL file.  The error was:
Traceback (most recent call last):
  File "/home/gijs/prefactor-cwl/.virtualenv/local/lib/python2.7/site-packages/cwltool/main.py", line 818, in main
    skip_schemas=args.skip_schemas)
  File "/home/gijs/prefactor-cwl/.virtualenv/local/lib/python2.7/site-packages/cwltool/load_tool.py", line 214, in validate_document
    _convert_stdstreams_to_files(workflowobj)
  File "/home/gijs/prefactor-cwl/.virtualenv/local/lib/python2.7/site-packages/cwltool/load_tool.py", line 85, in _convert_stdstreams_to_files
    if out.get('type') == streamtype:
AttributeError: 'NoneType' object has no attribute 'get'
```

CWL file
https://github.com/gijzelaerr/prefactor-cwl/blob/master/ndppp_prep_cal.cwl

mr-c:
I've update the original post to reference to the exact file that cause the unhelpful error

PR #635, opened 2018-01-31
Fix type mismatches that caused bad error messages

anton-khodak:
closes https://github.com/common-workflow-language/cwltool/issues/553, #555


cwltool/load_tool.py
```diff
@@ -124,6 +124,8 @@ def _convert_stdstreams_to_files(workflowobj):
     if isinstance(workflowobj, dict):
         if workflowobj.get('class') == 'CommandLineTool':
             for out in workflowobj.get('outputs', []):
+                if type(out) is not CommentedMap:
+                    raise ValidationException("Output '%s' is not valid" % out)
```
mr-c: the error message becomes:
```
$ cwltool --debug https://raw.githubusercontent.com/EOSC-LOFAR/prefactor-cwl/c06f92eb9cea4ab34cfffcbcc48c5d32aad2a1a7/ndppp_prep_cal.cwl --help
/home/michael/cwltool/env/bin/cwltool 1.0.20180201074953
Tool definition failed validation:
Output 'None' is not valid
  File "/home/michael/cwltool/cwltool/main.py", line 873, in main
    overrides=overrides)
  File "/home/michael/cwltool/cwltool/load_tool.py", line 269, in validate_document
    _convert_stdstreams_to_files(workflowobj)
  File "/home/michael/cwltool/cwltool/load_tool.py", line 128, in _convert_stdstreams_to_files
    raise ValidationException("Output '%s' is not valid" % out)
ValidationException: Output 'None' is not valid
```
which is still not helpful to the user, can you add the line number using the same technique from https://github.com/common-workflow-language/cwltool/pull/635/files#diff-bfbf952b5c2585de423b12dcf6f43938R210 ?

tests/test_examples.py
```diff
@@ -514,6 +514,19 @@ def test_lifting(self):
...
+    def test_malformed_outputs(self):
+        # check that tool validation fails if one of the outputs is not a valid CWL type
+        f = cwltool.factory.Factory()
+        with self.assertRaises(schema_salad.validate.ValidationException):
+            echo = f.make(get_data("tests/wf/malformed_outputs.cwl"))
+            echo()
+
+    def test_separate_without_prefix(self):
+        # check that setting 'separate = false' on an inputBinding without prefix does not fail the workflow
+        f = cwltool.factory.Factory()
```
mr-c: I disagree, that should fail the workflow as it is nonsensical.

PR #649, opened 2018-02-08
implement logging timestamps and record container ID in a file

MichaelTong:
Added options:
```
--record-container-id
                      If enabled, a file with suffix ".cid" will be created
                      storing the container ID under CIDFILE_DIR
--cidfile-dir CIDFILE_DIR
                      Directory for storing cidfiles. Default at /tmp/
--cidfile-prefix CIDFILE_PREFIX
                      Give a prefix to cidfile. Final file name will be
                      followed by a timestamp. Default empty.
--logtstp             Print timestamps with logging
```

### DIFF
--- a/cwltool/job.py
+++ b/cwltool/job.py
@@ -364,7 +364,7 @@ class DockerCommandLineJob(JobBase):
                         docker_windows_path_adjust(vol.target)))
 
     def run(self, pull_image=True, rm_container=True,
-            record_container_id=False, cidfile_dir="/tmp/",
+            record_container_id=False, cidfile_dir="",
             cidfile_prefix="",
             rm_tmpdir=True, move_outputs="move", **kwargs):
         # type: (bool, bool, bool, Text, Text, bool, Text, **Any) -> None
@@ -469,21 +469,22 @@ class DockerCommandLineJob(JobBase):
         
         # add parameters to docker to write a container ID file
         if record_container_id:
-            if cidfile_dir:
+            if cidfile_dir != "":
                 if not os.path.isdir(cidfile_dir):
                     _logger.error("--cidfile-dir %s error:\n%s", cidfile_dir,
-                                  cidfile_dir + " is not a directory or directory doesn't exist, please check it first")
-                    exit(-1)
+                                  cidfile_dir + " is not a directory or "
+                                  "directory doesn't exist, please check it first")
+                    exit(2)
                 if not os.path.exists(cidfile_dir):
                     _logger.error("--cidfile-dir %s error:\n%s", cidfile_dir,
                                   "directory doesn't exist, please create it first")
-                    exit(-1)
+                    exit(2)
             else:
-                cidfile_dir = "/tmp/"
+                cidfile_dir = os.getcwd()
             cidfile_name = datetime.datetime.now().strftime("%Y%m%d%H%M%S-%f")+".cid"
             if cidfile_prefix != "":
                 cidfile_name = str(cidfile_prefix + "-" + cidfile_name)
-            cidfile_path = cidfile_dir + cidfile_name
+            cidfile_path = os.path.join(cidfile_dir, cidfile_name)
             runtime.append(u"--cidfile=%s" % cidfile_path)
 
         for t, v in self.environment.items():
--- a/cwltool/job.py
+++ b/cwltool/job.py
@@ -482,7 +482,7 @@ class DockerCommandLineJob(JobBase):
                 cidfile_dir = "/tmp/"
             cidfile_name = datetime.datetime.now().strftime("%Y%m%d%H%M%S-%f")+".cid"
             if cidfile_prefix != "":
-                cidfile_name = cidfile_prefix + "-" + cidfile_name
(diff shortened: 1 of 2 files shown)


mr-c:
Jenkins, okay to test

MichaelTong:
Not sure why failed Jenkins... Console Output didn't give much information..

cwltool/main.py
```diff
@@ -84,6 +84,20 @@ def arg_parser():  # type: () -> argparse.ArgumentParser
                          default=True, help="Do not delete Docker container used by jobs after they exit",
                          dest="rm_container")
 
+    parser.add_argument("--record-container-id", action="store_true", default=False,
+                        help="If enabled, a file with suffix \".cid\" will be created storing the container ID under CIDFILE_DIR",
+                        dest="record_container_id")
+
+    parser.add_argument("--cidfile-dir", type=Text,
+                        help="Directory for storing cidfiles. Default at /tmp/",
+                        default="/tmp/",
```
mr-c: I think the current directory is a better default and is more portable.

cwltool/main.py
```diff
@@ -84,6 +84,20 @@ def arg_parser():  # type: () -> argparse.ArgumentParser
                          default=True, help="Do not delete Docker container used by jobs after they exit",
                          dest="rm_container")
 
+    parser.add_argument("--record-container-id", action="store_true", default=False,
+                        help="If enabled, a file with suffix \".cid\" will be created storing the container ID under CIDFILE_DIR",
```
mr-c: Can you linewrap at 80 for easier reviews?

mr-c: "storing the container ID under the directory specified by --cidfile-dir"

mr-c: These new options would be best to put in their own https://docs.python.org/2.7/library/argparse.html#argument-groups. You can use that help to explain what a "cidfile" is.

cwltool/main.py
```diff
@@ -161,6 +175,7 @@ def arg_parser():  # type: () -> argparse.ArgumentParser
     exgroup.add_argument("--verbose", action="store_true", help="Default logging")
     exgroup.add_argument("--quiet", action="store_true", help="Only print warnings and errors.")
     exgroup.add_argument("--debug", action="store_true", help="Print even more logging")
+    parser.add_argument("--logtstp", action="store_true", help="Print timestamps with logging")
```
mr-c: I think `--timestamps` would be a better name and "Add timestamps to the errors, warnings, and notifications."

cwltool/job.py
```diff
@@ -463,6 +466,25 @@ def run(self, pull_image=True, rm_container=True,
...
         # runtime.append("--env=HOME=/tmp")
         runtime.append(u"--env=HOME=%s" % self.builder.outdir)
+        
+        # add parameters to docker to write a container ID file
+        if record_container_id:
+            if cidfile_dir:
+                if not os.path.isdir(cidfile_dir):
+                    _logger.error("--cidfile-dir %s error:\n%s", cidfile_dir,
+                                  cidfile_dir + " is not a directory or directory doesn't exist, please check it first")
+                    exit(-1)
```
mr-c: Should be `exit(2)`: https://docs.python.org/2/library/sys.html#sys.exit

cwltool/job.py
```diff
@@ -463,6 +466,25 @@ def run(self, pull_image=True, rm_container=True,
...
+                if not os.path.exists(cidfile_dir):
+                    _logger.error("--cidfile-dir %s error:\n%s", cidfile_dir,
+                                  "directory doesn't exist, please create it first")
+                    exit(-1)
```
mr-c: Should be `exit(2)`: https://docs.python.org/2/library/sys.html#sys.exit

cwltool/job.py
```diff
@@ -463,6 +466,25 @@ def run(self, pull_image=True, rm_container=True,
...
+            else:
+                cidfile_dir = "/tmp/"
+            cidfile_name = datetime.datetime.now().strftime("%Y%m%d%H%M%S-%f")+".cid"
+            if cidfile_prefix != "":
+                cidfile_name = str(cidfile_prefix + "-" + cidfile_name)
+            cidfile_path = cidfile_dir + cidfile_name
+            runtime.append(u"--cidfile=%s" % cidfile_path)
```
mr-c: Is there an equivalent for Singularity?

mr-c:
Jenkins, test this please

mr-c:
Jenkins, test this please

cwltool/job.py
MichaelTong: Sorry but I don't think I understand this. Can you explain it a little bit.
I have finished other suggested changes

mr-c: http://singularity.lbl.gov/

If this option only works for the Docker runtime then we should say so. If we can add the equivalent option for Singularity then that would be nice.

MichaelTong: As far as I know, it comes with docker. I didn't know singularity before, but by reading its documents, I believe it's possible to add the equivalent option.

PR #651, opened 2018-02-09
 Add default command line arguments for cwltool as Factory 

### LINKED ISSUE #596: Set default arguments in cwltool.factory.Factory
In the README, the recommended way to use cwltool as a factory is to use the `cwltool.factory.Factory` class (https://github.com/common-workflow-language/cwltool#import-as-a-module). However, this doesn't set default command line options - these are normally set in the command line using argparse in main by: https://github.com/common-workflow-language/cwltool/blob/f691bf1aa66f2836694716c9845731962147079b/cwltool/main.py#L756 If no default command line options are passed when instantiating the Factory, then this produces lots of unintended effects (e.g. use_container is False)

anton-khodak:
and moving all command line parsing utilities from `main.py` to a new file `argparser.py`.
Closes #596

### DIFF
--- a/cwltool/argparser.py
+++ b/cwltool/argparser.py
@@ -217,6 +217,16 @@ def arg_parser():  # type: () -> argparse.ArgumentParser
     return parser
 
 
+def get_default_args():
+    # type: () -> Dict[str, Any]
+    """
+    Get default values of cwltool's command line options
+    """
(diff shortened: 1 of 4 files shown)


mr-c:
Thanks! I confirmed that this didn't break `toil-cwl-runner`.

PR #656, opened 2018-02-13
Fix --skip-schemas default.

tetron:
Wrong default meant that schemas were _never_ being used, resulting in warnings.

Need a separate schema-salad fix to suppress warnings when skip_schemas is in effect.


tetron:
See https://github.com/common-workflow-language/schema_salad/pull/152 for related warning fix.

PR #659, opened 2018-02-14
Fix duplicated logging, only initialize defaultStreamHandler once.

tetron:


anton-khodak:
Thank you, @tetron, my bad!

PR #661, opened 2018-02-15
Single quote instead of backticks in output

mr-c:
While backticks are good for reStructuredText Docstring formatting
(PEP-0287) they aren't appropriate for printing to the user


mr-c:
Jenkins, test this please

PR #664, opened 2018-02-20
Fix eval_timeout in workflow

philres:
I noticed that the `evl_timeout` cli parameter is not passed to `expression.do_eval` in `workflow.py`. This caused problems for me when running cwltool on a busy machine.


mr-c:
@philres Thank you!

Jenkins, okay to test

Karhide:
This would improve my life immeasurably

PR #683, opened 2018-03-05
Experimental extension for "secret" parameters.

tetron:
These are string input parameters that represent sensitive information
such as passwords.  This feature deliberately obscures them by
replacing them with a placeholder during workflow evaluation and then
substituting the real value at the last moment.


tetron:
Whoops, I actually did add write some tests and forgot to commit the file.

tetron:
Also any initial work dir files containing secrets needs to be deleted immediately after the run finishes so it doesn't hang around.

PR #619, opened 2018-01-23
Update Dockerfile to use official docker image instead of dind

tom-tan:
It is redone of #156.
#156 is not merged because it depends on [dind]( https://github.com/jpetazzo/dind) that is already obsolete.

This request updates Dockerfile to use official [`docker`](https://hub.docker.com/_/docker/) image that includes docker-in-docker feature.

Here are the list of detailed changes.
- I remove the contents of `cwltool_module.Dockerfile` because dind feature in `cwltool_module.Dockerfile` is included in the official `docker` image.
- I update the contents of `cwltool.Dockerfile` for official `docker` image, which is based on Alpine Linux. I also rename `cwltool.Dockerfile` to `Dockerfile`.
- I replace `MAINTAINER` with `LABEL maintainer` because `MAINTAINER` is [deprecated](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated).

Here is the running example.
```yml
# echo.cwl
class: CommandLineTool
cwlVersion: v1.0
id: echo_cwl
baseCommand:
  - cowsay
inputs:
  - id: input
    type: string
    inputBinding:
      position: 0
outputs:
  output:
    type: stdout
stdout: output
requirements:
  - class: DockerRequirement
    dockerPull: docker/whalesay
```

```console
$ docker build -t cwltool .
# (snip)
$ docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v $PWD:/tmp/workdir cwltool --outdir /tmp/workdir /tmp/workdir/echo.cwl --input 'Hello CWL!'
/usr/bin/cwltool 1.0
Resolved '/tmp/workdir/echo.cwl' to 'file:///tmp/workdir/echo.cwl'
[job echo.cwl] /tmp/tmpaOnsGY$ docker \
    run \
    -i \
    --volume=/tmp/tmpaOnsGY:/var/spool/cwl:rw \
    --volume=/tmp/tmpVmWRtN:/tmp:rw \
    --workdir=/var/spool/cwl \
    --read-only=true \
    --log-driver=none \
    --user=0:0 \
    --rm \
    --env=TMPDIR=/tmp \
    --env=HOME=/var/spool/cwl \
    docker/whalesay \
    cowsay \
    'Hello CWL!' > /tmp/tmpaOnsGY/output
[job echo.cwl] completed success
{
    "output": {
        "checksum": "sha1$74be426e67ee8d8e437116e1f4aed9af1014b7ac",
        "basename": "output",
[...]

PR #690, opened 2018-03-16
Stop using Docker in Docker for container image

undu:
Docker in docker has been frowned upon for quite some time since it requires elevated privileges: http://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/
Furthermore, it didn't work with newer versions as the wrapper script was giving errors (problems about mounting special folders, meaning it could not connect to the docker daemon)
The image was also missing nodejs, so the inline expressions were failing.

We can avoid dind :)
The only caveat is that a new mountpoint needs to be define between the host and the cwltool container, so cwltool has visibility of the files produced by the tools. (usually /tmp)

This changes allow us to drop one of the two images that were being used, as well as the wrapper script.
I also took it upon myself to use multi-stage dockerfiles to try and reduce the image file.
Now the base image is Alpine instead of Ubuntu and has nodejs included.
These changes mean the image is now 227 MB instead of the hefty 781MB

What would be missing still would be to check if the shell scripts still work, and changing the documentation to update it as this new image could mean braking some use-cases.

I almost forgot: I dodn't see a regression, but apparently the libc used  in alpine can be slower than glibc, if it's a problem I can try making a build from a Debian base.


undu:
I just saw https://github.com/common-workflow-language/cwltool/pull/619 I guess this one supersedes it, depending on whether we want to trade the --privileged for the additional mountpoint or not.

tetron:
So #619 uses an official image but otherwise continues to use nested Docker, which requires the outer container to be privileged.  This branch instead mounts the host's Docker socket, which has a different set of tradeoffs.

What's your use case for running cwltool in Docker?

tom-tan:
> What's your use case for running cwltool in Docker?

It is the most simplest way to try cwltool.
It makes easier to try cwltool because we can use it without installing its dependencies.

By the way, I do not have a strong opinion about whether we will use an official d-in-d container or mount host's docker socket to achieve docker-in-docker. It will be fine for me if the docker image for cwltool (with version tag, I hope) is available in Dockerhub.

undu:
>What's your use case for running cwltool in Docker?

I want to run cwltool in gitlab's CI environment, which uses docker for the most part.

The currrent container does work because the script is [broken](https://gitlab.com/unduthegun/stellaris-emblem-lab/-/jobs/39531992).

The way I've made it work in there is less than optimal, as I need to use docker in docker: https://gitlab.com/unduthegun/stellaris-emblem-lab/blob/aca472a8f0b95118e227f98fc0d550ad651ff631/.gitlab-ci.yml

tetron:
Okay, we'll go with mounting the Docker socket.  Both solutions have the drawback that the container has to be invoked with additional parameters in order to function correctly.

@undu can you add an entrypoint script which checks if the container was invoked correctly (for example checks that the docker socket is available) and if not, prints out an error telling the user what to do?

(another check you could do is to make the container's default working directory something like `/error` and then check that the actual cwd when the container runs is not `/error`).

build-cwl-docker.sh
```diff
@@ -1,4 +1,3 @@
 #!/bin/sh
 set -e
-docker build --file=cwltool_module.Dockerfile --tag=commonworkflowlanguage/cwltool_module .
```
tetron: Can you restore the cwltool_module Docker image?  Some things actually do rely on it.  Its purpose is to provide an environment for running schema salad and cwltool for operations that don't require docker.

undu: Would the new image with `sh` as an entrypoint work for this purpose?

tetron: Yes, I think that would work.  The cwltool_module image is used to run "python -mschema_salad.makedoc", the key difference is that cwltool_module shouldn't have `WORKDIR /error` and `ENTRYPOINT`.  The simple solution is to layer the cwltool image on top of cwltool_module.

undu: Decided to build cwltool upon cwltool_module, just like it was before.

Let me know it this change is uncomfortable because some parameters used in the command that builds the cwltool_module images changed a bit.

build-cwl-docker.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/sh
 set -e
+docker build --file=cwltool.Dockerfile --tag=commonworkflowlanguage/cwltool-module --target module .
```
tetron: I think this is supposed to be "--target builder" or `FROM python:3.6-alpine as builder` should be `FROM python:3.6-alpine as module` ?

tetron:
Ok, last comment, please replace update the contents of `cwl-docker.sh` with this:

```sh
#!/bin/sh
exec docker run -v /var/run/docker.sock:/var/run/docker.sock -v /tmp:/tmp -v "$PWD":"$PWD" -w="$PWD" commonworkflowlanguage/cwltool "$@"
```

PR #526, opened 2017-08-15
[WIP] Config to setup coveralls

manu-chroma:
Use coveralls will report code coverage on the PR.
This will only measure coverage using unit tests for the time being.

PR #691, opened 2018-03-16
Coverage reporting

psafont:
This adds coverage reporting and uploading to both coveralls and codecov, any of which can be easily disabled.

A common-workflow-language admin should give access to codecov or coveralls to enable uploads to them.

This PR supersedes https://github.com/common-workflow-language/cwltool/pull/526


mr-c:
Can you also look at the Makefile which has some related targets? It is good to enable local feedback too. Thanks!

psafont:
What is the functionality that is needed in the Makefile?
Generate coverage of all the tests on the default python version and generate a report?

mr-c:
Yes, which might already exist but should be documented in our non-existent developer guide 😋

Makefile
```diff
@@ -113,23 +113,10 @@ pylint_report.txt: ${PYSOURCES}
...
-	coverage combine ${CWL} ${CWL}/draft-3/ ./
+.coverage: tests
 
-coverage.xml: .coverage
-	python-coverage xml
-
-coverage.html: htmlcov/index.html
-
-htmlcov/index.html: .coverage
-	python-coverage html
```
mr-c: can we keep the HTML report?

psafont: In retrospect, we can also keep the xml report

manu-chroma:
this is awesome!

PR #694, opened 2018-03-20
V1.1.0-dev1 support

tetron:


tetron:
Tracking v1.1 spec work, merge when first v1.1 milestone is ready (should be soon).

PR #695, opened 2018-03-20
Added support for Singularity environment variables

uniqueg:
To allow pre-fetching of images in cluster scenarios where workers don't have network access (such as ours), I have modified the Singularity module of the CWL reference runner such that it now checks for available images in `$SINGULARITY_CACHEDIR` (if set), `$SINGULARITY_PULLFOLDER` (if set) and the current working directory, in that order. This largely mirrors Singularity's default behavior as per http://singularity.lbl.gov/docs-pull#pull-to-different-folder.

Note that without the suggested changes, only the current working directory is scanned for available images. However, this may not be of much use because (as far as I can tell) the execution directories are typically generated on-the-fly and their names cannot be predicted, thus making pre-fetching of images and copying them to the required locations prior to workflow execution unfeasible.

## Testing
After setting either of the Singularity environment variables and pulling required images from the login nodes, running the CWL tests via Toil on our Slurm-managed HPC without network access inside Singularity containers now largely works. 5 tests fail for me with Toil execution (remote and locally) and 4 with the CWL reference runner (local execution only). However, the same tests also fail with the unmodified version on the current master branch, so I assume that this is due to my environment. Could someone for whom all tests pass please test this?

Make sure to export `SINGULARITY_PULLFOLDER` or `SINGULARITY_CACHEDIR`:
```sh
export SINGULARITY_PULLFOLDER="$TMP/SINGULARITY_IMAGES"
mkdir -p "$SINGULARITY_PULLFOLDER"
```

Then pre-fetch the images required for the tests:
```sh
singularity pull --name python-2-slim.img docker://python:2-slim
singularity pull --name java-7.img docker://java:7
singularity pull --name alpine.img docker://alpine
singularity pull --name debian-stretch-slim.img docker://debian:stretch-slim
```

Finally run the tests (make sure you have the latest versions of the runners):
```sh
# cwltool
./run_test.sh -j${cores} RUNNER="cwltool" EXTRA='--singularity'
# toil  -  note that not all conformance tests pass for the latest version of toil
./run_test.sh -j${cores} RUNNER="toil-cwl-runner" EXTRA='--singularity'
```

If you have configured Toil to execute commands remotely, add the `--batchSystem` option.

## Discussion points:
- Note that there is one difference with `singularity pull`'s default behavior: Singularity will immediately proceed with trying to pull the desired image from the registry if it finds that the desired image cannot be found in a directory associated with an environment variable that is **set** (i.e. it continues to look in other locations only if the variable is unset). The behavior in the suggest change is more permissive: an attempt is made to pull an image from the registry only if a local image is not found in **either** location. Let me know if you think that we should rather go with the default behavior of `singularity pull`.
- A deep integration of a pre-fetching/copying strategy for container images etc might render the suggested changes and their benefits redundant. But this would likely require considerable changes/time. Plus the reference implementation may not be the right place for such functionality. But please let me know if you are aware of any plans in that direction.


mr-c:
Jenkins, add to test list

tetron:
Jenkins, test this please

PR #697, opened 2018-03-22
Fix for directory literals returned by InitialWorkDir expression.

tetron:


tetron:
Conformance tests added: https://github.com/common-workflow-language/common-workflow-language/pull/657

PR #698, opened 2018-03-22
Don't try to update 'pip' on appveyor

tetron:


mr-c:
Did they fix their problem? The build used to fail without this.

tetron:
Well now the build fails with it, it gets a permission error when it tries to uninstall the old version of pip.

PR #702, opened 2018-03-26
Add Appveyor and codecov badges

psafont:
This change adds 2 badges to the top of the readme, the appveyor one with status of windows builds, and the codecov with coverage info.
It also adds relevant text to the badges and removes the text that was next to them.

I would have changed the CWL conformance one to include the text inside the [badge](https://img.shields.io/jenkins/s/https/ci.commonwl.org/job/cwltool-conformance.svg?label=CWL%20conformance).
Unfortunately shields.io doesn't retrieve the info from Jenkins even though the [API endpoint](https://ci.commonwl.org/job/cwltool-conformance/api/json?tree=color) that should be queried works.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,11 +2,21 @@
 Common Workflow Language tool description reference implementation
 ==================================================================
 
-CWL conformance tests: |Build Status| Travis CI: |Unix Build Status|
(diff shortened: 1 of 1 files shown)


mr-c:
Thanks!

PR #706, opened 2018-04-02
Add flag to skip validation.

tetron:
Option to skip validation (intended for use when validation is redundant, eg a nested execution of cwltool where the workflow has already been validated and doesn't need to be re-validated).


cwltool/argparser.py
```diff
@@ -214,6 +214,8 @@ def arg_parser():  # type: () -> argparse.ArgumentParser
     parser.add_argument("--custom-net", type=Text,
                         help="Will be passed to `docker run` as the '--net' "
                              "parameter. Implies '--enable-net'.")
+    parser.add_argument("--disable-validate", dest="do_validate",
+                        action="store_false", default=True, help="Skip CWL document validation.")
```
mr-c: Maybe a hidden option?

tetron:
jenkins, test this please

Issue #704, opened 2018-03-30
Regression in handling enum type option


partha-edico:
## Expected Behavior
cwltool should handle an input type such as the following:
    type:
    - 'null'
    - type: enum
      symbols: ['across', 'commas','long']

## Actual Behavior
As of release 1.0.20180322194411 (NOTE: Previous release 1.0.20180306163216 worked OK), we get cwltool parsing errors:
=> cwl_test.yaml:13:7: invalid field `name`, expected one of: 'symbols', 'type', 'label', 'inputBinding'

### cwl_tool.yaml
```
baseCommand: /bin/ls
class: CommandLineTool
cwlVersion: v1.0
hints: []
requirements:
- {class: InlineJavascriptRequirement}
inputs:
  ls_format:
    type:
    - 'null'
    - type: enum
      symbols: ['across','commas','long']
    label: Format type
    inputBinding: {position: 3, prefix: --format, itemSeparator: '='}
outputs:
  bam_output:
    type: File?
    outputBinding: {glob: '*.bam'}
stderr: stderr.txt
stdout: stdout.txt
```
### test_input.yaml
```
ls_format: 'long'
```

## Full Traceback
No exception

## Your Environment
* cwltool version: 1.0.20180322194411

tetron:
Is this a fatal error or a warning?

partha-edico:
Hi Peter,
This is not a fatal eror. cwltool keeps running and doesn't exit. However, there seem to be binding errors that have a ripple effect that prevents my full CWL from being properly parsed. The example I gave is a very simplistic one that anyone can run. When I try to run our tool (called "dragen") I get the error below:
```
Resolved '/staging/eTES/working/cwl_tool.yaml' to 'file:///staging/eTES/working/cwl_tool.yaml'
../working/cwl_tool.yaml:108:13: invalid field `name`, expected one of: 'symbols', 'type', 'label', 'inputBinding'
../working/cwl_tool.yaml:108:13: invalid field `name`, expected one of: 'symbols', 'type', 'label', 'inputBinding'
../working/cwl_tool.yaml:121:13: invalid field `name`, expected one of: 'symbols', 'type', 'label', 'inputBinding'
../working/cwl_tool.yaml:121:13: invalid field `name`, expected one of: 'symbols', 'type', 'label', 'inputBinding'
../working/cwl_tool.yaml:133:13: invalid field `name`, expected one of: 'symbols', 'type', 'label', 'inputBinding'
[...]

PR #708, opened 2018-04-02
Suppress spurious "invalid fields `name`" warnings

tetron:
refs #704


cwltool/validate_js.py
```diff
@@ -25,6 +25,17 @@ def is_expression(tool, schema):
     # type: (Union[CommentedMap, Any], avro.schema.Schema) -> bool
     return isinstance(schema, avro.schema.EnumSchema) and schema.name == "Expression" and isinstance(tool, (str, Text))
 
+class SuppressLog(logging.Filter):
+    def __init__(self, name):  # type: (Text, logging.Logger) -> None
```
mr-c: Just `type: (Text) -> None`?

tetron: Whoops, yes, I copied it from somewhere else.

PR #717, opened 2018-04-12
Specify absolute path of config file for dependency resolver

khillion:
Hi all,

The dependency resolver is only using the path of the config file specified by the user. If this path is relative and there is any change of working directory (for instance, `cwltoil` is changing its working directory), the dependency resolver will not be able to find the config file.
I propose a little fix to always transmit the absolute path of the config file to prevent this.

Cheers,
Kenzo


mr-c:
Thank you @khillion !

PR #730, opened 2018-04-26
Fix Python3 compatibility

jsurkont:
cwltool fails when run with --singularity on Python3

Popen returns bytes in Python3, it is necessary to convert it to string
before comparison.


mr-c:
Thank you @jsurkont for this PR. Alas we're having problems running Singularity on ci.commonwl.org right now, so our testing coverage is low.

I've already got a fix for this over at https://github.com/common-workflow-language/cwltool/pull/729 I think I'm going to go with that one as it doesn't presume UTF-8.

PR #731, opened 2018-04-26
Get rid of non-spec-conforming "magic"

tetron:
https://github.com/common-workflow-language/common-workflow-language/issues/675

https://github.com/common-workflow-language/common-workflow-language/issues/679


mr-c:
@tetron Is this PR from 2018 still relevant?

tetron:
This PR should probably be closed. However I looked at 1.2 and I don't think we documented the workaround we ultimately agreed on -- so I think there is another to-do item there.

PR #732, opened 2018-04-30
Fix scandeps with File object inside a seperately declared Directory object

tetron:
When scandeps merges a real directory with a synthetic directory literal, real directory takes precedence.


mr-c:
@tetron Tricky! Can I see a test case to improve my understanding?

PR #733, opened 2018-04-30
Move var/spool/cwl detector to process.py and make it a warning.

tetron:
Do the check alongside other static checks.

Need to decide how best to fix the test that previously assumed this
was a validation error.

### DIFF
--- a/cwltool/process.py
+++ b/cwltool/process.py
@@ -28,6 +28,7 @@ from ruamel.yaml.comments import CommentedMap, CommentedSeq
 from schema_salad.ref_resolver import Loader, file_uri
 from schema_salad.sourceline import SourceLine
 from six.moves import urllib
+from six import iteritems, itervalues, string_types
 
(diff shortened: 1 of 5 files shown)


cwltool/process.py
```diff
@@ -499,6 +521,10 @@ def __init__(self, toolpath_object, **kwargs):
 
             validate_js_expressions(cast(CommentedMap, toolpath_object), self.doc_schema.names[toolpath_object["class"]], validate_js_options)
 
+        dockerReq, is_req = self.get_requirement("DockerRequirement")
+        if not (kwargs.get("use_container") and dockerReq and dockerReq.get("dockerOutputDirectory") == "/var/spool/cwl"):
```
mr-c: I disagree with checking for `use_container`

also: `and dockerReq and is_req …` so that specifying `dockerOutputDirectory` under `hints` isn't sufficient

tetron: good idea

PR #735, opened 2018-05-02
Replace cwltool_module with cwltool-module

tom-tan:
This request fixes docker image name for cwltool module.

Related:  #403


mr-c:
Thanks!

PR #739, opened 2018-05-09
WIP handle state variables in js comments

ionox0:
I haven't tested this, but just wanted to gauge interest in support for some of these STATE variables in the expression tools.

I had a situation like this:
```
${
  // this comment's got a quote in it!
  var asdf = ...
...
```
Which caused:
```
Substitution error, unfinished block starting at position 0: ${
```
I believe because of the quote in the comment. Can add a test or refactor if interested

### DIFF
--- a/cwltool/expression.py
+++ b/cwltool/expression.py
@@ -53,13 +53,27 @@ def scanner(scan):  # type: (Text) -> List[int]
     SINGLE_QUOTE = 4
     DOUBLE_QUOTE = 5
     BACKSLASH = 6
+    COMMENT = 7
 
     i = 0
     stack = [DEFAULT]
     start = 0
+    com_check = ''
     while i < len(scan):
         state = stack[-1]
         c = scan[i]
+
+        if state == COMMENT:
+            if c == '\n':
+                stack.pop()
+                com_check = ''
+            pass
+        com_check += c
+        if com_check == '//':
+            stack.append(COMMENT)
+            pass
+        elif com_check != '/':
+            com_check = ''
 
         if state == DEFAULT:
             if c == '$':


mr-c:
Jenkins, okay to test

ionox0:
I do see this bit here down in `interpolate`, so not sure if I'm seeing the full picture on this one:
```
        elif scan[w[0]] == '\\':
            e = scan[w[1] - 1]
            parts.append(e)
```

cwltool/expression.py
```diff
@@ -53,14 +53,28 @@ def scanner(scan):  # type: (Text) -> List[int]
...
+                com_check = ''
+            pass
+        com_check += c
+        if com_check == '//':
+            stack.append(COMMENT)
+            pass
+        elif com_check != '/':
+            com_check = ''
+
         if state == DEFAULT:
```
mr-c: FYI: This might allow `//` style comments inside CWL paramenter references, which is contrary to the CWL v1.0.x standards.

ionox0: Ah I see didn't realize this was used to parse `valueFrom`'s thought it was just for ExpressionTools. Perhaps I should close this and return when I have a fuller picture of the parsing environment

tetron: This method is used to extract the text inside `${}` but there is a separate regex that determines if the text is a plain parameter reference or arbitrary javascript.  So I think adding comment parsing is okay.

tetron: (The purpose is to parse just enough javascript grammar to be able to recognize where the expression begins and ends)

ionox0: Looks like an issue here:
```
Test failed: /home/jenkins/workspace/cwltool-pr-conformance-multiver/version/v1.0/cwltool-venv2/bin/cwltool --outdir=/tmp/tmppDHNEh --quiet v1.0/test-cwl-out2.cwl v1.0/empty.json
Test support for reading cwl.output.json when running in a Docker container
and just 'location' is provided.
```
Perhaps you point me in the direction of this failing test?

cwltool/expression.py
```diff
@@ -53,14 +53,28 @@ def scanner(scan):  # type: (Text) -> List[int]
...
     while i < len(scan):
         state = stack[-1]
         c = scan[i]
 
+        if state == COMMENT:
+            if c == '\n':
+                stack.pop()
+                com_check = ''
+            pass
+        com_check += c
```
mr-c: error: Incompatible types in assignment (expression has type "unicode", variable has type "str")

cwltool/expression.py
mr-c: Looks like you triggered on the `//` in a `file://` URI:

```
:1:1: Expression evaluation error:
      Substitution error, unfinished block starting at position 0: echo foo > foo && echo '{"foo":
      {"location": "file://$(runtime.outdir)/foo", "class": "File"} }' > cwl.output.json
```

cwltool/expression.py
ionox0: Unfortunately I don't know if I'll actually be able to become familiar enough with the tests as these next few weeks are looking a bit busy for me. Perhaps I can reopen this when I get some time to understand better the system

cwltool/expression.py
tetron: If that's happening, the test for `//` is happening in the wrong parser state.  I'll dig into it when I get a chance.

mr-c: @tetron Can you take a look? Having a full URI in the `format` field is leading to `Dockstore.cwl:45:5: Substitution error, unfinished block starting at position 0: 'http://edamontology.org/data_3671' stack was [0, 7]`

https://github.com/common-workflow-language/cwltool/pull/739/checks?check_run_id=3461661076#step:9:352

PR #762, opened 2018-05-18
Include subprocess32 when building for Mac OS X.

jmchilton:
The code gates imports on os.name == "posix", so the dependency gates should match right? I'm guessing pip installing subprocess32 locally fixing problems means this was likely an issue - but let me know if this doesn't make sense.


mr-c:
Thanks!

PR #768, opened 2018-05-23
Provenance nested

mr-c:


cwltool/command_line_tool.py
```diff
@@ -346,6 +346,8 @@ def rm_pending_output_callback(output_callbacks, jobcachepending,
                     # https://github.com/python/mypy/issues/797
                     partial(rm_pending_output_callback, output_callbacks,
                             jobcachepending))
+
+        builder = self._init_job(job_order, provObj, **kwargs)
```
mr-c: Twice? See below

FarahZKhan: fixed

cwltool/executors.py
```diff
@@ -46,6 +52,7 @@ def run_jobs(self,
     def execute(self,
                 t,                 # type: Process
                 job_order_object,  # type: Dict[Text, Any]
+                make_fs_access=None,
```
mr-c: SyntaxError: duplicate argument 'make_fs_access' in function definition

FarahZKhan: fixed in 2022b3d

Issue #764, opened 2018-05-21
cwltool --singularity failure with `ERROR  : Could not create /dev/loopX: Permission denied`


pvanheus:
## Expected Behavior
`cwltool --singularity` should work the same as without `--singularity`

(Surprisingly (to me) `cwltoil` version 3.15.0 works correctly despite seeming to use the same version of `cwltool`)

## Actual Behavior
`cwltool --singularity reverse/reverse.cwl input.json` (CWL and  input JSON below) fails with an error:
`ERROR  : Could not create /dev/loop10: Permission denied`

## Workflow Code

```
cwlVersion: v1.0
class: CommandLineTool

hints:
  DockerRequirement:
    dockerPull: pvanheus/reverse:latest

baseCommand: reverse.py

inputs:
  dnafile:
    type: File
    inputBinding:
      position: 1

stdout: $(inputs.dnafile.nameroot)_reversed$(inputs.dnafile.nameext)

outputs:
  rev_dnafile:
    type: stdout
```
with input.json:
```json
dnafile:
  class: File
  path: data/dna.txt
```

where `dna/dna.txt` is simply a text file with the string `GATACCA`.

## Full Traceback
```pytb
$ cwltool --singularity --debug reverse/reverse.cwl input.json
/home/pvh/anaconda3/bin/cwltool 1.0.20180518123035
Resolved 'reverse/reverse.cwl' to 'file:///home/pvh/Documents/SANBI/biocourse2017/cwltutorial/reverse/reverse.cwl'
[job reverse.cwl] initializing from file:///home/pvh/Documents/SANBI/biocourse2017/cwltutorial/reverse/reverse.cwl
[job reverse.cwl] {
    "dnafile": {
        "class": "File",
        "location": "file:///home/pvh/Documents/SANBI/biocourse2017/cwltutorial/data/dna.txt",
        "size": 8,
        "basename": "dna.txt",
        "nameroot": "dna",
        "nameext": ".txt"
    }
}
[job reverse.cwl] path mappings is {
    "file:///home/pvh/Documents/SANBI/biocourse2017/cwltutorial/data/dna.txt": [
        "/home/pvh/Documents/SANBI/biocourse2017/cwltutorial/data/dna.txt",
        "/var/lib/cwl/stgf6ec8e0d-317b-4ea4-947f-39cb93d45be7/dna.txt",
        "File",
        true
    ]
}
[job reverse.cwl] command line bindings is [
    {
        "datum": "reverse.py",
[...]

PR #771, opened 2018-05-25
Added fix for detecting with singularity --userns is broken

pvanheus:
Uses a real (4127 byte) singularity container to test if `--userns` works. This is a fix for #764.


pvanheus:
Does that codecov patch imply that the `singularity.py` code is not tested enough?

mr-c:
Jenkins, add to test list

mr-c:
I think lines 32-38 from https://github.com/boutiques/boutiques/pull/251/files#diff-354f30a63fb0907d4ad57269548329e3R32 should be sufficient to install Singularity in TravisCI

Do you want to add it, or shall I?

mr-c:
Jenkins, add to testlist

PR #777, opened 2018-05-29
fix #753

### LINKED ISSUE #753: singularity --contain incompatibility with beegfs
It looks like the `--contain`, `--pid` and `--ipc` flags cause more troubles and helping:
https://github.com/common-workflow-language/cwltool/blob/master/cwltool/singularity.py#L216

if I remove these it works. not sure if beegfs is the problem.

gijzelaerr:


mr-c:
Closing for now. Please re-open if this is still a problem

PR #794, opened 2018-06-14
Simplify interface of main()

tetron:
Can provide custom loadingContext/runtimeContext.


tetron:
So this turned into an additional update of internal interfaces as a follow on the previous major update.  On the plus side, mypy is able to check for a lot more things.

PR #798, opened 2018-06-18
Add poll interval

kghosesbg:
This saves CWLtool from hammering the CPU when running
a workflow with parallel components

### DIFF
--- a/cwltool/argparser.py
+++ b/cwltool/argparser.py
@@ -25,6 +25,10 @@ def arg_parser():  # type: () -> argparse.ArgumentParser
                         help="[experimental] Run jobs in parallel. "
                              "Does not currently keep track of ResourceRequirements like the number of cores"
                              "or memory and can overload this system")
+
+    parser.add_argument("--poll-interval", type=float, default=0,
+                        help="Delay polling of nodes to avoid hammering CPU in parallel workflows")
+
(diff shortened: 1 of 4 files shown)


tetron:
That's not the right place for it.  The wait periods should be in the toplevel parallel executor when the workflow iterator returns None (indicating no new jobs are ready).

kghosesbg:
@tetron , will this do it? Thanks!

tetron:
The real problem here seems to be that wait_for_next_completion() doesn't actually wait for anything.

kghosesbg:
@tetron are you thinking that we should use a Python3 feature like promises or something? Or are you ok with a explicit time delay, just thinking about where it should go? Thanks!

mr-c:
Jenkins, okay to test

tetron:
I would use a condition variable (eg threading.Event).

But there's actually some other (preexisting) problems with the current parallel executor.  It doesn't have a mutex protecting callbacks, so it is possible to be evaluating the workflow and processing callbacks at the same time, which is likely to cause race conditions and incorrect behavior.  The busywait loop is just a symptom of not managing the signaling between the threads correctly.  It's been on my todo list to fix one of these days.

tetron:
Superceded by #800

PR #800, opened 2018-06-22
Multiprocess executor with resource accounting

tetron:


kghosesbg:
Works on my mac with my simple test case that involves scatter. I can verify that the `--parallel` option no longer hammers the CPU. I see the `--memory=XX` option go out but see no `cpu` option. Is this expected?

tetron:
Yes, the --memory option limits the memory that can be used by a container, but limiting the numebr of CPUs that can be used is a little more complicated (I think you have to actually bind the container to specific CPUs).

With this branch cwltool does keep track of how much ram and cores are claimed and throttles the number of jobs to avoid overloading the system.

ThomasHickman:
https://github.com/common-workflow-language/cwltool/blob/2ba677ac25c90e9400c6b0d0d324fe7e11cb5fed/cwltool/argparser.py#L24 probably needs to be changed in the pull request as well

cwltool/executors.py
```diff
@@ -128,35 +131,86 @@ class MultithreadedJobExecutor(JobExecutor):
...
+    optimize usage.
     """
+
     def __init__(self):  # type: () -> None
         super(MultithreadedJobExecutor, self).__init__()
         self.threads = set()  # type: Set[threading.Thread]
         self.exceptions = []  # type: List[WorkflowException]
+        self.pending_jobs = []  # type: List[JobBase]
+
+        self.max_ram = psutil.virtual_memory().total / 2**20
```
ThomasHickman: This probably needs to be something `psutil.virtual_memory().available - PADDING` - other processes will be normally be run on a machine (e.g. OS processes) which may fluctuate their memory usage while cwltool is running.

mr-c:
Testing against the conformance tests with `EXTRA=--parallel` 

```
Test timed out: /home/michael/cwltool/env3/bin/cwltool --parallel --outdir=/tmp/tmpzsqzuayz --quiet v1.0/scatter-wf1.cwl v1.0/scatter-job1.json
Test workflow scatter with single scatter parameter
```
I'm investigating this now

mr-c:
Thank you @tetron, @kghosesbg  and @ThomasHickman !

PR #804, opened 2018-06-28
WIP Add support for optional secondary files in v1.1.0

bogdang989:
Pull request for the issue https://github.com/common-workflow-language/common-workflow-language/issues/717

- [x] add test that CWL 1.0 document can't access this feature


mr-c:
Jenkins, add to testlist

tetron:
@bogdang989 the edits to Process.yml needs to made in a PR to https://github.com/common-workflow-language/cwl-v1.1

bogdang989:
Hi @tetron 
This might be it - https://github.com/common-workflow-language/cwl-v1.1/pull/4/commits/578b29792bcb53bff8d1374d912032df5b49cd2a

Marked this as WIP, since I additionally added the alternative '?' syntax, so I should probably add that to the spec description as well.

tetron:
Thanks for linking the other ticket.

cwltool/builder.py
```diff
@@ -301,13 +303,15 @@ def bind_input(self,
                         if isinstance(sf, dict) or "$(" in sf or "${" in sf:
                             sfpath = self.do_eval(sf, context=datum)
                         else:
-                            if sf.endswith('?'):
+                            if sf.endswith('?') and \
+                                            self.cwl_version == 'v1.1.0-dev1':
```
bogdang989: @tetron @mr-c  I added this naive check for the CWL version for now. I guess this should be done in a nicer way, to cover all the versions of the future and the past?

mr-c:
https://github.com/common-workflow-language/cwltool/blob/master/cwltool/command_line_tool.py#L282

cwltool/builder.py
```diff
@@ -296,18 +301,19 @@ def bind_input(self,
...
                             sf = sf['pattern']
                         else:
                             sf_required = True
-                        if isinstance(sf, dict) or "$(" in sf or "${" in sf:
+                        if isinstance(sf, MutableMapping) or "$(" in sf or "${" in sf:
                             sfpath = self.do_eval(sf, context=datum)
                         else:
-                            if sf.endswith('?'):
+                            if sf.endswith('?') and \
+                               self.cwl_version != "v1.0":
```
mr-c: how about `self.cwl_version is in ['v1.1.0-dev1']`? Seems safer

bogdang989: Yup, I was considering the same thing, I was just reluctant because with each new version "v1.1", "v1.2", "v2.0" the list will need to be updated.  Now that I think about it, that's why we have conformance tests, so it will probably be covered with tests, so I'll change this :)

tetron:
Schema PR is https://github.com/common-workflow-language/cwl-v1.1/pull/4

PR #810, opened 2018-07-04
Improve Research Object manifest

stain:
This PR ensure that `metadata/` **is** included in the RO `metadata/manifest.json`, including then `metadata/provenance/primary.cwlprov.provn` and friends.  Note that  `metadata/manifest.json` itself is **not** included in its own aggregation, but should be (and is) included in the `tagmanifest*` files.

As well this PR fixes some relative paths in the generated manifest, they need `../` to access `workflow/` etc as the manifest file resides inside `metadata/`.

The `conformsTo` for cwlprov files is also set to CWLPROV_VERSION (aka https://w3id.org/cwl/prov/0.3.0) rather than the `__citation__` DOI.  

This may however add confusion as to if this profile specifies the RO structure or the structure of the PROV files, but the PROV files also say they conformTo their corresponding PROV spec, e.g. (shortened):

```json
        {
            "uri": "provenance/primary.cwlprov.nt",
            "mediatype": "application/n-triples",
            "conformsTo": [
                "http://www.w3.org/TR/2013/REC-prov-o-20130430/",
                "https://w3id.org/cwl/prov/0.3.0"
            ]
        },
```

### DIFF
--- a/cwltool/provenance.py
+++ b/cwltool/provenance.py
@@ -1020,8 +1020,10 @@ class ResearchObject():
         })
 
         # How was it run?
+        # FIXME: Only primary*
         prov_files = [posixpath.relpath(p, METADATA) for p in self.tagfiles
-                      if p.startswith(_posix_path(PROVENANCE))]
+                      if p.startswith(_posix_path(PROVENANCE))
+                         and "/primary." in p]
         annotations.append({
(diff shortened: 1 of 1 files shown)


mr-c:
Thank you @stain and @FarahZKhan !

PR #820, opened 2018-07-12
Test Research Object manifest

stain:
Adds a test for --provenance to ensure the manifest includes the expected files, and that the paths match what exists on file system.


mr-c:
This will need to wait for a new schema_salad release with https://github.com/common-workflow-language/schema_salad/pull/191

Issue #822, opened 2018-07-13
cwltool fails to install in python3.7


ThomasHickman:
```$ docker run -it python:3.7 /bin/bash
root@0b95bcccdca4:/# pip install cwltool
Collecting cwltool
  Downloading https://files.pythonhosted.org/packages/a4/4c/2e4034c9dc7f32bd0a4bf86cbf0dd9eeff4e10c1611e4d92aad73b28d5c5/cwltool-1.0.20180711112827-py2.py3-none-any.whl (666kB)
    100% |████████████████████████████████| 675kB 11.3MB/s
Collecting bagit>=1.6.4 (from cwltool)
  Downloading https://files.pythonhosted.org/packages/bd/d6/7f505929284541705c6acf83726eb5c74b092e7eaed3a166bc428215bb89/bagit-1.7.0-py2.py3-none-any.whl
Collecting six>=1.8.0 (from cwltool)
  Downloading https://files.pythonhosted.org/packages/67/4b/141a581104b1f6397bfa78ac9d43d8ad29a7ca43ea90a2d863fe3056e86a/six-1.11.0-py2.py3-none-any.whl
Collecting typing>=3.5.3 (from cwltool)
  Downloading https://files.pythonhosted.org/packages/05/2b/2b05bf1d5a9dd450447c9a5df3e118a465e5d3cb12b73b7220a5064a403f/typing-3.6.4-py3-none-any.whl
Collecting schema-salad<3,>=2.6.20170927145003 (from cwltool)
  Downloading https://files.pythonhosted.org/packages/37/eb/9ed3748d0cc19acb4ac4d17893a647d70103cb37668aa5182fe2ee3f54ba/schema_salad-2.7.20180611133406-py2.py3-none-any.whl (380kB)
    100% |████████████████████████████████| 389kB 13.4MB/s
Collecting subprocess32>=3.5.0; os_name == "posix" (from cwltool)
  Downloading https://files.pythonhosted.org/packages/c3/5f/7117737fc7114061837a4f51670d863dd7f7f9c762a6546fa8a0dcfe61c8/subprocess32-3.5.2.tar.gz (96kB)
    100% |████████████████████████████████| 102kB 6.0MB/s
Collecting shellescape<3.5,>=3.4.1 (from cwltool)
[...]

PR #831, opened 2018-07-17
Relax ruamel.yaml dependancy cap and support python 3.7

ThomasHickman:
Following on from https://github.com/common-workflow-language/schema_salad/pull/189 and fixes https://github.com/common-workflow-language/cwltool/issues/822


mr-c:
`lxml` has wheels for `manylinux1` & py37 but not yet windows: https://pypi.org/project/lxml/#files

ThomasHickman:
Note: I added an issue to typings for the python 3.7 typing bug we are finding: https://github.com/python/typing/issues/573. Their solution is to not require typing for python 3.5+, as it is built in (which is fair enough). I think we should implement this

ThomasHickman:
^ changed that to basically `typing>=3.5.3; python_version < "3.7"` as python 3.7 is the first release where typing is not released as "provisional"

Issue #838, opened 2018-07-23
[Regression] --record-container-id causes internal exception


tom-tan:
When we run `cwltool` with `--record-container-id` to execute a workflows that use docker, the execution fails due to the internal exception.
This issue was introduced by #835 (commit: 8ddb82d81daf165442d90f73926838972612739d).

## Expected Behavior
```console
$ mkdir -p tmp
$ python -mcwltool.main --record-container-id --outdir $PWD/tmp cwltool/schemas/v1.0/v1.0/search.cwl cwltool/schemas/v1.0/v1.0/search-job.json
...
Final process status is success
```

## Actual Behavior
```console
$ mkdir -p tmp
$ python -mcwltool.main --record-container-id --outdir $PWD/tmp cwltool/schemas/v1.0/v1.0/search.cwl cwltool/schemas/v1.0/v1.0/search-job.json
...
[workflow ] start
[workflow ] starting step index
[step index] start
Got workflow error
Traceback (most recent call last):
  File "/Users/tom-tan/cwltool/cwltool/executors.py", line 162, in run_jobs
    job.run(runtime_context)
  File "/Users/tom-tan/cwltool/cwltool/job.py", line 484, in run
    runtime = self.create_runtime(env, runtimeContext)
  File "/Users/tom-tan/cwltool/cwltool/docker.py", line 337, in create_runtime
    cidfile_dir = runtimeContext.cidfile_
AttributeError: 'RuntimeContext' object has no attribute 'cidfile_'
Workflow error, try again with --debug for more information:
'RuntimeContext' object has no attribute 'cidfile_'
```

## Workflow Code
[search.cwl in this repository](https://github.com/common-workflow-language/cwltool/blob/master/cwltool/schemas/v1.0/v1.0/search.cwl) is an example that can reproduce this issue.
I also confirm that it can be reproduced with another workflow that uses `DockerRequirement`.

## Full Traceback
```pytb
/Users/tom-tan/cwltool/cwltool/main.py 1.0.20180720074456
Resolved 'cwltool/schemas/v1.0/v1.0/search.cwl' to 'file:///Users/tom-tan/cwltool/cwltool/schemas/v1.0/v1.0/search.cwl'
Workflow checker warning:
cwltool/schemas/v1.0/v1.0/search.cwl:108:9:  Sink 'file'
cwltool/schemas/v1.0/v1.0/search.cwl:57:7:     expects secondaryFiles: ['.idx1', '^.idx2',
[...]

PR #840, opened 2018-07-24
Replace `cidfile_` with `cidfile_dir`

tom-tan:
This request is for #838.
It just replace `cidfile_` with `cidfile_dir` to fix a typo.


tom-tan:
No problem. Thank you for fixing the issue!

PR #842, opened 2018-07-24
Kill child processes when cwltool exits

### LINKED ISSUE #832: Feature: kill child tasks when CWL exits
If cwltool exists, because of a user interrupting cwltool, a bug in cwltool or an error in a workflow, child tasks which cwltool has run are not killed. It would be nice if cwltool killed these tasks.

### LINKED ISSUE #833: Kill node.js process when cwltool exits
At the moment the node.js process spawned by cwltool for executing JavaScript expressions and validation of JavaScript doesn't get killed when cwltool exits. It would be nice if this happened.

anowlcalledjosh:
(This is a bit of an ugly solution, but I can't see a better way to do it.)

As processes are spawned, they are added to a global list of processes (`utils.processes_to_kill`); when the program exits (either due to execution finishing, an exception being raised, or SIGTERM), `_terminate_processes` is called, and kills all the processes in that list. A new entrypoint `run` is added to do the necessary try/finally to make this happen.

Caveats:

- since Python doesn't allow to kill individual threads, those threads can continue to execute while we're pulling the rug from under their feet, potentially leading to unexpected behaviour
- there's always the chance that a thread could spawn a process after we've killed all the processes, but before Python exits (though this should be very rare)

Alternatives:

- we could [use process groups](https://stackoverflow.com/a/322317/5951320), but that doesn't work on Windows, and also means that cwltool has to SIGKILL itself to kill its children
- there's a way to [use prctl](https://stackoverflow.com/a/27295927/5951320) to get the kernel to send a child process an arbitrary signal when its parent dies, but it's Linux-only and doesn't necessarily work as desired with multiple threads (and also requires using `preexec_fn`, which can lead to deadlocks when there are multiple threads around)

Questions:

- Do we need to handle any signals apart from SIGTERM? SIGINT should already be handled by producing a KeyboardInterrupt.
- Should we send SIGTERM to child processes before SIGKILL, or is the extra complexity for that not worth it?

Fixes #832, fixes #833.

### DIFF
--- a/cwltool/utils.py
+++ b/cwltool/utils.py
@@ -10,11 +10,8 @@ import stat
 import pkg_resources
 from functools import partial  # pylint: disable=unused-import
 from typing import (IO, Any, AnyStr, Callable,  # pylint: disable=unused-import
-                    Dict, Iterable, List, Optional, Text, Tuple, TypeVar, Union,
-                    TYPE_CHECKING)
-
-if TYPE_CHECKING:
-    from typing import Deque
+                    Dict, Iterable, List, Optional, Text, Tuple, TypeVar, Union)
+from typing_extensions import Deque
 
 import six
 from six.moves import urllib, zip_longest
--- a/cwltool/utils.py
+++ b/cwltool/utils.py
@@ -10,8 +10,7 @@ import stat
 import pkg_resources
 from functools import partial  # pylint: disable=unused-import
 from typing import (IO, Any, AnyStr, Callable,  # pylint: disable=unused-import
-                    Deque, Dict, Iterable, List, Optional, Text, Tuple, TypeVar,
-                    Union)
+                    Dict, Iterable, List, Optional, Text, Tuple, TypeVar, Union)
 
 import six
 from six.moves import urllib, zip_longest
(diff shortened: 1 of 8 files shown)


mr-c:
Jenkins, add to testlist

cwltool/utils.py
```diff
@@ -9,8 +10,11 @@
 import pkg_resources
 from functools import partial  # pylint: disable=unused-import
 from typing import (IO, Any, AnyStr, Callable,  # pylint: disable=unused-import
-                    Dict, Iterable, List, Optional, Text, Tuple, TypeVar,
-                    Union)
+                    Dict, Iterable, List, Optional, Text, Tuple, TypeVar, Union,
+                    TYPE_CHECKING)
+
+if TYPE_CHECKING:
+    from typing import Deque
```
mr-c: Ugh, so annoying that appveyor's Pyhon 3.5 is 3.5.3 and not 3.5.4!

Looks like the canonical way to fix this is to use https://pypi.org/project/typing-extensions/ and
`from typing_extensions import Deque`

anowlcalledjosh: Done, hopefully I added the requirement everywhere I needed to.

mr-c:
Jenkins, test this please

mr-c:
Thank you @anowlcalledjosh !

PR #850, opened 2018-08-01
Make data relocation at the end of workflow  runs faster

psafont:
- [x] When moving files, only calculate realpaths of links instead of doing it
for all the files in the destination directory
- [x] Avoid stating through all the files in the destination folder
- [ ] Avoid stating through all the folders in the destination folder

### DIFF
--- a/cwltool/process.py
+++ b/cwltool/process.py
@@ -277,7 +277,7 @@ def collectFilesAndDirs(obj, out):
 
 
 def relocateOutputs(outputObj,             # type: Union[Dict[Text, Any],List[Dict[Text, Any]]]
-                    outdir,                # type: Text
+                    destination_path,      # type: Text
                     output_dirs,           # type: Set[Text]
                     action,                # type: Text
                     fs_access,             # type: StdFsAccess
@@ -315,14 +315,14 @@ def relocateOutputs(outputObj,             # type: Union[Dict[Text, Any],List[Di
 
     outfiles = []  # type: List[Dict[Text, Any]]
     collectFilesAndDirs(outputObj, outfiles)
-    pm = PathMapper(outfiles, "", outdir, separateDirs=False)
+    pm = PathMapper(outfiles, "", destination_path, separateDirs=False)
     stageFiles(pm, stageFunc=moveIt, symLink=False)
 
-    def _check_adjust(f):
-        f["location"] = file_uri(pm.mapper(f["location"])[1])
-        if "contents" in f:
-            del f["contents"]
-        return f
+    def _check_adjust(file):
+        file["location"] = file_uri(pm.mapper(file["location"])[1])
+        if "contents" in file:
+            del file["contents"]
+        return file
 
     visit_class(outputObj, ("File", "Directory"), _check_adjust)
     if compute_checksum:
@@ -333,11 +333,11 @@ def relocateOutputs(outputObj,             # type: Union[Dict[Text, Any],List[Di
     # make an internal relative symlink.
     if action == "move":
         relinked = {}  # type: Dict[Text, Text]
-        for root, dirs, files in os.walk(outdir):
+        for root, dirs, files in os.walk(destination_path):
             for f in dirs+files:
                 path = os.path.join(root, f)
-                rp = os.path.realpath(path)
-                if path != rp:
+                if os.path.islink(path):
+                    rp = os.path.realpath(path)
                     if rp in relinked:
                         if onWindows():
                             if os.path.isfile(path):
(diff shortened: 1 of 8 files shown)


cwltool/process.py
```diff
@@ -616,9 +617,10 @@ def _init_job(self, joborder, runtimeContext):
             tmpdir = runtimeContext.docker_tmpdir or "/tmp"
             stagedir = runtimeContext.docker_stagedir or "/var/lib/cwl"
         else:
-            outdir = fs_access.realpath(runtimeContext.outdir or
-                tempfile.mkdtemp(prefix=getdefault(runtimeContext.tmp_outdir_prefix,
-                                                   DEFAULT_TMP_PREFIX)))
+            outdir = (runtimeContext.outdir or
+                      tempfile.mkdtemp(prefix=getdefault(runtimeContext.tmp_outdir_prefix,
+                                                         DEFAULT_TMP_PREFIX)))
+            outdir = fs_access.realpath(outdir)
```
mr-c: Why? I wouldn't call `fs_access.realpath` on the output of `tempfile.mkdtemp`..

psafont: I did not mean to change the behaviour here, just the formatting. having said that, I understood that `fs_access.realpath` was already being called on the output of `tempfile.mkdtemp` Did I misunderstand the statement?

mr-c: Sorry, missed that this was a pure reformat. I'll investigate further, elsewhere.

psafont:
I'm happy with the performance of the code right now, although it is still scanning the destination folder. I've tried to simplify the code so further changes are easier to do, if they are needed. (like saving all symlinks in the outputs and then only changing those, without going through all the folder in the output folder)

The only thing I'm not sure about is if I included the `scandir` in the dependencies correctly since it's not needed for python 3.5+

I've not replaced the `os.walk` in pathmapper since I think that adding code for a library to make it a bit faster does not seem like it's going to give a tangible performance improvement.

mr-c:
Thank you @psafont for your work on this.

Did you see the conformance test failures?

https://ci.commonwl.org/job/cwltool-pr-conformance-multiver/1718/version=v1.0/testReport/junit/conformance_test_v1/0/Test_directory_output/
https://ci.commonwl.org/job/cwltool-pr-conformance-multiver/1718/version=v1.0/testReport/junit/conformance_test_v1/0/Test_directory_literal_output_created_by_ExpressionTool/

cwltool/process.py
```diff
@@ -287,20 +279,38 @@ def relocateOutputs(outputObj,             # type: Union[Dict[Text, Any],List[Di
     if action not in ("move", "copy"):
         return outputObj
 
-    def moveIt(src, dst):
+    def _collectDirEntries(obj):
+        # type: (Union[Dict[Text, Any], List[Dict[Text, Any]]], List[Dict[Text, Any]]) -> Iterator[Dict[Text, Any]]
```
mr-c: `error: Type signature has too many arguments`

psafont: Fixed! Tried to also fix the type problems from the consolidation into a single temp variable, then processing it contents that, but couldn't make it work.

psafont:
The two failures might have been because I used the wrong variable in a specific location. (strange the normal tests do not complain)

mr-c:
> Not sure why this happens in Travis' python 3.5 when it fine on python 3.6+
We run mypy to do the type checking using Python 3.5 (in both py3 and py2 mode) as the results are presumed to be the same from any version. Perhaps we should check the types in each of py{3.5,6,7}?

The error
```
cwltool/process.py:13: error: Module 'os' has no attribute 'scandir'
cwltool/process.py:15: error: Cannot find module named 'scandir'
```
Is due to lack of type annotations for scandir.

psafont:
Getting back to the complex flow of an inlined function makes it work. I tried to make the behaviour clearer by removing an indent level.

It's fishy, though, if the order is to 'move', why doesn't move actually move the files and they are copied instead?

psafont:
Seeing the coverage report, the most complex parts of the function are not being used: Mainly the folder merging on the relocate subfunction and relinking of symlinks at the end of the process.

Is there any case where this functionality may be useful?
If it is, it may be worth adding it to conformance tests since it seems to be a rare corner case.
If it isn't it may be worth rescoping the purpose of the function and surrounding code to simplify it.

tetron:
@psafont could you add some unit tests?  These functions are pretty self contained so it should be straightforward to write unit tests for them.  The solution to not enough coverage for the tricky bits is to improve the coverage :-)

psafont:
Since this seems to be part of an extension I'll add tests to `tests/test_ext.py`

PR #862, opened 2018-08-09
Provenance: PROV links for nested workflows

stain:
Tests and fixes for provenance for nested workflows and directories.

# Provenance file names for nested workflows

Renamed nested workflow provenance files to include run UUID:

```
stain@biggie:/tmp/tmp_y11y58iro/metadata/provenance$ ls
primary.cwlprov.json
primary.cwlprov.jsonld
primary.cwlprov.nt
primary.cwlprov.provn
primary.cwlprov.ttl
primary.cwlprov.xml
workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.json
workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.jsonld
workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.nt
workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.provn
workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.ttl
workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.xml
```

This avoids using `%` in escaping as such filenames would need to be escaped as `%25` in `manifest.json` and other PROV files.  We can argue if a human-readable prefix like `workflow_20compile` is actually needed(here derived from the step name `compile` I think?), alternateively  or we just always call it `workflow.35c7bff2-ed72-464b-8aac-4722a57ed9e5` or `nested.35c7bff2-ed72-464b-8aac-4722a57ed9e5` or just `35c7bff2-ed72-464b-8aac-4722a57ed9e5` alone?

Not yet tested, but by including the UUID the filenames should be unique per run rather than by per step name, thus should work better if there are iterations over a step, or if sub-sub-workflow have steps with the same name.

# Links to nested provenance files

Before it was hard to find the corresponding provenance file for a nested workflow. This adds two new relations, using [prov:has_provenance](https://www.w3.org/TR/prov-aq/) relation to say there is more provenance info in another file.

In `metadata/manifest.json` based directly from the wf run UUID:

```
        {
            "about": "urn:uuid:35c7bff2-ed72-464b-8aac-4722a57ed9e5",
            "content": [
                "provenance/workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.xml",
                "provenance/workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.provn",
                "provenance/workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.json",
                "provenance/workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.ttl",
                "provenance/workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.nt",
                "provenance/workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.jsonld"
            ],
            "oa:motivatedBy": {
                "@id": "http://www.w3.org/ns/prov#has_provenance"
            }
        }
```

and in PROV from `primary.cwlprov.*` to the `workflow-*.cwlprov.*` files:

```provn
  activity(id:35c7bff2-ed72-464b-8aac-4722a57ed9e5, -, -, [prov:type='wfprov:ProcessRun', prov:label="Run of workflow/packed.cwl#main/compile"])
  activity(id:35c7bff2-ed72-464b-8aac-4722a57ed9e5, -, -, 
     [prov:has_provenance='provenance:workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.ttl', 
     prov:has_provenance='provenance:workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.nt', 
     prov:has_provenance='provenance:workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.xml', 
     prov:has_provenance='provenance:workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.jsonld', 
     prov:has_provenance='provenance:workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.json', 
     prov:has_provenance='provenance:workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.provn'])
```

In both cases we see that `35c7bff2-ed72-464b-8aac-4722a57ed9e5` has more provenance in the other files. This approach should also work with other eventual provenance files from command line tools (white box provenance), even if those files are not `cwlprov` or even not even PROV - but there would need to be a way to indicate to cwltool that such files are provenance.

Untested: If there's a nested nested workflow then there would be similar has_provenance links in the intermediate PROV file, but not in the master PROV file.

(Note: As these prov files are all semantically equivalent it should not really necessary to list all of them - either with content-negotiation or `Link rel=alternate` relations to find the others. Unfortunately in RO we don't have a clear way at the moment to relate alternates so for now they are all listed directly.)

# UUID consistency

After the refactoring to split out `CreateProvProfile` class there were some attributes duplicated on `workflow_run_uuid` and so on that caused some inconsistencies in identifiers. The reason was that the old code would derive the RO UUID from the first mention of the master workflow UUID, but now the wrong attributes would be updated.

I suggest now to rather do it instead more predictable, so that the RO UUID is fixed from start, and the UUID is instead passed into the `CreateProvProfile` as a optional parameter. This also makes it easier to use the RO arcp URI earlier, such as in the new `has_provenance` links.

I've fixed so that the UUID of the RO root base URI is again equivalent to the master workflow run:

```
stain@biggie:/tmp/tmp_y11y58iro$ grep External bag-info.txt 
External-Identifier: arcp://uuid,2f2348a1-1375-4c93-8ac7-c60deff5e790/

stain@biggie:/tmp/tmp_y11y58iro$ grep :2f2348a1-1375-4c93-8ac7-c60deff5e790 metadata/provenance/primary.cwlprov.provn 
  activity(id:2f2348a1-1375-4c93-8ac7-c60deff5e790, 2018-08-09T09:33:44.444167, -, [prov:type='wfprov:WorkflowRun', prov:label="Run of workflow/packed.cwl#main"])
```

and that now the nested workflow prov files use the same arcp base URI rather than a fresh one:

```
stain@biggie:/tmp/tmp_y11y58iro$ grep arcp metadata/provenance/workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.provn 
  prefix researchobject <arcp://uuid,2f2348a1-1375-4c93-8ac7-c60deff5e790/>
```

Similarly the UUID inside the nested PROV file of a nested workflow activity matches the UUID of the step in the upper PROV file (and correspond to the filename).

However the upper PROV do not currently mark the nested step as a `WorkflowRun` rather than `ProcessRun` (should it?) - to find out that you would currently have to dive into the nested PROV files or look at the plan.
 
# TODO: Wrong plan URIs

Note that the prospective/plan for nested workflows provenance still wrongly claim they are executing eleements of `packed.cwl#main`:

```
# from workflow_20compile.35c7bff2-ed72-464b-8aac-4722a57ed9e5.cwlprov.provn
  activity(id:47f527a0-0417-4e54-a0cc-5729f498747b, -, -, [prov:type='wfprov:ProcessRun', prov:label="Run of workflow/packed.cwl#main/untar"])
```

Here this should have been `packed.cwl#1st-workflow.cwl/untar` instead of using `#main`. I'll try to update this pull request to fix this.

# Directories and collections

Support for `class:Directory`, represented in PROV-N using [PROV Dictionary](https://www.w3.org/TR/prov-dictionary/):

```
  entity(id:d6ba95f7-bd67-4f2e-80c0-78840ef4b85b, [prov:type='prov:Dictionary',
           prov:type='cwlprov:Directory', prov:type='wfprov:Artifact', prov:type='prov:Collection', 
           prov:hadDictionaryMember='id:ba4ff3db-314a-4f19-b231-43d206f9fb5c', 
           prov:hadDictionaryMember='id:3a136630-30ca-4643-bec6-806d2e63b235', 
           prov:hadDictionaryMember='id:d6887434-dec2-4223-af70-6428f3af67f5'])
  hadMember(id:d6ba95f7-bd67-4f2e-80c0-78840ef4b85b, 
                     data:84a516841ba77a5b4648de2cd0dfcb30ea46dbb4)
  entity(id:d6887434-dec2-4223-af70-6428f3af67f5, 
            [prov:type='prov:KeyEntityPair', prov:pairKey="c", 
             prov:pairEntity='data:84a516841ba77a5b4648de2cd0dfcb30ea46dbb4'])
  entity(data:86f7e437faa5a7fce15d1ddcb9eaeaea377667b8, [prov:type='wfprov:Artifact'])
```

Note that currently directories themselves are identified by UUID, which could mean there could become multiple identifiers for what is effectively the same directory structure.

# Tests!

Last, but not least, I've added a bit of tests that try to verify the above, including loading up the PROV files as RDF and find the nested PROV files through the `has_provenance` link. 

I didn't use the PROV Python API for reading, as although it was able to load the PROV files it had little support for looking up. Thus some of the queries are a bit more complicated than they seem in PROVN as occassionally they have to follow n-ary relations like `qualifiedStart`.

With these identifiers sorted and tests working it should be easier to add/test more complicated PROV scenarios like nested nested workflows, filesets or scattering.  Some of the manifest/PROV testing could however with benefits be moved out to RDF Shapes to avoid deep graph traversal in Python.


mr-c:
```
cwltool/provenance.py:730: error: "List[Identifier]" has no attribute "uri"
cwltool/provenance.py:812: error: Incompatible types in assignment (expression has type "List[<nothing>]", variable has type "Dict[Any, Any]")
cwltool/provenance.py:1067: error: "Dict[Any, Any]" has no attribute "append"
cwltool/provenance.py:1068: error: Incompatible return value type (got "_Collection[str]", expected "str")
cwltool/workflow.py:260: error: Argument 2 to "activity_has_provenance" of "CreateProvProfile" has incompatible type "*List[Identifier]"; expected "List[Identifier]"
cwltool/workflow.py:260: note: "List" is invariant -- see http://mypy.readthedocs.io/en/latest/common_issues.html#variance
cwltool/workflow.py:260: note: Consider using "Sequence" instead, which is covariant
cwltool/workflow.py:505: error: Name 'uuid' is not defined
```

stain:
I'll try to fix these typing errors. Updated line numbers:

```
cwltool/provenance.py:552: error: Argument 1 to "BytesIO" has incompatible type "Union[str, bytes]"; expected "bytes"
cwltool/provenance.py:571: error: Call to untyped function "has_data_file" in typed context
cwltool/provenance.py:588: error: Name 'declare_artefact' is not defined
cwltool/provenance.py:613: error: Call to untyped function "membership" in typed context
cwltool/provenance.py:647: note: Redundant cast to "Dict[Any, Any]"
cwltool/provenance.py:649: error: Call to untyped function "membership" in typed context
cwltool/provenance.py:685: error: Call to untyped function "membership" in typed context
cwltool/provenance.py:827: error: "List[Identifier]" has no attribute "uri"
cwltool/provenance.py:1165: error: Incompatible return value type (got "_Collection[str]", expected "str")
cwltool/job.py:229: error: Argument 2 to "used_artefacts" of "CreateProvProfile" has incompatible type "Optional[str]"; expected "str"
cwltool/workflow.py:260: error: Argument 2 to "activity_has_provenance" of "CreateProvProfile" has incompatible type "*List[Identifier]"; expected "List[Identifier]"
cwltool/workflow.py:260: note: "List" is invariant -- see http://mypy.readthedocs.io/en/latest/common_issues.html#variance
cwltool/workflow.py:260: note: Consider using "Sequence" instead, which is covariant
cwltool/workflow.py:505: error: Name 'uuid' is not defined
```

FarahZKhan:
It has this one python2 vs. 3 issue I guess: 
```
____________________ TestProvenance.test_directory_workflow ____________________
self = <tests.test_provenance.TestProvenance testMethod=test_directory_workflow>
    def test_directory_workflow(self):
        dir2 = os.path.join(tempfile.mkdtemp("test_directory_workflow"),
            "dir2")
        os.makedirs(dir2)
        sha1 = {
            # Expected hashes of ASCII letters (no linefeed)
            # as returned from:
            ## for x in a b c ; do echo -n $x | sha1sum ; done
            "a": "86f7e437faa5a7fce15d1ddcb9eaeaea377667b8",
            "b": "e9d71f5ee7c92d6dc9e92ffdad17b8bd49418f98",
            "c": "84a516841ba77a5b4648de2cd0dfcb30ea46dbb4",
        }
        for x in ("a", "b", "c"):
            # Make test files with predictable hashes
            with open(os.path.join(dir2, x), "w", encoding="ascii") as f:
>               f.write(x)
E               TypeError: write() argument 1 must be unicode, not str
tests/test_provenance.py:103: TypeError
```

PR #864, opened 2018-08-09
Support ruamel 0.15.52+

tetron:
Companion schema salad update https://github.com/common-workflow-language/schema_salad/pull/195


cjllanwarne:
Given the statement on https://pypi.org/project/ruamel.yaml/ ...

> Starting with version 0.15.0 the way YAML files are loaded and dumped is changing. See the API doc for details. Currently existing functionality will throw a warning before being changed/removed. For production systems you should pin the version being used with ``ruamel.yaml<=0.15``. There might be bug fixes in the 0.14 series, but new functionality is likely only to be available via the new API.

... is it worth pinning the ruamel version to `<=15.52` in case another breaking change comes along in the near future?

tetron:
We used to have it pinned, but recently relaxed the pin because the pin was creating problems, and for some set of 0.15.x versions it works fine.  Ubuntu 18.04, for example, packages a 0.15.x version of ruamel.yaml.  But setting a version cap is probably a good idea.

tetron:
It is passing conformance tests, but there are still some lose ends.  Also, because CommentedMap is not a Dict, type annotations have to be updated to use MutableMapping as well.

tetron:
In the latest ruamel.yaml we need to use MutableSequence (instead of list) as well, which has broken more tests.  Work in progress.

Issue #867, opened 2018-08-11
cwltool use of singularity fails with bioconda containers with default output directory (/var/spool/cwl)


pvanheus:
## Expected Behavior
When cwltool uses a container, it maps a directory on the host filesystem to a working directory path within the container. By default this path is `/var/spool/cwl`. This is passed to the singularity command line where it is used in a bind-mount and for the `--pwd` parameter. For the CWL code below it yields:

```
singularity \
    --quiet \
    exec \
    --contain \
    --pid \
    --ipc \
    --bind \
    /tmp/tmpmMeF3o:/var/spool/cwl:rw \
    --bind \
    /tmp/tmp6fUHbo:/tmp:rw \
    --pwd \
    /var/spool/cwl \
    /some/path/quay.io-biocontainers-blast-2.7.1--boost1.64_1.img \
    blastp \
    . > /tmp/tmpmMeF3o/listing.txt
```

This is all correct from the `cwltool` side. It leads to a failure with bioconda containers, as described below.
## Actual Behavior
Bioconda containers are built on a `busybox` base container where `/var/spool` is a symbolic link to `/var/tmp` which itself is a symbolic link to `/tmp`.

Singularity does a bind-mount between the host system's `/tmp` and `/tmp` within the container. Because of the symbolic links described above, `/var/spool/cwl` is effectively `/tmp` so the second bind-mount fails (silently). The result is that inside the container `/var/spool/cwl` does not exist, resulting in the error shown in the traceback. I have reported the Singularity issue: https://github.com/singularityware/singularity/issues/1833

A workaround is to use a `dockerRequirement` under `requirements` and specify a `dockerOutputDirectory` that does not clash with an existing bind mount. E.g.

```
requirements:
   - class: DockerRequirement
     dockerPull: quay.io/biocontainers/blast:2.7.1--boost1.64_1
     dockerOutputDirectory: /var/cwl
```

## Workflow Code
```
cwlVersion: v1.0
class: CommandLineTool

hints:
  - class: DockerRequirement
    dockerPull: quay.io/biocontainers/blast:2.7.1--boost1.64_1

inputs:
  filename:
    type: string
    default: "."
    inputBinding:
      position: 1

outputs:
  output_file:
    type: File
    outputBinding:
      glob: listing.txt

stdout: listing.txt

baseCommand: ls
```
## Full Traceback
```
$ cwltool --singularity example_failure.cwl 
[...]

PR #868, opened 2018-08-11
Add random outdir support

pvanheus:
Currently, when `cwltool` uses a container runtime, it does a bind mount of a temporary directory on the host into `/var/spool/cwl` inside the container. This is related to the problem reported in #867.

A safer strategy (proposed by @mr-c and @tetron) that using `/var/spool/cwl` is choosing a randomly generated directory name off the filesystem root (e.g. `/OJzbZZ`). This PR implements this behaviour, using a new utility function, `random_outdir` that generates a suitable directory name and then stores it in a function attribute, allowing the same name to be returned on subsequent invocations.


mr-c:
Thank you @pvanheus !

Issue #872, opened 2018-08-17
pip install cwltool failed with mypy-extension==0.4.0


inutano:
## Expected Behavior
`pip install cwltool` will give you `cwltool`

## Actual Behavior
`pip install cwltool` failed with error of `Downloading/unpacking mypy-extensions`.

I guess the recent version of `mypy-extension==0.4.0` will cause the problem, since `pip install mypy-extension==0.3.0` followed by `pip install cwltool` goes fine.

## Full Traceback

This Dockerfile

```
FROM debian:jessie-slim
RUN apt-get update -qq
RUN apt-get install -y python-dev python-pip libxml2-dev libxslt1-dev zlib1g-dev
RUN pip install cwltool
```

Failed.

```
Step 4/4 : RUN pip install cwltool
 ---> Running in c7d559c034e2
Downloading/unpacking cwltool
Downloading/unpacking bagit>=1.6.4 (from cwltool)
  Downloading bagit-1.7.0-py2.py3-none-any.whl
Downloading/unpacking subprocess32>=3.5.0 (from cwltool)
  Running setup.py (path:/tmp/pip-build-67BGbi/subprocess32/setup.py) egg_info for package subprocess32
    /usr/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'python_requires'
      warnings.warn(msg)

    no previously-included directories found matching 'build'
    no previously-included directories found matching 'dist'
    no previously-included directories found matching 'config.*'
    no previously-included directories found matching '*.pyc'
Downloading/unpacking mypy-extensions (from cwltool)
  Downloading mypy_extensions-0.4.0.tar.gz
  Running setup.py (path:/tmp/pip-build-67BGbi/mypy-extensions/setup.py) egg_info for package mypy-extensions
    error in mypy_extensions setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers
    Complete output from command python setup.py egg_info:
    error in mypy_extensions setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers

----------------------------------------
Cleaning up...
Command python setup.py egg_info failed with error code 1 in /tmp/pip-build-67BGbi/mypy-extensions
Storing debug log for failure in /root/.pip/pip.log
The command '/bin/sh -c pip install cwltool' returned a non-zero code: 1
```

## Your Environment
* cwltool version: without version specification in `pip install`

manabuishii:
[...]

PR #877, opened 2018-08-20
Bump six and requests dependency versions

anowlcalledjosh:
cwltool itself doesn't require these versions, but its dependencies do.

Since Pip doesn't have a backtracking dependency resolver yet, it can fail to find a correct solution if the package being installed has low requirements that are already installed, but a dependency has a higher requirement. For example, cwltool requires six 1.8.0, but prov requires six 1.9.0; Debian Jessie packages six 1.8.0, so `pip install cwltool` sees that the installed six meets cwltool's requirements, but it isn't able to determine that it should install a newer version of six once it scans prov's dependencies. (More recent versions of Pip will at least print a warning in this case.)

The Requests version bump is to work around an issue with old Requests versions and CacheControl (https://github.com/ionrock/cachecontrol/issues/137).

See #872.


mr-c:
Thanks!

PR #879, opened 2018-08-21
WIP InitialWorkdirRequirement array and null

bogdang989:
This works on the premises that the expression in InitialWorkDirRequirement can be:
- File[]
- Directory[]
- null


tetron:
> Or we could amend the conformance tests to specify realistic minimums?

I suppose, but that involves changing 100 files instead of just 1 😕

This is one of those areas where it is under specified and we should probably have a documented default in the spec.

whlavina:
The commit on this issue caused issue #1076.

PR #887, opened 2018-08-29
Enable code coverage reporting in appveyor

psafont:
Force codecov to collect coverage.xml in appveyor


psafont:
Also it would be interesting to add the environment variable `$CODECOV_TOKEN` to the windows jenkins worker so those coverage reports are uploaded too.

Issue #882, opened 2018-08-27
--provenance doesn't capture the final output YAML/JSON object


mr-c:
It would be great if this had a predictable file name instead of a checksum based file name

I think the output object that is sent to stdout should also match (that is, the path/location should point within the generated RO folder)

PR #893, opened 2018-09-05
add output object to RO

FarahZKhan:
This pull request adds the CWL output JSON object to the Research Object with the updated locations referring to data artefacts in the RO.  The output object is also listed in the manifest file. A very basic test is also added to test_provenance. 

Fixes https://github.com/common-workflow-language/cwltool/issues/882


cwltool/main.py
```diff
@@ -713,6 +713,7 @@ def my_represent_none(self, data):  # pylint: disable=unused-argument
                                      logger=_logger)
 
             if out is not None:
+                runtimeContext.research_obj.create_job(None, out, True)
```
mr-c: `runtimeContext.research_obj` could be `None` here

FarahZKhan: Thanks :) fixed it in new commit

Issue #910, opened 2018-09-19
cwltool relocateOutputs 'move' action for type Directory output


carcason:
## Expected Behavior
 relocateFunction moves src directory into dst directoy

## Actual Behavior
 CRASH

## Workflow Code
```
cwlVersion: v1.0
class: ExpressionTool

requirements:
  - class: InlineJavascriptRequirement

id: "files_to_folder"
inputs:
  input_files:
    type:
    - File[]
    - File
outputs:
  folder: Directory
expression: |
  ${
    var folder = {
      "class": "Directory",
      "basename": "",
      "listing": []
    }
    var files = [];
    if (!Array.isArray(inputs.input_files)){
      files.push (inputs.input_files)
    } else {
      files = inputs.input_files
    }
    for (var i = 0; i < files.length; i++){
      if (files[i].secondaryFiles && files[i].secondaryFiles.length > 0){
        Array.prototype.push.apply(folder.listing, files[i].secondaryFiles);
        delete files[i].secondaryFiles;
      }
      folder.listing.push (files[i])
    }
    return { "folder": folder };
  }

doc: |
  Returns Directory object with a listing set from File or File[]. Only one nested level of secondaryFiles
  is supported
```

## Full Traceback
```pytb
[2018-09-19 09:43:43,806] {base_task_runner.py:107} INFO - Job 138: Subtask CWLJobGatherer [2018-09-19 09:43:43,806] {cli.py:492} INFO - Running <TaskInstance: bowtie-index.CWLJobGatherer 2018-09-19T0
[2018-09-19 09:43:43,826] {models.py:1736} ERROR - 'posix.DirEntry' object has no attribute 'startswith'                                                                                                
Traceback (most recent call last):                                                                                                                                                                      
  File "/usr/local/lib/python3.6/site-packages/airflow/models.py", line 1633, in _run_raw_task                                                                                                          
    result = task_copy.execute(context=context)                                                                                                                                                         
  File "/usr/local/lib/python3.6/site-packages/cwl_airflow_parser/operators/cwljobgatherer.py", line 109, in execute                                                                                    
    return self.cwl_gather(context)                                                                                                                                                                     
  File "/usr/local/lib/python3.6/site-packages/cwl_airflow_parser/operators/cwljobgatherer.py", line 92, in cwl_gather                                                                                  
    StdFsAccess(""))                                                                                                                                                                                    
[...]

PR #912, opened 2018-09-21
Fix folder relocations on existing folders

psafont:
As shown in #910 it fails on certain cases


psafont:
For some reason, in older versions of Python files are being created *before* their parent folders.

I created a workaround for it, but maybe we want to remove that and change the order which add_volumes is called before pulling the other changes

PR #926, opened 2018-10-01
Remove link handling in relocation

psafont:
There are no tests for this complicated piece of code.

If they are not added and it doesn't match any use-case we might as well get rid of it.


mr-c:
@psafont Thanks!

Issue #775, opened 2018-05-28
--singularity and InitialWorkDirRequirement are not functional


mr-c:
See https://github.com/singularityware/singularity/issues/1607
We should document this this outside this issue

KerstenBreuer:
Is there any solution on the horizon? We would really need this functionality.

mr-c:
Hello @KerstenBreuer there is now a workaround at #935

PR #935, opened 2018-10-04
workaround for Singularity + InitialWorkDirRequriement

mr-c:
Does cause increased IO for writable: True

Fixes https://github.com/common-workflow-language/cwltool/issues/775


mr-c:
@gijzelaerr Once you confirm this fix I'll merge, make a release, and patch `toil-cwl-runner` to use this

PR #939, opened 2018-10-07
cwlprov: stop copying non-CWL files into snapshot/

mr-c:
and clean up --print-deps and --print-inputs-deps codepaths
Normalize output of --print-input-deps with command line parameters to better match those provided via input object file


mr-c:
Jenkins, test this please

PR #942, opened 2018-10-12
Use explicits checks when possible

psafont:
This includes checking against None and checking if keys are in a dict.

Some type tweaks were added to help with this too.

In provenance there also was some low-hanging fruit to removed dead code.


cwltool/provenance.py
```diff
@@ -485,28 +483,16 @@ def start_process(self, process_name, process_run_id=None):
...
-                process_run_id, None, None,
-                {provM.PROV_TYPE: WFPROV["ProcessRun"],
-                 provM.PROV_LABEL: prov_label})
-            self.document.wasAssociatedWith(
-                process_run_id, self.engine_uuid, str("wf:main/" + process_name))
-            self.document.wasStartedBy(
-                process_run_id, None, self.workflow_run_uri,
-                datetime.datetime.now(), None, None)
-        else:
-            prov_label = "Run of CommandLineTool/packed.cwl#main/"
```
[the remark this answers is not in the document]
psafont: It's dead code, self.workflow_run_uri can't be an empty string nor None. (line 332 and 334)

cwltool/provenance.py
```diff
@@ -1095,19 +1082,20 @@ def add_tagfile(self, path, when=None):
             # are not too large..?
 
             checksums[SHA1] = checksum_copy(tag_file, hasher=hashlib.sha1)
-            tag_file.seek(0)
```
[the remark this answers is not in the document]
psafont: It's redundant: The file only needs to be reset it needs to be read again to calculate another checksum, but the rest of the code is already doing that.

cwltool/provenance.py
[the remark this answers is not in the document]
mr-c: Tagging in @stain and @FarahZKhan to think about this

psafont:
My guess is that the coverage is low because usually the objects are rarely None in real cases.

It might mean that they shouldn't be Optional. I've changed a couple of cases where that was true.

mr-c:
Jenkins, test this please

cwltool/builder.py
```diff
@@ -109,67 +109,43 @@ def get_requirement(self,
...
-                 stagedir=u"",              # type: Text
-                 job_script_provider=None   # type: Optional[Any]
+                 job,                         # type: Dict[Text, Union[Dict[Text, Any], List, Text, None]]
+                 timeout,                     # type: float
+                 files=[],                    # type: List[Dict[Text, Text]]
+                 bindings=[],                 # type: List[Dict[Text, Any]]
+                 schemaDefs={},               # type: Dict[Text, Dict[Text, Any]]
+                 names=Names(),               # type: Names
+                 requirements=[],             # type: List[Dict[Text, Any]]
+                 hints=[],                    # type: List[Dict[Text, Any]]
```
mr-c: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

psafont: Good catch, wasn't aware of the behaviour: I've removed all lists and dicts as default arguments

cwltool/argparser.py
```diff
@@ -418,7 +418,7 @@ def add_argument(toolparser, name, inptype, records, description="",
                 toolparser, fieldname, fieldtype, records,
                 fielddescription)
         return
-    if inptype == "string":
+    elif inptype == "string":
```
mr-c: no need for an `elif` after a `return`

cwltool/load_tool.py
```diff
@@ -195,21 +195,24 @@ def _add_blank_ids(workflowobj):
...
+                      strict=True,               # type: bool
+                      preprocess_only=False,     # type: bool
                       fetcher_constructor=None,  # type: FetcherConstructorType
-                      skip_schemas=None,  # type: bool
-                      overrides=None,  # type: List[Dict]
-                      metadata=None,  # type: Optional[Dict]
-                      do_validate=True
+                      skip_schemas=None,         # type: bool
+                      overrides=None,            # type: List[Dict]
+                      metadata=None,             # type: Dict[Text, Any]
```
mr-c: This type should still be marked `optional` if the default value is `None`. This is enforced by newer versions of `mypy`

cwltool/process.py
```diff
@@ -596,13 +596,15 @@ def __init__(self,
 
         dockerReq, is_req = self.get_requirement("DockerRequirement")
 
-        if dockerReq and dockerReq.get("dockerOutputDirectory") and not is_req:
+        if dockerReq is not None and dockerReq.get("dockerOutputDirectory")\
+                and is_req is not None and not is_req:
```
mr-c: while you're at it, might as well say `"dockerOutputDirectory" in dockerReq` instead of the `.get()`

mr-c:
Looking good! Now to address the test coverage ...

cwltool/argparser.py
psafont: This would be true if all the cases before this one would end the function.

If I don't put the `elif`, the last `else` returns `None` when `inptype` is `'File'`,  `'Directory'`, etc

cwltool/provenance.py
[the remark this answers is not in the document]
stain: This `else` block was meant to handle when executing a command line tool CWL directly, are you saying this code would wrongly always call the `workflow_run_uri` also if a command line tool is executed? See [test_hello_single_tool](https://github.com/common-workflow-language/cwltool/blob/master/tests/test_provenance.py#L62)

cwltool/provenance.py
```diff
@@ -324,13 +322,13 @@ def __init__(self,
         self.user_provenance = user_provenance
         self.engine_uuid = research_object.engine_uuid
         self.add_to_manifest = self.research_object.add_to_manifest
-        if self.orcid:
+        if self.orcid is not None:
```
stain: empty orcid string is not valid and this would cause exception later on.

cwltool/provenance.py
```diff
@@ -324,13 +322,13 @@ def __init__(self,
...
             _logger.debug(u"[provenance] Creator ORCID: %s", self.orcid)
         self.full_name = full_name or None
-        if self.full_name:
+        if self.full_name is not None:
```
stain: Empty "full name" is non-nonsensical and should not be logged in provenance as such. 

While perhaps undocumented, setting an empty name can in the old code be used on command line to override `CWL_FULL_NAME` environment variable, as it's not possible to say `--full-name=None` on the command line.

cwltool/provenance.py
```diff
@@ -324,13 +322,13 @@ def __init__(self,
...
             _logger.debug(u"[provenance] Creator Full name: %s",
                           self.full_name)
-        if not run_uuid:
+        if run_uuid is None:
```
stain: An empty string as `run_uuid` will (should!) cause failure as it is not a valid UUID.

stain:
While I would generally think it's easier code to read without the `is None` bits and so would be reluctant to this pull request; but I see here that making this explicit can open up test coverage that should be detected.

cwltool/process.py
psafont: Not sure why github doesn't mark this as outdated, it's fixed.

cwltool/provenance.py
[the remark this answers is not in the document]
psafont: The only expression I found in the code where `CreateProvProfile.workflow_run_uri` is changed is in line 334, which invariably sets it to `run_uuid.urn`.

I'll see what changes when only a tool is run.

mr-c:
https://ci.commonwl.org/view/All/job/cwltool-pr-conformance-multiver/2006/version=v1.0/testReport/junit/conformance_test_v1/0/Test_workflow_scatter_with_two_scatter_parapmeters_and_nested_crossproduct_join_method_and_valueFrom_on_step_input/

psafont:
I can't reproduce the error on my local machine, regardless of python version used.

Also the stderr is about resource usage:

```
Job "step1_3" requested more resources ({'cores': 1, 'ram': 1024}) than are available (max ram is 13343.000000, max cores is 2.000000)
Job "step1_3" requested more resources ({'cores': 1, 'ram': 1024}) than are available (max ram is 13343.000000, max cores is 2.000000)
Job "step1_3" requested more resources ({'cores': 1, 'ram': 1024}) than are available (max ram is 13343.000000, max cores is 2.000000)
Job "step1_4" requested more resources ({'cores': 1, 'ram': 1024}) than are available (max ram is 13343.000000, max cores is 2.000000)
```

Not too sure which of my changes may affect resource usage.

mr-c:
@psafont That test machine has only two cores, and the tests are being run with `cwltool --parallel`. I can't reproduce it on my developer machine either, but I can on that same test node when I run the tests in parallel but not when I run them one at a time.

Alas, the error is present in the main branch, without your changes, so you are off the hook

cwltool/provenance.py
psafont: Since it doesn't seem to be a difference between the usage if `None` and `''` as they should both be ignore, I'm leaning towards removing the value `None` from the code:
1. Set the default value to `''` instead of `None`
2. Don't allow the variable to be None

I'm not seeing any tests dealing with the name, could you give some hints on how to check normal behaviour? I might be able to create some corner cases from there.

mr-c: Look at the following options for `cwltool`:
```
Options for recording provenance information of the execution:
  --provenance PROVENANCE
                        Save provenance to specified folder as a Research
                        Object that captures and aggregates workflow execution
                        and data products.
  --enable-user-provenance
                        Record user account info as part of provenance.
  --disable-user-provenance
                        Do not record user account info in provenance.
  --enable-host-provenance
                        Record host info as part of provenance.
  --disable-host-provenance
                        Do not record host info in provenance.
  --orcid ORCID         Record user ORCID identifier as part of provenance,
                        e.g. https://orcid.org/0000-0002-1825-0097 or
                        0000-0002-1825-0097. Alternatively the environment
                        variable ORCID may be set.
  --full-name CWL_FULL_NAME
                        Record full name of user as part of provenance, e.g.
                        Josiah Carberry. You may need to use shell quotes to
                        preserve spaces. Alternatively the environment
                        variable CWL_FULL_NAME may be set.
```

psafont: I was more worried about where is the result stored:

I'm seeing 21 hits across 9 files when I set a full-name, I wanted to know what is the expected result when the name is blank, are these fields written? if not, how so?

cwltool/provenance.py
psafont: Now we're only checking for the empty string instead of None

cwltool/provenance.py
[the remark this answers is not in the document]
psafont: @stain I've checked again:
That function (`start_process`) never gets called when running a single tool (L456), it only gets called when running a tool inside a workflow (L475)

tests/test_singularity.py
```diff
@@ -33,9 +20,19 @@ def test_singularity_workflow():
...
-    result = main(
+    result_code = main(
         ['--singularity', '--default-container', 'debian',
          get_data("tests/wf/iwdr-entry.cwl"), "--message", "hello"])
-    assert result == 0
+    singularity_installed = bool(distutils.spawn.find_executable('singularity'))
+    if singularity_installed:
+        assert result_code == 0
+    else:
+        assert result_code != 0
```
mr-c: :+1:

cwltool/provenance.py
```diff
@@ -138,14 +134,23 @@ def _local_path(posix_path):
...
     """Return the current operating system account as (username, fullname)."""
     username = getuser()
-    fullname = username
-    if GET_PW_NAM:
-        pwnam = getpwnam(username)
-        if pwnam:
-            fullname = pwnam.pw_gecos.split(",", 1)[0]
+    try:
+        if onWindows():
+            get_user_name = ctypes.windll.secur32.GetUserNameExW  # type: ignore
```
mr-c: :heart_eyes:

psafont: Had to grab a Windows machine to make sure it works properly ;)

mr-c:
Jenkins, test this please.

mr-c:
Jenkins, test this please

mr-c:
Jenkins, test this please

mr-c:
Jenkins, test this please

mr-c:
Thank you @psafont !

psafont:
yay!

PR #947, opened 2018-10-13
Measure Docker container memory usage

mr-c:
This works, but fails on TravisCI due to the single core allocated.


mr-c:
See also, the Arvados code that does something similar: https://github.com/curoverse/arvados/blob/master/lib/crunchstat/crunchstat.go

PR #949, opened 2018-10-18
extract/execute workflow subgraph

tetron:
Adds --target, --print-targets, and --print-subgraph

--target takes the id of a parameter (input or output) or step of the toplevel workflow and trims the workflow to just the dependencies and/or downstream products of the parameter or step.

--print-targets prints the output parameters

--print-subgraph prints the toplevel workflow (can be combined with --target)

The purpose of this feature is to support Workflows that consist of several independent subgraphs, so it is possible to select a particular subgraph to execute, or execute all of them.  The use case is to be able to group and manage a set of related tasks into a single workflow.


tetron:
This is WIP, I haven't added tests or type annotations

Issue #954, opened 2018-10-24
SoftwareRequirement fails if in a Python3  virtualenv


mr-c:
From https://www.biostars.org/p/344955/

## Expected Behavior
Tell us what should happen

## Actual Behavior
```
openjdk-8.0.12 100% |##############################################################| Time: 0:00:01  59.85 MB/s
rtg-tools-3.8. 100% |##############################################################| Time: 0:00:00  15.25 MB/s
Extracting packages ...
[      COMPLETE      ]|#################################################################################| 100%
Linking packages ...
[      COMPLETE      ]|#################################################################################| 100%
#
# To activate this environment, use:
# > source activate __rtg-tools@3.8.2
#
# To deactivate this environment, use:
# > source deactivate __rtg-tools@3.8.2
#

Exception while running job
Traceback (most recent call last):
  File "/nethome/usevani/miniconda2/lib/python2.7/site-packages/cwltool/job.py", line 278, in _execute
    name=self.name
  File "/nethome/usevani/miniconda2/lib/python2.7/site-packages/cwltool/job.py", line 575, in _job_popen
    mode="wb") as job_file:
ValueError: binary mode doesn't take an encoding argument
[job rtg.cwl] completed permanentFail
{}
Final process status is permanentFail
```

## Workflow Code
```
cwlVersion: v1.0
class: CommandLineTool
baseCommand: rtg

requirements:
  EnvVarRequirement:
    envDef:
      TOIL_GRIDENGINE_PE: smp

hints:
  SoftwareRequirement:
    packages:
    - package: 'rtg'
      version:
      - '3.8.2'
      specs:
      - https://anaconda.org/bioconda/rtg-tools

arguments:
- valueFrom: vcfeval
  position: 2
- valueFrom: 20G
  position: 1
  prefix: RTG_MEM=
  separate: false
- valueFrom: "$(runtime.outdir)/out"
  prefix: -o
  position: 3
- valueFrom: $(runtime.cores)
  prefix: -T
  position: 4

inputs:
  rtg_reference:
    type: Directory
    inputBinding:
      position: 5
      prefix: -t
  gold_standard:
    type: File
    inputBinding:
      position: 6
      prefix: -b
    secondaryFiles:
    - .tbi
    doc: Genome-In-A-Bottle gold start VCF.
  input:
    type: File
    inputBinding:
      position: 7
      prefix: -c
    secondaryFiles:
    - .tbi
    doc: input compressed and index VCF
  all_records:
[...]

PR #961, opened 2018-11-01
Cleanup around testing and packaging lifecycle

psafont:
I've simplified managing with python packaging, removing test-requirements.txt and removing the direct usage of setup.py for installing the package and dependencies: that can be handled through pip and PEP518.

I also think there is a case to remove requirements.txt as it is, since it does not list all the dependencies needed in the environment, nor they are locked.

Let me know your thoughts, especially about requirements.txt. I haven't deleted the file but there is no other piece of code that actually uses it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,20 +26,12 @@ MODULE=cwltool
 # `SHELL=bash` doesn't work for some, so don't use BASH-isms like
 # `[[` conditional expressions.
 PYSOURCES=$(wildcard ${MODULE}/**.py tests/*.py) setup.py
-DEVPKGS=pycodestyle diff_cover autopep8 pylint coverage pydocstyle flake8 \
-	pytest-xdist isort
+DEVPKGS=pycodestyle diff_cover autopep8 pylint coverage pydocstyle flake8 isort
 DEBDEVPKGS=pep8 python-autopep8 pylint python-coverage pydocstyle sloccount \
 	   python-flake8 python-mock shellcheck
 VERSION=1.0.$(shell date +%Y%m%d%H%M%S --utc --date=`git log --first-parent \
 	--max-count=1 --format=format:%cI`)
 mkfile_dir := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
-UNAME_S=$(shell uname -s)
-ifeq ($(UNAME_S),Linux)
-	nproc=$(shell nproc)
-endif
-ifeq ($(UNAME_S),Darwin)
-	nproc=$(shell sysctl -n hw.physicalcpu)
-endif
 
 ## all         : default task
 all:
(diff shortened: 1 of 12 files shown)


psafont:
Python 3.4 is failing on appveyor because the wheel for 5.4.8 is not uploaded to PyPI: the compilation fails: https://github.com/giampaolo/psutil/issues/1356

psafont:
For Travis the failure appeared when enabling [isolated builds](https://tox.readthedocs.io/en/latest/config.html#conf-isolated_build) (PEP518)

I've reported the issue to tox-travis, to try to see where the issue is really: https://github.com/tox-dev/tox-travis/issues/119

tox.ini
```diff
@@ -7,39 +7,24 @@ envlist =
...
-  3.5: py35
-  3.6: py36
-  3.7: py37
-
 [testenv]
-passenv =
-  CI
-  TRAVIS
-  TRAVIS_*
+passenv = CI TRAVIS*
```
ryanhiebert: This shouldn't be necessary to make tox-travis work. I don't think it should cause issues either, though.

mr-c:
The travis failure is a manifestation of https://github.com/common-workflow-language/cwltool/issues/954

tox.ini
psafont: It's necessary to be able to upload coverage reports to codecov, though

psafont:
I need to make sure the coverage reporting still works fine. After that the only issue there is is the (unrelated?) Travis failure.


After that is fixed there are a some thing that might be worth doing, maybe for another PR:
- Activate isolated builds for tox when setuptools releases a version fixing the setup.py bug.
- Clean jenkins' scripts, I'm not sure why `coverage` is used the way it's used, among other things.
- Clean Makefile: is all that's there actually used?
- Use [this script](https://gist.github.com/eli-collins/ef91add2bc635ccaf61738cd805cb18b) to filter which environments are run on appveyor, otherwise tox detects all the python versions installed and executes all tests.

psafont:
Coverage reporting seems to work find according to https://codecov.io/gh/common-workflow-language/cwltool/pull/961?src=pr&el=desc (80.43%, +2.23%)

mr-c:
Would including `setup.py` in MANIFEST.in be sufficient to fix the setuptools bug?

psafont:
I also was thinking if we would want to delete of `release-test.sh`: the test of building the package in an  isolated virtual environment and installing it in another is already being done with tox :)

Makefile
```diff
@@ -75,10 +67,8 @@ dist/${MODULE}-$(VERSION).tar.gz: $(SOURCES)
 
 ## clean       : clean up all temporary / machine-generated files
 clean: FORCE
-	rm -f ${MODILE}/*.pyc tests/*.pyc
+	git clean -ffdx || true
```
mr-c: No, this will delete too much . `git clean -fdx` is appropriate for `distclean` but not `clean`

psafont:
Code coverage on codecov seems to be unaffected by parallellizing the tests, compared to https://github.com/common-workflow-language/cwltool/pull/961#issuecomment-435844827

PR #964, opened 2018-11-07
WIP: Avoid file location conflicts in output

### LINKED ISSUE #925: Plan for dealing with conflicting filenames in the final output
1. determine final output locations
2. if duplicate locations for different files conflict with each other; or there is a conflict with an existing file then
  a. do not rename the files
  b. but put them in uniquely named subdirectories. Suggested: the output name, followed by "-1" or "-2" et cetera if needed.

Exception: For InplaceUpdate Files & Directories don't do any renaming

psafont:
Right now if two outputs have the same location, only one is written to disk: closes #925

### DIFF
--- a/cwltool/process.py
+++ b/cwltool/process.py
@@ -225,12 +225,56 @@ def adjustFilesWithSecondary(rec, op, primary=None):
         for d in rec:
             adjustFilesWithSecondary(d, op, primary)
 
+def _stage_file(src, dst):
+    if src == dst:
+        return
+    if onWindows():
+        if os.path.isfile(src):
+            shutil.copy(src, dst)
+        elif os.path.isdir(src):
+            if os.path.exists(dst) and os.path.isdir(dst):
+                shutil.rmtree(dst)
+            copytree_with_merge(src, dst)
+    else:
+        os.symlink(src, dst)
 
-def stage_files(pathmapper,             # type: PathMapper
-                stage_func=None,        # type: Callable[..., Any]
-                ignore_writable=False,  # type: bool
-                symlink=True,           # type: bool
-                secret_store=None       # type: SecretStore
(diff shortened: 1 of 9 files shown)


tests/wf/925.cwl
```diff
@@ -0,0 +1,27 @@
...
+  first:
+    run: touch_tool.cwl
+    in:
+      message: file
+    out: [out]
+
+  second:
+    run: touch_tool.cwl
+    in:
+      message: file
```
mr-c: It would be better if the `message` varied between the files. If the output is the same then it is fine to use the same `location`; it is when the contents vary and that data is lost that we have a problem!

psafont: Would it make sense to produce two files in this case even though the contents are the same? Semantically, they are supposed to be two different outputs.

I think it's worth maintaining the test to document this edge case. In any case I'll prepare a test that doesn't use touch.

psafont: I'll document it on the test

cwltool/process.py
```diff
@@ -226,27 +226,60 @@ def adjustFilesWithSecondary(rec, op, primary=None):
...
+                elif os.path.isfile(dst):
+                    os.unlink(dst)
+                shutil.copytree(src, dst)
+            else:
+                shutil.copy2(src, dst)
+    return _stage
+
+
+def stageFiles(pm, stageFunc=_stage_file, ignoreWritable=False, secret_store=None):
     # type: (PathMapper, Callable[..., Any], bool, bool, SecretStore) -> None
```
mr-c: `error: Type signature has too many arguments`
```suggestion
    # type: (PathMapper, Callable[..., Any], bool, SecretStore) -> None
```

PR #971, opened 2018-11-13
fix cid default bug

mr-c:
Regression fix: prevents the creation of `20181113043853-303248.cid` files by default


cwltool/docker.py
```diff
@@ -338,9 +338,9 @@ def create_runtime(self, env, runtimeContext):
         runtime.append(u"--env=HOME=%s" % self.builder.outdir)
 
         # add parameters to docker to write a container ID file
-        if runtimeContext.record_container_id is not None:
+        if runtimeContext.record_container_id is True:
```
psafont: No need to compare a boolean against True, checking its value is enough

PR #974, opened 2018-11-14
Bump ruamel.yaml + fix handling overrides in job order

tetron:


tetron:
Not sure what test to add?  This is from fixing failing tests, so there isn't anything here that wasn't already covered by the tests and/or mypy.

PR #975, opened 2018-11-14
maintenance: refactor code in provenance

psafont:
Remove make_fs_access from ProvenanceProfile
Decouple adding start and stop times to profiles from the actual time it
happens
Decouple adding prospective provenance from the start of a process

This is preparatory work for https://github.com/DataBiosphere/toil/issues/2390

### DIFF
--- a/cwltool/provenance.py
+++ b/cwltool/provenance.py
@@ -475,7 +475,7 @@ class ProvenanceProfile():
         return process_run_id
 
     def start_process(self, process_name, when, process_run_id=None):
-        # type: (Any, Text, datetime.datetime, Optional[str]) -> str
+        # type: (Text, datetime.datetime, Optional[str]) -> str
         """Record the start of each Process."""
         if process_run_id is None:
             process_run_id = uuid.uuid4().urn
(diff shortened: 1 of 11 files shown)


cwltool/provenance.py
```diff
@@ -448,41 +448,34 @@ def evaluate(self,
...
+        elif not hasattr(job, 'workflow'):
+            # commandline tool execution as part of workflow
+            name = str(job.name) if hasattr(job, 'name') else ''
             process_name = urllib.parse.quote(name, safe=":/,#")
-            process_run_id = self.start_process(process_name)
+            process_run_id = self.start_process(process_name, datetime.datetime.now())
         return process_run_id
 
-    def start_process(self, process_name, process_run_id=None):
-        # type: (Any, str, str) -> str
```
stain: Odd change from `str` to `Text` - but this type declaration was wrong to start with and should never have had `Any` - probably it has to be `Text` as a `process_name` ultimately will have come from a YAML parsing of a .cwl file, which can have Unicode `id` values.

mr-c:
Thank you @psafont !

PR #981, opened 2018-11-18
cwl:requirements needs cwlVersion >= v1.1

mr-c:
- [x] remove `cwltool:requirements`, not needed


mr-c:
Jenkins, test this please

(the --parallel / scatter / null bug strikes again!)

mr-c:
@tetron agrees that `cwl:requirements` shouldn’t be available in the reference runner for v1.0 documents as there is an existing cwltool specific feature that accomplishes the same thing (“cwltool:overrides”)

PR #988, opened 2018-11-28
Eliminate unnecessary deep copies of job order objects.

tetron:
Tweak mypy annotations so the method parameter is a read-only Mapping.

Also includes efficiency tweak from #986


tetron:
Reduces memory usage by 50% with really large inputs.

Issue #997, opened 2018-12-04
cwl-runner --singularity fails with inputs of type int


rvanharen:
## Expected Behavior
cwl-runner --singularity supports inputs of type int

## Actual Behavior
cwl-runner --singularity fail with TypeError if input is of type int

## Workflow Code
```
#!/usr/bin/env cwl-runner

cwlVersion: v1.0
class: CommandLineTool
baseCommand: download_gefs.sh
hints:
  DockerRequirement:
    dockerPull: ewatercycle/ewatercycle_forecast_getforcing
requirements:
  EnvVarRequirement:
    envDef:
      ISO_DATE: $(inputs.iso_date)
inputs:
  iso_date:
    type: int
outputs:
  download_gefs:
    type: File
    outputBinding:
glob: gefs_$(inputs.iso_date).tar.bz2
```

## Full Traceback
```pytb
cwl-runner --singularity download_gefs.cwl --iso_date 20181202
/nfs/home2/rhwater/venv-forecast/bin/cwl-runner 1.0.20181201184214
Resolved 'download_gefs.cwl' to 'file:///scratch/shared/tmp1/forecast-docker/getforcing/cwl/download_gefs.cwl'
Using local copy of Singularity image found in /scratch/shared/tmp1/forecast-docker/getforcing/cwl
ISO_DATE
Got workflow error
Traceback (most recent call last):
  File "/nfs/home2/rhwater/venv-forecast/lib/python2.7/site-packages/cwltool/executors.py", line 182, in run_jobs
    job.run(runtime_context)
  File "/nfs/home2/rhwater/venv-forecast/lib/python2.7/site-packages/cwltool/job.py", line 605, in run
    runtime = self.create_runtime(env, runtimeContext)
  File "/nfs/home2/rhwater/venv-forecast/lib/python2.7/site-packages/cwltool/singularity.py", line 283, in create_runtime
    env["SINGULARITYENV_{}".format(name)] = value
  File "/nfs/home2/rhwater/venv-forecast/lib/python2.7/os.py", line 473, in __setitem__
    putenv(key, item)
TypeError: must be string, not int
Workflow error, try again with --debug for more information:
must be string, not int
```

## Your Environment
* cwltool version:  1.0.20181201184214

PR #998, opened 2018-12-04
cwl-runner --singularity fails with inputs of type int #997

rvanharen:
Fixes https://github.com/common-workflow-language/cwltool/issues/997


mr-c:
Jenkins, okay to test

PR #1001, opened 2018-12-08
use subprocess from stdlib on python>=3.5

croth1:
This patch allows dropping the subprocess32 dependency for python>=3.5.

I do not have py3.2-3.5 here, so I haven't tested on those.

### DIFF
--- a/cwltool/singularity.py
+++ b/cwltool/singularity.py
@@ -23,10 +23,17 @@ from .pathmapper import PathMapper, MapperEnt  # pylint: disable=unused-import
 from .pathmapper import ensure_writable, ensure_non_writable
 from .process import UnsupportedRequirement
 from .utils import docker_windows_path_adjust
+
 if os.name == 'posix':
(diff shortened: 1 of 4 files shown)


cwltool/singularity.py
```diff
@@ -23,10 +23,17 @@
...
-        TimeoutExpired)
+    if sys.version_info < (3, 5):
+        from subprocess32 import (  # pylint: disable=import-error,no-name-in-module
+            check_call, check_output, CalledProcessError, DEVNULL, PIPE, Popen,
+            TimeoutExpired)
+    else:
+        from subprocess import (  # pylint: disable=import-error,no-name-in-module
+            check_call, check_output, CalledProcessError, DEVNULL, PIPE, Popen,
+            TimeoutExpired)
+
```
mr-c: Okay. Let's use the subprocess from cwltool.utils then, and not duplicate

PR #1006, opened 2018-12-11
Added memory monitoring for non-dockerised tools

mb1069:


cwltool/job.py
```diff
@@ -385,6 +386,30 @@ def _execute(self,
...
+        monitor = psutil.Process(sproc.pid)
+        memory_usage = []
+
+        def get_tree_mem_usage(container):
+            children = monitor.children()
+            rss = monitor.memory_info().rss
+            while len(children):
+                rss += sum([process.memory_info().rss for process in children])
+                children = list(itertools.chain(*[process.children() for process in children]))
+            container.append(rss)
```
mr-c: I'd rather we only keep the highest memory value, not a super long list

PR #1010, opened 2018-12-13
Udocker mem monitoring fix

mb1069:
Udocker memory was not being stored by monitoring function. Fixed this issue and added test to validate the fix.


mb1069:
CLOSED as feature branch #1006 was not yet merged, so fixes could be applied in that branch before merge to master.

PR #1015, opened 2018-12-14
Versioning with lastest python libs

pjotrp:
These versions pass on my Guix build. You may need to edit some.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,7 +58,7 @@ setup(name='cwltool',
           'six >= 1.9.0',  # >= 1.9.0 required by prov
           'psutil',
           'scandir',
-          'prov == 1.5.1',
+          'prov == 1.5.1, >= 1.5.2',
           'bagit >= 1.6.4',
           'typing-extensions',
       ],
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -58,7 +58,7 @@
           'six >= 1.9.0',  # >= 1.9.0 required by prov
           'psutil',
           'scandir',
-          'prov == 1.5.1',
+          'prov == 1.5.1, >= 1.5.2',
```
mb1069:
```suggestion
          'prov == 1.5.1, >= 1.5.3',
```
Should this be 1.5.3, to match the requirements.txt?

pjotrp:
I have cwltool running with Python 3.7. Had to disable type_extensions with https://github.com/common-workflow-language/cwltool/pull/1015/commits/15539fba76993f951af9eba913bea6d677c74005

mr-c:
@pjotrp Thanks for the update. How shall we proceed? We still support Python before 3.7, and your PR causes the tests to fail :-P

PR #1017, opened 2018-12-17
Update how to setup pip environment

manabuishii:


mr-c:
Jenkins, test this please

CONTRIBUTING.md
```diff
@@ -12,7 +12,8 @@ Here's a rough
 - Create a virtual environment: ``virtualenv cwltool``
 - To begin using the virtual environment, it needs to be activated: ``source bin/activate``
 - To check if you have the virtual environment set up: ``which python`` and it should point to python executable in your virtual env
-- Install cwltool in the virtual environment: ``pip install .``
+- Install cwltool in the virtual environment: ``pip install -e .``
+- Install more tools via pip: ``pip install pytest arcp mock pytest-mock rdflib-jsonld``
```
psafont: I would recommend using `pip install -e .[test]` instead, to do both at the same time

mr-c: @psafont I'm not seeing `test` in the extras list?

https://github.com/common-workflow-language/cwltool/blob/master/setup.py#L65

I agree that it would be good to keep this synchronized so we don't have to manually update this document.

There is also `pip install test-requirements.txt` to make use of https://github.com/common-workflow-language/cwltool/blob/master/test-requirements.txt

psafont: Try it! `setuptools` has its mysterious ways ;)

The tests requirements are defined in https://github.com/common-workflow-language/cwltool/blob/master/setup.py#L74

mr-c:
```
$ pip show pip setuptools
Name: pip
Version: 18.1
[…]
---
Name: setuptools
Version: 40.6.2
[…]
$ pip install -e .[test]
Obtaining file:///home/michael/cwltool
  Installing build dependencies ... done
  cwltool 1.0.20181214034931 does not provide the extra 'test'
```

mr-c (approved):

PR #1035, opened 2019-01-17
Inplace input v1.1

tetron:


tetron:
> Job "step1_2" requested more resources ({'cores': 1, 'ram': 1024}) than are available (max ram is 1359.000000, max cores is 8.000000)

Wat

PR #1040, opened 2019-01-22
add lock for multi-threaded directory creation

ionox0:
@mr-c This has fixed the issue for me, let me know your thoughts.

However it's mostly just to show the issue, because I haven't added any tests for it. 

It seems that the lock needs to be passed into the thread explicitly, I'm no expert with locking in Python, but does that check out with your understanding of how it works?

https://github.com/common-workflow-language/cwltool/pull/1037
https://github.com/common-workflow-language/cwltool/issues/1036


cwltool/job.py
```diff
@@ -414,11 +414,17 @@ def get_tree_mem_usage(memory_usage):
...
 class CommandLineJob(JobBase):
     def run(self,
-            runtimeContext  # type: RuntimeContext
+            runtimeContext,         # type: RuntimeContext
+            TMPDIR_LOCK=None        # type: threading.Lock
             ):  # type: (...) -> None
 
-        if not os.path.exists(self.tmpdir):
-            os.makedirs(self.tmpdir)
+        if TMPDIR_LOCK != None:
```
ionox0: I did this becasue I'm not passing the lock in from the SingleThreadExecutor class

cwltool/job.py
```diff
@@ -414,11 +414,17 @@ def get_tree_mem_usage(memory_usage):
 
 class CommandLineJob(JobBase):
     def run(self,
-            runtimeContext  # type: RuntimeContext
+            runtimeContext,         # type: RuntimeContext
+            TMPDIR_LOCK=None        # type: threading.Lock
```
mr-c:
```suggestion
            tmpdirLock=None        # type: threading.Lock
```

cwltool/job.py
mr-c:
```suggestion
        if tmpdirLock != None:
```

cwltool/job.py
```diff
@@ -414,11 +414,17 @@ def get_tree_mem_usage(memory_usage):
...
             ):  # type: (...) -> None
 
-        if not os.path.exists(self.tmpdir):
-            os.makedirs(self.tmpdir)
+        if TMPDIR_LOCK != None:
+            if not os.path.exists(self.tmpdir):
```
mr-c: isn't this missing `with tmpdirLock:` before it?

ionox0:
I addressed these three comments, however I'm seeing that there is a `JobBase` class that uses the `runtimeContext` which might be a better place for this `tmpdir_lock` to reside. 

I also see that the RuntimeContext has a `workflow_eval_lock`. I believe this would be a better option, rather than creating a new lock just for this tmpdir creation.

cwltool/job.py
ionox0: Yes, my mistake, I had included it in the version I was testing with but didn't commit it here.

mr-c:
My understanding is that it is better for locks to have a narrow, specific purpose. Here we are trying to avoid contention around creating the same directory whereas `workflow_eval_lock` is for managing waiting for workflow components to finish.

mr-c:
Jenkins, okay to test

cwltool/executors.py
```diff
@@ -231,10 +234,10 @@ def select_resources(self, request, runtime_context):  # pylint: disable=unused-
 
         return result
 
-    def _runner(self, job, runtime_context):
+    def _runner(self, job, runtime_context, TMPDIR_LOCK):
         """ Job running thread. """
         try:
-            job.run(runtime_context)
+            job.run(runtime_context, TMPDIR_LOCK)
```
mr-c: `TypeError: run() takes 2 positional arguments but 3 were given`

mr-c:
@ionox0 Can you enable edits to your branch from maintainers? https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

mr-c:
@ionox0 Nevermind, was my error

PR #1048, opened 2019-01-25
Limit loadContents/File literal contents length.

jmchilton:
We decided to limit these and cause a hard error if these buffer sizes were exceeded. We decided this would retro-actively apply to v1.0 - since silently failing (the unspecified but actual behavior of cwltool) seems to be the worst of all behaviors.

Corresponding PR to conformance tests at https://github.com/common-workflow-language/common-workflow-language/pull/822.

### DIFF
--- a/cwltool/builder.py
+++ b/cwltool/builder.py
@@ -18,6 +18,7 @@ from schema_salad.avro.schema import make_avsc_object, Schema
 from schema_salad.sourceline import SourceLine
 from schema_salad.ref_resolver import uri_file_path
 from six import iteritems, string_types
+from typing import IO
 from typing_extensions import (TYPE_CHECKING,  # pylint: disable=unused-import
                                Text, Type)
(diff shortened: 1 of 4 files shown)


jmchilton:
This actually seemed to work first try modulo type linting issues (happy for advice on this if the Travis tells me it is still wrong).

Easy to test against this branch of CWL conformance v1.0 conformance tests https://github.com/common-workflow-language/common-workflow-language/pull/822.

```
cwltest --test conformance_test_v1.0.yaml -n 198
cwltest --test conformance_test_v1.0.yaml -n 199
cwltest --test conformance_test_v1.0.yaml -n 200
cwltest --test conformance_test_v1.0.yaml -n 201
cwltest --test conformance_test_v1.0.yaml -n 202
cwltest --test conformance_test_v1.0.yaml -n 203
```

The controls pass with cwltool from pip but the active tests only pass with the changes in this PR.

mr-c:
Thank you @jmchilton !

PR #1066, opened 2019-02-21
Rename v1.1.0-dev1 requirement ToolTimeLimit

bogdang989:
Should probably not be merged before https://github.com/common-workflow-language/cwl-v1.1/pull/44 (Until CommandLineTool.yml is marked as done)

I only renamed the v1.1.0-dev1 requirement, and left cwltool:TimeLimit for v1.0 to avoid breaking existing tools.


tetron:
Jenkins, test this please

PR #1067, opened 2019-02-21
Fix regression in compute checksum.

tetron:
Should not be affected by the loadContents limit.


mr-c:
A regression fix needs a test

PR #1072, opened 2019-02-26
Input binding.position expr

mr-c:
implements: https://github.com/common-workflow-language/cwl-v1.1/pull/49

For the full emoji experience try https://github.com/eosrei/twemoji-color-font

```
$ cwltool --enable-dev tests/echo-position-expr.cwl tests/echo-position-expr-job.yml 
/home/michael/cwl-v1.1/env3/bin/cwltool 1.0.20190226140034
Resolved 'tests/echo-position-expr.cwl' to 'file:///home/michael/cwl-v1.1/tests/echo-position-expr.cwl'
[job echo-position-expr.cwl] /tmp/_jxl_sgk$ echo \
    🕺 \
    1 \
    singular \
    sensation! > /tmp/_jxl_sgk/out.txt
Could not collect memory usage, job ended before monitoring began.
[job echo-position-expr.cwl] completed success
{
    "out": "🕺 1 singular sensation!\n"
}
Final process status is success
```


tetron:
From chat: v1.0 should not validate expressions, which are strings, so an explicit version test shouln't be necessary.  Add unit test to verify this.

mr-c:
@tetron you were right, thanks!

```
l$ cwltool tests/echo-position-expr.cwl tests/echo-position-expr-job.yml 
/home/michael/cwl-v1.1/env3/bin/cwltool 1.0.20190226140034
Resolved 'tests/echo-position-expr.cwl' to 'file:///home/michael/cwltool/tests/echo-position-expr.cwl'
Tool definition failed validation:
tests/echo-position-expr.cwl:3:1:  Object `tests/echo-position-expr.cwl` is not valid because
                                     tried `CommandLineTool` but
tests/echo-position-expr.cwl:7:1:      * the `inputs` field is not valid because
tests/echo-position-expr.cwl:8:3:          item is invalid because
tests/echo-position-expr.cwl:10:5:           the `inputBinding` field is not valid because
                                               tried CommandLineBinding but
tests/echo-position-expr.cwl:11:7:               the `position` field is not valid
                                                 because
                                                   tried int but
                                                     `'$(self)'` is not int
tests/echo-position-expr.cwl:17:1:     * the `arguments` field is not valid because
                                           tried array of <string or Expression or
                                           CommandLineBinding> but
tests/echo-position-expr.cwl:18:5:           item is invalid because
                                               tried CommandLineBinding but
                                                 the `position` field is not valid because
                                                   tried int but
                                                     `'${return 2;}'` is not int
```

cwltool/builder.py
```diff
@@ -211,7 +211,13 @@ def bind_input(self,
 
             bp = list(aslist(lead_pos))
             if "position" in binding:
-                bp.extend(aslist(binding["position"]))
+                position = binding["position"]
+                if isinstance(position, str):  # no need to test the CWL Version
+                                               # the schema for v1.0 only allow ints
+                    binding['position'] = self.do_eval(position, context=datum)
```
tetron: You can simplify this whole thing to just `position = self.do_eval(binding["position"], context=datum)` because if the expression isn't a string it will just return it as-is.

cwltool/process.py
```diff
@@ -722,7 +722,14 @@ def inc(d):  # type: (List[int]) -> None
                 if isinstance(arg, MutableMapping):
                     arg = copy.deepcopy(arg)
                     if arg.get("position"):
-                        arg["position"] = [arg["position"], i]
+                        position = arg.get("position")
```
tetron: ditto

cwltool/process.py
```diff
@@ -722,7 +722,14 @@ def inc(d):  # type: (List[int]) -> None
...
+                        if isinstance(position, str):  # no need to test the
```
tetron: ditto

cwltool/builder.py
tetron: but there should be a typecheck on the return value

mr-c: I'd rather keep the `isinstance`

mr-c:
Jenkins, test this please

PR #1073, opened 2019-02-27
Add runtime.exitCode to outputEval

tetron:


tetron:
Probably needs a version check so it doesn't get set on v1.0 documents...

mr-c:
Combining with https://github.com/common-workflow-language/cwltool/pull/1072

Issue #1074, opened 2019-02-28
in an ExpressionTool, the `File`s inside a `Directory` lack path part fields


mr-c:
Though a regular input of type `File` has `basename`, `nameroot`, `nameext` fields..

The problem exists even with `cwltool:LoadListingRequirement: { loadListing: deep_listing }`

>biostars_365953.cwl
``` cwl
class: ExpressionTool
cwlVersion: v1.0
$namespaces: { cwltool: "http://commonwl.org/cwltool#" }
inputs:
  randodir: Directory
  randofile: File
outputs:
  exts: string[]
requirements:
  InlineJavascriptRequirement: {}
  cwltool:LoadListingRequirement:
    loadListing: deep_listing
expression: |
  ${
    var exts = [];
    
    exts.push(inputs.randofile.nameext)
    for (var i = 0; i < inputs.randodir.listing.length; i++) {
      var file = inputs.randodir.listing[i];
        exts.push(file.nameext);
      }
    return {"exts": exts};
   }
```
```
$ mkdir -p test1/test2
$ touch test1/first.txt test1/test2/second.foo test1/test2/third.bar
$ cwltool --enable-ext biostars_365953.cwl --randofile test1/first.txt --randodir test1/test2
/home/michael/cwltool/env3/bin/cwltool 1.0.20190227152456
Resolved 'biostars_365953.cwl' to 'file:///home/michael/cwltool/biostars_365953.cwl'
{
    "exts": [
        ".txt",
        null,
        null
    ]
}
Final process status is success
```
Version of the same, but putting the `expression` inside a `CommandLineTool`:
> biostars_365953_clt.cwl

``` cwl
class: CommandLineTool
cwlVersion: v1.0
inputs:
  randofile: File
  randodir: Directory
outputs:
  exts:
   type:
     type: array
     items: string
   outputBinding:
     outputEval: |
       ${
        var exts = [];
        exts.push(inputs.randofile.nameext)
        for (var i = 0; i < inputs.randodir.listing.length; i++) {
          var file = inputs.randodir.listing[i];
            exts.push(file.nameext);
          }
        return exts;
        }

requirements:
  InlineJavascriptRequirement: {}
baseCommand: /bin/true

```

```
$ cwltool --enable-ext biostars_365953_clt.cwl --randofile test1/first.txt --randodir test1/test2
/home/michael/cwltool/env3/bin/cwltool 1.0.20190227152456
[...]

PR #1075, opened 2019-02-28
normalize directory entries before executing an ExpressionTool

mr-c:
Fixes https://github.com/common-workflow-language/cwltool/issues/1074

- [ ] needs a confirmation test


tetron:
Thanks!  One of our users just reported this bug in arvados-cwl-runner and I hadn't investigated, but I thought it was probably something like this.

Issue #1081, opened 2019-03-15
RFC handling of umask for temporary directories


tiborsimko:
## Expected Behavior

Temporary directories created by `cwltool` should respect the `umask` settings of the parent environment. For example, when I set `umask 002` before running `cwltool`, I would expect `cwltool` to create any files or directories according to this umask. 

## Actual Behavior

The `cwltool` seems to impose fixed 755 permissions in several places:

```console
$ git grep -Hni 0o0
cwltool/docker.py:262:                os.makedirs(host_outdir_tgt, 0o0755)
cwltool/job.py:536:            os.makedirs(dirname, 0o0755)
cwltool/process.py:228:            os.makedirs(os.path.dirname(entry.target), 0o0755)
cwltool/process.py:245:            os.makedirs(entry.target, 0o0755)
cwltool/process.py:251:                os.makedirs(entry.target, 0o0755)
cwltool/singularity.py:229:            os.makedirs(new_dir, 0o0755)
```

## Motivation

When running CWL workflows on the containerised [REANA](http://www.reana.io/) reproducible analysis platform, there are situations where the user wants to be able to run the Docker image under some particular non-root user identity (say UID=501). In order to guarantee being able to write to the mounted workspace in the job container, the user (1) belongs to a particular group (say GID=0 following [OpenShift guidelines](https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#openshift-specific-guidelines)) and (2) permissions on the workspace are set  via `umask 002` to be of the `drwxrwxr-x` nature. This is usually enough to address the need.  However, we have encountered an issue that `cwltool` seems to create some directories with fixed 755 permissions (see above), which fails. It would be good `cwltool` respects the parent umask setting or if this behaviour were configurable.

## RFC

Is there a particular reason for imposing fixed 755 in `cwltool`? Could we simply drop it and rely on the umask set by the parent environment? Alternatively, could we make it configurable?

tetron:
I don't see anything wrong with that.  Please provide a PR that changes them to 0o0777 with appropriate tests.

tiborsimko:
Closing as the PR was merged.

PR #1082, opened 2019-03-18
Use default mask in os.makedirs() (#1081)

tiborsimko:
* Amends `os.makedirs()` calls to use the default mask value (`0o777`) so that
  temporary directories and files created by `cwltool` would better respect the
  umask settings from the original environment. For example, the parent
  environment might use umask 002 to make sure that all directories and files
  are created with group-write permissions.

Signed-off-by: Tibor Simko <tibor.simko@cern.ch>


tiborsimko:
@tetron This implements what we discussed in #1081.  Instead of forcing `0o0777` I have simply removed the mode argument, since `0o0777` is the default for [os.makedirs()](https://docs.python.org/3/library/os.html#os.makedirs). Also, I'm not sure that a new dedicated test would be necessary, since the Python defaults are being used, so I simply "high-jacked" one existing test and added a umask assert on a produced file.  Please let me know if you'd like to see something more elaborate.

tetron:
https://travis-ci.org/common-workflow-language/cwltool/jobs/507749894#L747

```
AssertionError: assert '0664' == '0o664'
```

Setting/resetting umask should be in a `try: ... finally:` block

There are other test failures.  I'm not sure if they are related to this umask change.  Something may have changed upstream to cause tests to fail.  I'll have to investigate.

tiborsimko:
Thanks, the test rewritten to be both Python 2 and 3 compatible, the branch repushed. I also noticed that many tests in the `master` branch are failing... I have not investigated. In this PR, I simply run `pytest tests/test_examples.py::TestSecondaryFiles` to run only the wanted test.

tetron:
jenkins, test this please

PR #1106, opened 2019-04-25
"checklinks" flag should not be based on do_validate

tetron:
This flag is used internally by schema salad to determine when it is
appropriate to run checklinks, but checklinks must not be skipped
entirely as it includes updating fields in certain circumstances.

### DIFF
--- a/cwltool/load_tool.py
+++ b/cwltool/load_tool.py
@@ -282,8 +282,7 @@ def resolve_and_validate_document(loadingContext,
         _add_blank_ids(workflowobj)
 
     workflowobj["id"] = fileuri
-    processobj, metadata = document_loader.resolve_all(
-        workflowobj, fileuri, checklinks=loadingContext.do_validate)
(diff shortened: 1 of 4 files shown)


tetron:
jenkins, test this please

mr-c:
Can we get a test case that demonstrates the need for this change?

cwltool/load_tool.py
```diff
@@ -219,6 +219,11 @@ def resolve_and_validate_document(loadingContext,
     cwlVersion = loadingContext.metadata.get("cwlVersion")
     if not cwlVersion:
         cwlVersion = workflowobj.get("cwlVersion")
+    if not cwlVersion and fileuri != uri:
+        # The tool we're loading is a fragment of a bigger file.  Get
+        # the document root element and look for cwlVersion there
+        metadata = fetch_document(fileuri, loadingContext)[1]
+        cwlVersion = metadata.get("cwlVersion")
```
mr-c: This still has no test coverage

PR #1111, opened 2019-05-04
Singularity2to3

drkennetz:
Added brute force method to make this functional in singularity 2.x and singularity 3.x

PR #1113, opened 2019-05-08
Singularity23 fix

drkennetz:
I added a fun `_normalize_sif_id()` which creates a regex based on how singularity3.x will write out image files. 
Also added logic which will search for .img or .sif in $SINGULARITY_CACHEDIR and $SINGULARITY_PULLFOLDER only if they exist in users env. If they do not exist, the file will be written locally (this is how you previously had it). If the cwltool is run again, it will use the local copy if it is found in any of the locations `pwd`, `$SINGULARITY_CACHEDIR`, or `$SINGULARITY_PULLFOLDER`. If those are not in env, the image must be in the same directory as the tool in order to be found.

(continuation of https://github.com/common-workflow-language/cwltool/pull/1111)

It is also important to note that any images or sif's generate using singularity ad hoc (without cwl) that follow the singularity naming convention will also be found if they are written to these directories, which I think is desirable behavior.


cwltool/singularity.py
```diff
@@ -79,39 +86,55 @@ def get_image(dockerRequirement,  # type: Dict[Text, Text]
 
         if "dockerImageId" not in dockerRequirement and "dockerPull" in dockerRequirement:
             match = re.search(pattern=r'([a-z]*://)', string=dockerRequirement["dockerPull"])
-            candidate = _normalize_image_id(dockerRequirement['dockerPull'])
-            candidates.append(candidate)
-            dockerRequirement['dockerImageId'] = candidate
+            candidate_image = _normalize_image_id(dockerRequirement['dockerPull'])
+            candidates.append(candidate_image)
+            candidate_sif = _normalize_sif_id(dockerRequirement['dockerPull'])
+            candidates.append(candidate_sif)
```
drkennetz: This is currently only being added to `candidates[]` for searching but this could also be used in the future if singularity version is specified. For example, the .sif file is 3x specific, so something like if `if version == '3.x': candidate = _normalize_sif_id; else: candidate = _normalize_image_id`

cwltool/singularity.py
```diff
@@ -79,39 +86,55 @@ def get_image(dockerRequirement,  # type: Dict[Text, Text]
...
             if "dockerPull" in dockerRequirement:
-                cmd = ["singularity", "pull", "--force", "--name",
-                       str(dockerRequirement["dockerImageId"]),
-                       str(dockerRequirement["dockerPull"])]
-                _logger.info(Text(cmd))
-                check_call(cmd, stdout=sys.stderr)
-                found = True
+                if "SINGULARITY_PULLFOLDER" in os.environ:
+                    pull_folder = str(os.environ["SINGULARITY_PULLFOLDER"])
+                    cmd = ["singularity", "pull", "--force", "--name",
```
drkennetz: if image is not found, and $SINGULARITY_PULLFOLDER is in environment, write the new image to this folder. Else write it locally.

drkennetz:
@mr-c here is the new PR whenever you get some time. Let me know if you need anything from me!

.travis.yml
```diff
@@ -39,5 +39,6 @@ script: tox
 branches:
   only:
     - master
+    - singularity23_fix
```
mr-c:
```suggestion
```

cwltool/singularity.py
mr-c: I think we established that 2.x uses `.img` and 3.x uses `.sif`?

mr-c:
Jenkins, Okay to test

mr-c:
Jenkins, okay to test

cwltool/singularity.py
drkennetz: We did, but there is no code which checks for singularity version at runtime. I currently have the code checking for both .img and .sif because of this. In the future we can choose if we'd like to search only for .sif or .img depending on singularity version.

cwltool/singularity.py
```diff
@@ -79,39 +86,55 @@ def get_image(dockerRequirement,  # type: Dict[Text, Text]
...
         if (force_pull or not found) and pull_image:
             cmd = []  # type: List[Text]
             if "dockerPull" in dockerRequirement:
-                cmd = ["singularity", "pull", "--force", "--name",
-                       str(dockerRequirement["dockerImageId"]),
-                       str(dockerRequirement["dockerPull"])]
-                _logger.info(Text(cmd))
-                check_call(cmd, stdout=sys.stderr)  # nosec
-                found = True
+                if "SINGULARITY_PULLFOLDER" in os.environ:
```
mr-c: This section lacks test coverage

drkennetz: Would you like me to write a test for this?

drkennetz:
We will likely have access to singularity2.6 on our cluster in the next few days. I can work on version specific code if that seems like something you'd like to have. I will run some tests to see if Singularity3.x is backwards compatible with 2.x. If it is, I can write code to allow sing3.x to run 2.x and 3.x images, while limiting 2.x to Sing2.x specific images.

cwltool/singularity.py
mr-c: That'd be great, thanks!

drkennetz:
I'm learning a lot of git! I pulled the changes you made locally and then added my tests and pushed them now. The tests I added should test for images to be created in SINGULARITY_PULLFOLDER and locally depending on if the environmental variable exists. We shall see.

drkennetz:
It seems that the singularity_pull_folder test failed, perhaps I don't understand what get_data() does, but I figured it just ran the cwltool I passed it and would output anything that the tool would output. The tool writes an image file and that's it. Inside of the test directory I created a singularity_pullfolder and set it in env, and then checked to see if the .img was in the folder I designated as pull_folder. Let me know if you see anything wrong with the test!

tests/sing_pullfolder_test.cwl
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env cwl-runner
+cwlVersion: v1.0
+class: CommandLineTool
+
+requirements:
+  DockerRequirement:
+    dockerPull: debian
```
drkennetz: I used this because I saw that you had it in another tool you use for testing. It is only to test for .img filename generation. Does this look okay?

tests/test_singularity.py
```diff
@@ -41,11 +41,14 @@ def test_singularity_incorrect_image_pull():
     assert result_code != 0
```
drkennetz: The below caused all tests to pass but did not increase code coverage. So there now appears to be a test in place to determine that the code does in fact produce an image locally if no env variable for $SINGULARITY_PULLFOLDER is set or in the directory if the variable is set, but it did not increase code coverage. Any ideas?

drkennetz:
@mr-c I realized that after the final test this morning. My final check was to move the code where it creates the SINGULARITY_PULLFOLDER env to the top, because I realized last night that I was creating a local image before doing this and this portion of the code was not being entered by test. So at first I figured that was the issue, then I moved the test to the top of the test_singularity.py and it still remained uncovered so I figured the env was persisting. Thanks for validating that. Sorry for the 15 commits between last night and today it just took me multiple commits to actually figure out what was going wrong. 

Are you okay with me committing more as I figure out the try except?

mr-c:
No worries, just wanted to share my thinking. Go for it!

drkennetz:
Is the default branch failing because commits have been made?

mr-c:
@drkennetz No, that was a transient error on the Jenkins server

tests/test_singularity.py
```diff
@@ -13,6 +14,42 @@
 sys.argv = ['']
 
 
+@needs_singularity
+def test_singularity_pullfolder(tmp_path):
+    for directory, subdir, files in os.walk(os.getcwd()):
+        for image in files:
+            if image.startswith("debian"):
+                os.remove(os.path.join(directory, image))
+                print("removed image hiding away deep within!")
```
mr-c: Oh, lets not do this.

If it seems necessary, then change to the `tmp_path` (which is unique for every invocation of this function) and copy the CWL descriptions there before executing

tests/test_singularity.py
```diff
@@ -13,6 +14,42 @@
...
+
+    if "SINGULARITY_PULLFOLDER" in os.environ:
+        del os.environ["SINGULARITY_PULLFOLDER"]
```
mr-c: Great, now that the tests work we can refactor `get_main_output` to include a custom `env` instead of deleting from the global `os.environ`

mr-c: I can help with that

mr-c:
Almost there!

tests/test_singularity.py
drkennetz: This sounds good!

tests/test_singularity.py
drkennetz: Haha, this never seemed like a great idea to me but I wanted to give it a try to see if debian.img was actually persisting. I will remove that bit and push the change. It won't break the pipeline.

drkennetz:
I will let you resolve the get_data issue then we can see if this will increase cov.

Thanks for your help!

drkennetz:
@mr-c have you had a chance to review this at all? I believe I am waiting on your update to the get_data() function to increase coverage. After that, PR complete?

mr-c:
In the latest commit I've changed how the images are named, only `/` is replaced and no components are dropped

https://github.com/common-workflow-language/cwltool/commit/309919e593380c5c331c50b655d024a5400a7d2f

This way same named images from different organizations or registries aren't mistaken for each other.

@drkennetz This is different than `singularity`'s behavior, but I and others feel it is more correct and more useful. What do you think?

Thia probably means that we shouldn't use the `SINGULARITY_PULLFOLDER` but I want to study that some more

drkennetz:
@mr-c I think that is a good idea. I don't think you should strip any part of the name to keep them as unique as possible. However, I don't know why that would conflict with `SINGULARITY_PULLFODLER` as that is just a storage space. 

Singularity by default writes images to 

`$SINGULARITY_CACHEDIR/oci-tmp/<image_hash>/fullimage-bwa.version.sif`.

However, you have changed this behavior by including an image name and a write location to singularity image files in `cwd()`. The naming convention shouldn't matter. As long as you provide a name to `dockerPull:` that corresponds to the regex that your code will output, it should find the image in `SINGULARITY_PULLFOLDER` or `SINGULARITY_CACHEDIR`. Does this make sense?

Maybe I am misunderstanding you or what you are trying to change. I have a pretty intimate understanding of how cwl and singularity interact. Maybe you could tell me what you are trying to have as default behavior and I could comment on that or provide possible coding changes.

drkennetz:
@mr-c I think I misunderstood you the first time. If you guys want to change `$SINGULARITY_PULLFOLDER` to something like `$CWL_PULLFOLDER` because the image names will have different naming conventions than singularity's I am completely okay with that. You guys just let me know how you would like to rearchitect and I will comply.

Maybe we could bring it up in the next meeting.

mr-c:
Seems that Singularity 3.1 and 3.2 don't use the SINGULARITY_PULLFOLDER environment variable :crying_cat_face:

drkennetz:
> Seems that Singularity 3.1 and 3.2 don't use the SINGULARITY_PULLFOLDER environment variable 😿

I do see that and realize it is something we configured for singularity to write images to! The two options we could search from are SINGULARITY_CACHEDIR if version 3+ since you added this and SINGULARITY_LOCALCACHEDIR for 2.6. I can make those changes quickly if you are okay with that? I know you mentioned using an external directory because the naming changed. I am okay with whatever?

PR #1115, opened 2019-05-14
Introduced reference to Debian+Conda

smoe:
Just followed a link from the user guide - this should mention the Debian package that cwltool provides via the official Debian channels. I tend to think that the way this was formulated is very vendor-neutral.


mr-c:
Thanks!

PR #1116, opened 2019-05-21
Add an option to mount the docker socket volume.

arostamianfar:
This is useful for running docker within a docker image through sibling containers.

I realize that the nested docker run command can be implemented as a separate task, but there are situations where having this feature dramatically simplifies the pipeline.
Example use case: one of the tasks in a large pipeline calls a script that's packaged inside another docker image. The inputs/outputs are complex and the docker script needs to run several times for each task. Extracting the inner docker call to its own task adds a lot of unnecessary overhead to both the pipeline (passing many inputs/outputs that are not needed by other parts of the pipeline) and the workflow engine (needs to keep track of many additional smaller tasks for each input).

### DIFF
--- a/cwltool/argparser.py
+++ b/cwltool/argparser.py
@@ -292,6 +292,10 @@ def arg_parser():  # type: () -> argparse.ArgumentParser
     parser.add_argument("--force-docker-pull", action="store_true",
                         default=False, help="Pull latest docker image even if"
                                             " it is locally present", dest="force_docker_pull")
+    parser.add_argument("--mount-docker-socket-volume", action="store_true",
+                        default=False, help="Mount the /var/run/docker.sock volume when running the "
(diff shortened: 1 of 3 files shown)


mr-c:
Hello @arostamianfar , thank you for this PR and for your work over at https://github.com/DataBiosphere/toil/ !

However such an option would render the workflow non-portable to other CWL runners besides this one (and exposing the docker socket is not listed in the guarantees of the CWL spec). If you want to have a conversation about your particular situation over at https://www.biostars.org/t/cwl/ just drop the link to your post here and I'll have a look.

Cheers,

arostamianfar:
@mr-c thanks! That makes sense. I'll think of other workarounds for our specific use case and may post on biostars if I have any questions. Thanks again!

Issue #1110, opened 2019-05-03
CWLtools throwing error upon usage of secondary files for capturing log files from star aligner on two different versions


ranijames:
Dear All,

I have tried using java expression to capture all secondary files from STAR aligner in CWLtools version from 2018 October and 2019 February.

 **On 2018 October version it is running successfully without any issues**, whereas on **2019 February version it is throwing error.**

Any suggestions help is much appreciated!!
Thank you

## Actual Behavior

and on 2019 February version the same tool is throwing following error,

   ```
 Error collecting output for parameter 'star_bam':
star.cwl:172:2: Traceback (most recent call last):
star.cwl:172:2: 
star.cwl:172:2:   File "/cluster/home/user/software/anaconda/envs/tupro/lib/python2.7/site-packages/cwltool/command_line_tool.py", line 624, in collect_output_ports
star.cwl:172:2:     compute_checksum=compute_checksum)
star.cwl:172:2: 
star.cwl:172:2:   File "/cluster/home/user/software/anaconda/envs/tupro/lib/python2.7/site-packages/cwltool/command_line_tool.py", line 786, in collect_output
star.cwl:172:2:     primary["secondaryFiles"].append(sfitem)
star.cwl:172:2: 
star.cwl:172:2:   File "/cluster/home/user/software/anaconda/envs/tupro/lib/python2.7/site-packages/schema_salad/sourceline.py", line 247, in __exit__
star.cwl:172:2:     raise self.makeError("\n".join(traceback.format_exception(exc_type, exc_value, tb)))
star.cwl:172:2: 
WorkflowException: star.cwl:182:3: Traceback (most recent call last):
star.cwl:182:3: 
star.cwl:182:3:   File "/cluster/home/user/software/anaconda/envs/tupro/lib/python2.7/site-packages/cwltool/command_line_tool.py", line 765, in collect_output
star.cwl:182:3:     if "$(" in sf["pattern"] or "${" in sf["pattern"]:
star.cwl:182:3: 
star.cwl:182:3: TypeError: string indices must be integers
[job star] completed permanentFail
[job star] {}
[step star] Output is missing expected field file:///cluster/work/lab/projects/temp/fastq_temp/pipeline_part1.cwl#star/star_bam
[step star] produced output {}
[step star] completed permanentFail
[workflow ] completed permanentFail
[workflow ] {
    "alignment_out": null
}
[job star] Removing input staging directory /tmp/tmpS6xqCk
[...]

PR #1117, opened 2019-05-22
Fix secondaryFile as single string/expression in CWL1.0

bogdang989:
Fixes https://github.com/common-workflow-language/cwltool/issues/1110

### DIFF
--- a/cwltool/builder.py
+++ b/cwltool/builder.py
@@ -313,9 +313,6 @@ class Builder(HasReqsHints):
                         else:
                             sf_required = True
 
-                        # Convert to secondaryFileSchema for cw1.0 tools
-                        if isinstance(sf, str):
-                            sf = {"pattern": sf}
 
                         if "$(" in sf["pattern"] or "${" in sf["pattern"]:
                             sfpath = self.do_eval(sf["pattern"], context=datum)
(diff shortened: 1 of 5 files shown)


mr-c:
Jenkins, test this please

cwltool/builder.py
```diff
@@ -313,6 +313,10 @@ def _capture_files(f):
                         else:
                             sf_required = True
 
+                        # Convert to secondaryFileSchema for cw1.0 tools
+                        if isinstance(sf, str):
+                            sf = {"pattern": sf}
```
mr-c: I think this indicates an error in the v1.1 schema

tests/test_examples.py
```diff
@@ -805,6 +805,21 @@ def test_secondary_files(self):
         assert "completed success" in stderr
         assert error_code == 0
 
+    def test_secondary_files_v1_0(self):
+        test_file = "secondary-files-string-v1.cwl"
+        test_job_file = "secondary-files-job.yml"
+        try:
+            old_umask = os.umask(stat.S_IWOTH)  # test run with umask 002
+            error_code, _, stderr = get_main_output(
+                ["--enable-dev",
```
mr-c: why `--enable-dev` here?

bogdang989:
Hi @mr-c  thanks for reviewing. I took a closer look and found that the proper place to fix this would be in update.py

tests/test_examples.py
bogdang989: Yup, that's unnecessary :)

tetron:
Thanks!

Issue #1121, opened 2019-05-28
Can't return 0 for optional int/float output


bogdang989:
## Expected Behavior
Tool output should be 0

## Actual Behavior
Tool output is none 

## Workflow Code
```
cwlVersion: v1.0
class: CommandLineTool
baseCommand:
  - echo
  - "0"
stdout: a.txt
requirements:
- class: InlineJavascriptRequirement
inputs: []
outputs:
  idx:
    type: float?
    outputBinding:
      glob: "a.txt"
      loadContents: true
      outputEval: "${\n    return parseFloat(self[0].contents);\n}"

```

## Full Traceback
```pytb
completed success
{
    "idx": null
}
Final process status is success
```

## Your Environment
* cwltool version:  1.0.20190228155703
Check using ``cwltool --version``

PR #1122, opened 2019-05-28
Don't convert 0 to null on optional output

bogdang989:
Fix for https://github.com/common-workflow-language/cwltool/issues/1121


mr-c:
Thanks! Can we get a `cwltool` test for this? And a conformance test as well?

bogdang989:
Sure thing. I guess conf test should go for cwl1.1?

PR #1123, opened 2019-05-28
Requires schema-salad 4.2 for cwl v1.1 support.

tetron:


tetron:
jenkins, test this please

PR #1130, opened 2019-06-07
Pinned psutil <= 5.6.1

mb1069:


requirements.txt
```diff
@@ -8,7 +8,7 @@ pathlib2==2.3.2; python_version<"3"
 prov==1.5.1
 bagit==1.6.4
 mypy-extensions
-psutil
+psutil<=5.6.1
```
mr-c: If it was a bad release then we should ban just that release, and not all newer releases as well: 

```suggestion
psutil != 5.6.1
```

mr-c:
Superseded by https://github.com/common-workflow-language/cwltool/pull/1457 ; thanks!

PR #1132, opened 2019-06-10
Experimental feature to change substitution prefix.

tetron:
Example of combining this feature with attachments (https://github.com/common-workflow-language/schema_salad/pull/254) to simplify wrapping an inline bash script:

```
#!/usr/bin/env cwl-runner
cwlVersion: v1.0
class: CommandLineTool
$namespaces:
  cwltool: "http://commonwl.org/cwltool#"
baseCommand: [/bin/sh, script]
requirements:
  cwltool:SubstitutionPrefix:
    substitutionPrefix: "@"
  InitialWorkDirRequirement:
    listing:
      - entryname: script
        entry: {$include: "#attachment-1"}
inputs:
  msg: string
outputs:
  output:
    type: File
    outputBinding:
      glob: output.txt
--- |
echo Zip@(inputs.msg) $(date) > output.txt
```


mr-c:
For your example I would demonstrate use of `EnvVarRequirement` and turn off the substitution altogether

TMiguelT:
I like it. I think having this would solve a lot of issues.

However I still believe `InitialWorkDirRequirement` is not the most logical place for the inclusion of scripts. We still need a construct that is intuitively for including scripts and setting them as the tool's command, which is where my `ScriptTool` proposal etc come in. That said we could reasonably add this `substitutionPrefix` feature in independently from any better mechanism for referring to scripts.

PR #1133, opened 2019-06-10
Toolfactory

tetron:
The ToolFactory is a new process type which performs two steps.  First, it executes a "factory" process (command line tool or workflow) passing through the entire input document.  The "factory" process runs and produce an output file `runProcess` and optionally `runInputs`.  The file in `runProcess` must be a CWL document.  For the second step the file in runProcess is loaded into the current engine and executed, either with `runInputs` or passing through the original inputs.  The final output of the tool factory is the output from running the generated inner process.

In this proof-of-concept, a python script called tfbash.py takes an annotated bash script and produces a CWL wrapper on the fly.

```
cwlVersion: v1.0
$namespaces:
  cwltool: "http://commonwl.org/cwltool#"
class: ToolFactory
inputs: {}
outputs: {}
run:
  class: CommandLineTool
  inputs:
    processSource: File
    tfbash_py:
      type: File
      default:
        class: File
        location: tfbash.py
  outputs:
    runProcess:
      type: File
      outputBinding:
        glob: main.cwl
    runInputs:
      type: Any
      outputBinding:
        outputEval: $(inputs)
  arguments: [$(inputs.tfbash_py), $(inputs.processSource)]
  stdout: main.cwl
```


README.rst
```diff
@@ -145,6 +145,11 @@ the default cwl-runner use::
 
   cwltool [tool-or-workflow-description] [input-job-settings]
 
+You can set cwltool options in the environment with CWLTOOL_OPTIONS,
+these will be inserted at the beginning of the command line::
+
+  export CWLTOOL_OPTIONS="--debug"
+
```
mr-c: Can you pull this out to a separate PR?

cwltool/command_line_tool.py
```diff
@@ -169,9 +169,9 @@ def revmap_file(builder, outdir, f):
 
         if revmap_f and not builder.pathmapper.mapper(revmap_f[0]).type.startswith("Writable"):
             f["location"] = revmap_f[1]
-        elif uripath == outdir or uripath.startswith(outdir+os.sep):
+        elif uripath == outdir or uripath.startswith(outdir+os.sep) or uripath.startswith(outdir+'/'):
```
mr-c: Does this fix a bug? I'd like to see a unit test that this change fixes.

I'm confused as to why it is `os.sep` and not always `'/'`

cwltool/extensions.yml
```diff
@@ -135,3 +135,22 @@ $graph:
         "_type": "@vocab"
     - name: networkAccess
       type: [boolean, string]
+
+- name: WorkflowGenerator
```
mr-c: Or maybe "ProcessGenerator?

cwltool/job.py
```diff
@@ -543,7 +543,7 @@ def create_file_and_add_volume(self,
             tmp_dir, tmp_prefix = os.path.split(tmpdir_prefix)
             new_file = os.path.join(
                 tempfile.mkdtemp(prefix=tmp_prefix, dir=tmp_dir),
-                os.path.basename(volume.resolved))
+                os.path.basename(volume.target))
```
mr-c: Another bug fix? Again, would like to see a unit test & conformance test that triggers this bug.

cwltool/main.py
```diff
@@ -542,26 +695,7 @@ def main(argsl=None,                   # type: Optional[List[str]]
...
-            rdflib_logger.setLevel(logging.DEBUG)
-        fmtclass = coloredlogs.ColoredFormatter if args.enable_color else logging.Formatter
-        formatter = fmtclass("%(levelname)s %(message)s")
-        if args.timestamps:
-            formatter = fmtclass(
-                "[%(asctime)s] %(levelname)s %(message)s",
-                "%Y-%m-%d %H:%M:%S")
-        stderr_handler.setFormatter(formatter)
-        ##
+        configure_logging(args, stderr_handler, runtimeContext)
```
mr-c: I love this refactoring, can that be split out into a separate PR?

tests/test_wfgenerator.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+import os
+from .util import get_data, windows_needs_docker
+from cwltool.main import main
+
+@windows_needs_docker
+def test_missing_enable_ext():
+    # Requires --enable-ext and --enable-dev
+    if "CWLTOOL_OPTIONS" in os.environ:
+        del os.environ["CWLTOOL_OPTIONS"]
```
mr-c: Probably should restore this at the end of the test, to avoid side effects

README.rst
tetron: Sure

cwltool/command_line_tool.py
tetron: It was failing on Appveyor.  For some reason it was mixing `/` and `\` for path separators, and os.sep is `\` on Windows.  But then I added windows_requires_docker which causes that code to be skipped completely.  So I think there may be a Windows specific bug, and this is a brute force workaround.

cwltool/extensions.yml
tetron: That's even better.

cwltool/job.py
tetron: This is definitely a bug fix, but I have to go back and look at exactly what was happening.

tests/test_wfgenerator.py
tetron: Yes, you're right.

README.rst
tetron: https://github.com/common-workflow-language/cwltool/pull/1205

cwltool/command_line_tool.py
tetron: So, I believe this fixes a legitimate bug, and that mostly things don't fail on appveyor because we skip those tests, not because it was actually working before.  However I don't have access to a windows install to probe it thoroughly.  So I can leave it in as a basically harmless fix or take it out and know what we probably have a windows specific bug (that won't get fixed until someone else digs into it more deeply.)

tetron:
jenkins, test this please

PR #1138, opened 2019-06-20
Symlink err

tetron:
Report if there is a name conflict staging files (to assist in debugging errors like "cannot symlink XXX to YYY: File exists")


tetron:
jenkins, test this please

PR #1169, opened 2019-08-13
Add input_required flag to main().

tetron:
When false, command line parameters that provide workflow inputs are
always optional.  This makes it possible to pass the executor an
incomplete input object.  This is desirable if the executor is going
to do something other than run the workflow.


tetron:
```
typeshed/2and3/schema_salad/ref_resolver.py:217: error: Argument 1 to "urlsplit" has incompatible type "Optional[str]"; expected "str"

typeshed/2and3/schema_salad/ref_resolver.py:298: error: Incompatible return value type (got "Optional[str]", expected "str")

typeshed/2and3/schema_salad/ref_resolver.py:298: error: Value of type variable "AnyStr" of "urljoin" cannot be "Optional[str]"

typeshed/2and3/schema_salad/schema.py:553: error: Value of type variable "Avro" of "make_valid_avro" cannot be "Union[Any, Dict[str, str], str]"

cwltool/process.py:188: error: Argument "cache" to "load_schema" has incompatible type "Dict[str, Union[bytes, str]]"; expected "Optional[Dict[str, str]]"

cwltool/process.py:191: error: Argument "cache" to "load_schema" has incompatible type "Dict[str, Union[bytes, str]]"; expected "Optional[Dict[str, str]]"

cwltool/command_line_tool.py:654: error: Argument 3 to "SourceLine" has incompatible type "Callable[[Any], Any]"; expected "Union[Type[str], Type[Exception]]"

Makefile:186: recipe for target 'mypy3' failed
```

This is unrelated to anything on this branch, some dependency seems to have changed to make mypy unhappy.  Strangely, even though it is complaining about schema salad, if you run mypy in the schema salad repo itself, it doesn't produce any errors.

PR #1192, opened 2019-09-05
Provide cwltool_latest for badges

tom-tan:
It just copy `cwltool_$version` to `cwltool_latest` generated in Jenkins CI because symlink does not work in Github.
It is not the best solution but it is enough as a workaround.

### DIFF
--- a/jenkins.bash
+++ b/jenkins.bash
@@ -114,6 +114,7 @@ EOM
 
 	if [ -d conformance ]
 	then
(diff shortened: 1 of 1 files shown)


jenkins.bash
```diff
@@ -114,6 +114,7 @@ EOM
 
 	if [ -d conformance ]
 	then
+		cp -r conformance/cwltool/cwl_${version}/cwltool_${tool_ver} conformance/cwltool/cwl_${version}/cwltool_latest
```
mr-c: Does this work if the directory already exists?

tom-tan: No.
I fixed to remove `cwltool_latest` before copying the current conformance result.

Issue #1204, opened 2019-10-04
id in CommandLineTool causes dot graph to not be printed correctly


itajaja:
`cwltool --print-dot workflow.cwl` produces an empty output if `CommandLineTool` has an `id` field. I was not able to identify a further root cause, but If the `id` field is removed, than `--print-dot` works correctly as expected.

## Expected Behavior
```
digraph {compound=true
"workflow_with_id.cwl#my_step" [label="my_step"]
}%
```

this is the expected graph. This is what gets correctly printed if I remove the `id` root field from the my_step.cwl file

## Actual Behavior
```
digraph {compound=true
}%
```

## Workflow Code
workflow:

```yml
class: Workflow
cwlVersion: v1.0
requirements:
- class: StepInputExpressionRequirement
- class: InlineJavascriptRequirement
- class: MultipleInputFeatureRequirement

inputs:
  foo: string

outputs:
  bar:
    type: string
    outputSource: my_step/bar

steps:
  my_step:
    run: ./my_step.cwl
    in:
      foo:
        source: foo
    out: [bar]
```

step:

```yml
class: CommandLineTool
cwlVersion: v1.0
id: my_step
baseCommand:
  - echo
inputs:
  foo: string
outputs:
  bar:
    type: string

```

## Your Environment
`1.0.20190915164430`

itajaja:
cc @meyermj

itajaja:
if I do `--print-rdf` the two files look exactly the same except for the following line:

```
<file:///Users/gtagliabue/workspace/qsi/consular/tests/test_data/workflow_samples/miniasm.cwl#my_step> a cwl:CommandLineTool ;
```

vs

```
<file:///Users/gtagliabue/workspace/qsi/consular/tests/test_data/workflow_samples/miniasm.cwl> a cwl:CommandLineTool ;
```

and the fact that the order for the workflow and steps file is inverted

itajaja:
ping 👋

mr-c:
Perhaps we should be stripping these `id`s out, as they then don't match the generated IDs for the other components


```
ipdb> print(ruamel.yaml.round_trip_dump(wf.steps[0].embedded_tool.tool))                                                                                                                                    
class: CommandLineTool
id: file:///home/michael/cwltool/1204-step.cwl#my_step
baseCommand:
- echo
inputs:
- type: string
  id: file:///home/michael/cwltool/1204-step.cwl#foo
outputs:
- type: string

[...]

PR #1210, opened 2019-10-15
overwrite manually set ids

mr-c:
Solves #1204 but there may be a better way


stain:
I am not sure I understand this patch.. so if `name` is set on the top-level workflow loaded from a `*.cwl` file, then that becomes its new `id`? (presumably instead of `#main`?)

mr-c:
@stain `name` is a reserved field in `avro`, it gets set during parsing by schema-salad. In the circumstances of this bug it happens to contain the normal computed ID that is being overwritten by the user provided `id` and breaking the graph.

Like I said, this isn't a great fix, but it works

mr-c:
> top-level workflow

No, `workflowobj` is the contents of a workflow step `run` field after loading the URI reference.

mr-c:
Jenkins, test this please

PR #1219, opened 2019-11-22
Output destination

tetron:
Experimental feature to set output destination of output files and directories.  Simplifies file renaming and organizing output into a desired directory structure.

Work in progress.

Example:

```
outputs:
  bar:
    type: File
    outputBinding:
      glob: foo
hints:
  cwltool:OutputDestination:
    destinations:
      bar: bar/
```

This causes the "foo" file to be placed in the subdirectory "bar" (relative to the base output directory) in the final output.

### DIFF
--- a/cwltool/executors.py
+++ b/cwltool/executors.py
@@ -22,7 +22,7 @@ from .errors import WorkflowException
 from .job import JobBase  # pylint: disable=unused-import
 from .loghandler import _logger
 from .mutation import MutationManager
-from .process import Process  # pylint: disable=unused-import
+from .process import Process, shortname
 from .process import cleanIntermediate, relocateOutputs
 from .provenance import ProvenanceProfile
 from .utils import DEFAULT_TMP_PREFIX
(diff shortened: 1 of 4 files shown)


mr-c:
Is this destination for downstream consumers or only final outputs?

If they later, then it would be better scoped as only being allowed in Workflow outputs (for 1.2) or as a requirement in Workflow proceses with no parent (as an extension).

I see the user value,  but this feels like it could get messy/confusing as per your example. How to deal with conflicts, etc..

tetron:
> Is this destination for downstream consumers or only final outputs?

In the current implementation, it only applies to final outputs.  However, the behavior allows for renaming files and directories as well as the more obvious use of placing into subdirectories, which seems like it could be useful mid-workflow.

> I see the user value, but this feels like it could get messy/confusing as per your example. How to deal with conflicts, etc..

The primary motivation is that if you want to produce output with specific file names and/or directory structure, currently you need to use a Javascript hack like this:

https://github.com/common-workflow-language/cwl-website/blob/master/site/mergesecondary.cwl

With this feature, you directly output parameters with destination directories.  You might still need to propagate a list of destination directories, but comparatively it is much easier to for the end user.

A secondary motivation is that because "destination" gives you a way to specify output filenames that are static (or trivially derived from input) it seems like it could support reasoning about the behavior of a workflow -- thinking of rule-based systems like `make` where you backtrack from a desired target.  (I'm afraid to even mention this, I don't want to sidetrack discussion, the first use case is much more concrete and addresses a very clear need).

tetron:
Discussion points:

* What does destination do when specified within a workflow?  Possibilities:
  * ignore
  * disallow
  * apply renames, but not directories
  * associate destination with file object or URI, propagate destination to end
* Have single `destination` which is parsed to separate dirname + basename, or two values `destDir`, `destName` ?
* File name conflicts -- probably needs to be an error if there are conflicting files with the same destination, can rename conflicting output files that don't have an explicit destination

PR #1244, opened 2020-01-28
Use non-flat badges with the commonwl icon for conformance badges

tom-tan:
It contains two changes to the conformance badges:

- Use non-flat badges as same as the other type of badges
  - Before: before
  - After: after

- Add the commonwl icon to the conformance badges
  - Before: before
  - After: after


mr-c:
Thanks!

PR #1251, opened 2020-02-11
Change from --volume to --mount

### LINKED ISSUE #1250: Error if colon in input/output paths when running with DockerRequirement
## Expected Behavior
When run with `DockerRequirement`, input paths with colons should work.

## Actual Behavior
It fails: 

```
INFO [job workflow.cwl] /tmp/4cf898fg$ docker \
   run \
   -i \
   --volume=/tmp/4cf898fg:/HZfoyx:rw \
   --volume=/tmp/e8pu8i7i:/tmp:rw \
   --volume=/home/airflow/repositories/ingest-pipeline/src/ingest-pipeline/airflow/data/temp/trig_
_2020-02-08T22:34:44.290910+00:00/cwl_out/cluster-marker-genes:/var/lib/cwl/stg706d994b-d3e7-49fd-9
9a2-d236a8783326/cluster-marker-genes:ro \
   --workdir=/HZfoyx \
   --read-only=true \
   --user=1008:1008 \
...
docker: Error response from daemon: invalid volume specification: '/home/airflow/repositories/inges
t-pipeline/src/ingest-pipeline/airflow/data/temp/trig__2020-02-08T22:34:44.290910+00:00/cwl_out/clu
ster-marker-genes:/var/lib/cwl/stg706d994b-d3e7-49fd-99a2-d236a8783326/cluster-marker-genes:ro'.
```

(from https://github.com/hubmapconsortium/airflow-dev/issues/33)

## Root Cause
in `cwltool/docker.py`:
```
    def append_volume(runtime, source, target, writable=False):
        # type: (List[str], str, str, bool) -> None
        """Add binding arguments to the runtime list."""
        runtime.append(
            "--volume={}:{}:{}".format(
                docker_windows_path_adjust(source), target, "rw" if writable else "ro"
            )
        )
```
With the `--volume`, colons have a special meaning. The **[solution](https://github.com/moby/moby/issues/8604#issuecomment-332673783)** is to use `--mount` instead.

## Next Steps
I'll make a PR to address this.

## Workflow Code
(skip for now)
## Full Traceback
(skip for now)
## Your Environment
(skip for now)

mccalluc:
This will fix #1250. Paths with colons can not be specified with the `--volume` syntax. They can, using `--mount`... and the parameter can also use CSV quoting in case there is a comma in a path.

- Would you like a test that actually runs docker, with a mounted path with weird characters?
- `--volume` will create the directory if it does not exist. `--mount` will not. [docker docs](https://docs.docker.com/storage/bind-mounts/#differences-between--v-and---mount-behavior). I could add an `os.mkdir()` to handle this?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+# Generated during tests
+pytestdebug.log
+tmp/
+
 # Python temps
 __pycache__/
 *.py[cod]
--- a/cwltool/docker.py
+++ b/cwltool/docker.py
@@ -237,8 +237,8 @@ class DockerCommandLineJob(ContainerCommandLineJob):
             "--mount={}".format(mount_arg)
         )
         # Unlike "--volume", "--mount" will fail if the volume doesn't already exist.
-        if not os.path.isdir(target):
-            os.mkdir(target)
+        if not os.path.isdir(source):
+            os.mkdir(source)
 
     def add_file_or_directory_volume(
         self, runtime: List[str], volume: MapperEnt, host_outdir_tgt: Optional[str]
--- a/cwltool/docker.py
+++ b/cwltool/docker.py
@@ -1,6 +1,7 @@
 """Enables Docker software containers via the {dx-,u,}docker runtimes."""
 
 import datetime
+import csv
 import os
 import re
 import shutil
@@ -8,7 +9,7 @@ import sys
 import tempfile
 import threading
 from distutils import spawn
-from io import open  # pylint: disable=redefined-builtin
+from io import open, StringIO  # pylint: disable=redefined-builtin
 from typing import Dict, List, MutableMapping, Optional, Set, Tuple
 
 import requests
@@ -222,10 +223,18 @@ class DockerCommandLineJob(ContainerCommandLineJob):
     def append_volume(runtime, source, target, writable=False):
         # type: (List[str], str, str, bool) -> None
         """Add binding arguments to the runtime list."""
+        options = [
+            'type=bind',
+            'source=' + source,
+            'target=' + target,
+        ]
+        if not writable:
+            options.append('readonly')
+        output = StringIO()
+        csv.writer(output).writerow(options)
+        mount_arg = output.getvalue().strip()
         runtime.append(
-            "--volume={}:{}:{}".format(
-                docker_windows_path_adjust(source), target, "rw" if writable else "ro"
-            )
+            "--mount={}".format(mount_arg)
         )
 
     def add_file_or_directory_volume(
--- a/cwltool/docker.py
+++ b/cwltool/docker.py
@@ -236,6 +236,9 @@ class DockerCommandLineJob(ContainerCommandLineJob):
         runtime.append(
             "--mount={}".format(mount_arg)
         )
+        # Unlike "--volume", "--mount" will fail if the volume doesn't already exist.
+        if not os.path.isdir(target):
+            os.mkdir(target)
 
     def add_file_or_directory_volume(
         self, runtime: List[str], volume: MapperEnt, host_outdir_tgt: Optional[str]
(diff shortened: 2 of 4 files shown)


mr-c:
Thank you @mccalluc for your contribution!

From what version of Docker does it support the `--mount` option?

mccalluc:
I think it came in [17.06.1-ee-1](https://docs.docker.com/engine/release-notes/#17061-ee-1):
> Add --mount flag to docker run and docker create [#32251](https://github.com/moby/moby/pull/32251)

mccalluc:
@mr-c: I'm going to work on improving the test coverage... but if you don't think this PR will eventually be accepted, please let me know so we can think about other ways to address this. Thanks!

mr-c:
> I think it came in 17.06.1-ee-1

Okay, I think that is old enough to change the behaviour. If not, we can add a fallback.

mr-c:
`tests/test_iwdr.py::test_iwdr_permutations ` fails with this PR

```
INFO Resolved '/home/michael/cwltool/tests/wf/iwdr_permutations.cwl' to 'file:///home/michael/cwltool/tests/wf/iwdr_permutations.cwl'
INFO [job iwdr_permutations.cwl] /tmp/fnrm3pac$ docker \
    run \
    -i \
    --mount=type=bind,source=/tmp/fnrm3pac,target=/ghJbaK \
    --mount=type=bind,source=/tmp/0pd5jtz0,target=/tmp \
    --mount=type=bind,source=/tmp/cooih7wq/tmplrqq5nd6,target=/my_path/seventh_writable_directory \
    --mount=type=bind,source=/home/michael/cwltool/tmpe6a78y3m/tmp78m4epo3,target=/my_other_path/fourth_read_only_file,readonly \
    --mount=type=bind,source=/tmp/0iczbg07/tmpg5kvl7bi,target=/my_path/third_writable_file \
    --mount=type=bind,source=/home/michael/cwltool/tmpp_2yhsbu,target=/my_other_path/eighth_read_only_directory,readonly \
    --mount=type=bind,source=/tmp/75y3ci04/tenth_writable_directory_literal,target=/my_path/tenth_writable_directory_literal \
    --mount=type=bind,source=/home/michael/cwltool/tmpe6a78y3m/tmp_priww6b,target=/ghJbaK/second_read_only_file,readonly \
    --mount=type=bind,source=/home/michael/cwltool/tmpuzji88vn,target=/ghJbaK/sixth_read_only_directory,readonly \
    --mount=type=bind,source=/tmp/2643zgzi/my_file_literal,target=/my_path/my_file_literal,readonly \
    --workdir=/ghJbaK \
    --read-only=true \
    --user=1000:1000 \
    --rm \
    --env=TMPDIR=/tmp \
    --env=HOME=/ghJbaK \
    --cidfile=/tmp/rkq9y_0f/20200214112856-439582.cid \
    debian \
    bash \
    -c \
    'find .
find /my_path
find /my_other_path
echo "a" > first_writable_file
echo "b" > /my_path/third_writable_file
touch fifth_writable_directory/c
touch /my_path/seventh_writable_directory/d
'
docker: Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /tmp/75y3ci04/tenth_writable_directory_literal.
```

So, yes, please add the `os.mkdir` calls, as needed

mccalluc:
I'm not sure about the best place to add the mkdir:
- (a) Could add it in `append_volume`... but that would be a weird side-effect in a method which is otherwise just modifying the `runtime` field,
- (b) ... or I can add it outside `append_volume`... but since it gets called from many different points, that would get hairy.

I'll start on (a), and then come back to this point, make another branch, and try (b).

mccalluc:
@mr-c : For the alternate approach, see #1252.

mccalluc:
Tests still fail. The unit test run in isolation passes, but I've just started the full tox locally now, and will hopefully see the same error as on Travis. Sorry for the noise.

mccalluc:
@mr-c: Tests pass now. Merge if this looks good, or let me know if there are other changes you'd like to see, or if you'd prefer the mkdir to be at a higher level, like I had started in the other PR.

mr-c:
Thanks @mccalluc !

PR #1252, opened 2020-02-14
Change from --volume to --mount, avoid side-effects

mccalluc:
This is an alternative to #1251: Burying an os.mkdir call inside a method that before was just updating a field seemed like it could be an unexpected side-effect.


mr-c:
Closed in favor of #1251

PR #1253, opened 2020-02-21
RFC: first draft of running tools via MPI

rupertnash:
Dear cwltool maintainers,

This PR is not really intended for immediate acceptance, I'm really just soliciting comment on if this is something you might possibly accept and if so, is this the sort of approach you'd be happy with.

Use case: as part of a workflow, I need to run tools that are parallelised with MPI. These must be launched with an implementation-defined tool. E.g. the commonest is `mpirun`, but Cray machines use `aprun` and the SGI (now HP) use `mpt_mpiexec`. Hence I don't want to put them in the workflow definition.

My patch adds command line flags to cwl-runner which will prepend the appropriate `mpirun -n tasks` to the command executed.

Feedback gratefully received!

### DIFF
--- a/cwltool/argparser.py
+++ b/cwltool/argparser.py
@@ -608,6 +608,38 @@ def arg_parser() -> argparse.ArgumentParser:
         "listed targets (can provide more than once).",
     )
 
+    mpigroup = parser.add_argument_group("Options for controlling parallel execution")
+    mpigroup.add_argument(
+        "--mpi",
+        action="store_true",
+        dest="mpi_on",
+        help="Run with MPI"
+    )
+    mpigroup.add_argument(
+        "--mpi-run",
+        type=str,
+        default=None,
+        help="Command to launch MPI jobs"
+    )
+    mpigroup.add_argument(
+        "--mpi-tasks-flag",
+        type=str,
+        default=None,
+        help="Flag to specify number of MPI tasks"
+    )
+    mpigroup.add_argument(
+        "--mpi-tasks",
+        type=int,
+        default=1,
+        help="Number of MPI tasks to use"
+    )
+    mpigroup.add_argument(
+        "--mpi-extra",
+        action="append",
+        default=[],
+        help="Extra argument to the MPI command"
+    )
+
     parser.add_argument(
         "workflow",
         type=str,
--- a/cwltool/command_line_tool.py
+++ b/cwltool/command_line_tool.py
@@ -40,6 +40,7 @@ from schema_salad.utils import json_dumps
 from .builder import Builder, content_limit_respected_read_bytes, substitute
 from .context import LoadingContext, RuntimeContext, getdefault
 from .docker import DockerCommandLineJob
+from .mpi import MpiCommandLineJob
 from .errors import WorkflowException
 from .flatten import flatten
 from .job import CommandLineJob, JobBase
@@ -354,6 +355,8 @@ class CommandLineTool(Process):
                     "--no-container, but this CommandLineTool has "
                     "DockerRequirement under 'requirements'."
                 )
+        if runtimeContext.mpi_on:
+            return MpiCommandLineJob
         return CommandLineJob
 
     def make_path_mapper(self, reffiles, stagedir, runtimeContext, separateDirs):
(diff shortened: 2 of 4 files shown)


tetron:
Hi Rupert,

The recommended way of doing this is to create your own runner by wrapping cwltool and providing customized classes.

Three things to look at to get started:

JobBase

Derive a new class from this (eg MPIJob) that actually invokes a parallel task.

CommandLineTool.make_job_runner

Create a custom class (eg MPICommandLineTool) that overrides this method and returns your custom MPIJob object.

LoadingContext.construct_tool_object

Assign your own constructor method which returns MPICommandLineTool objects.

You "main" method can invoke the cwltool main method with your LoadingContext instance with construct_tool_object set.

mr-c:
@rupertnash are these tools runnable without MPI?

If not, then there should also be a new `MPIRequirement` or similar.

This would be a good topic for the next CWL video chat which is on [2020-03-24 14:30 UTC ](https://www.timeanddate.com/worldclock/fixedtime.html?msg=CWL+video+chat&iso=20200324T1430) via https://meet.jit.si/cwl

rupertnash:
Hi Michael- Yes and no... 

Some of the tools in my workflow only make sense to be run with MPI (in my example a high resolution weather simulation of a small area)

Others (here the preprocessing steps to extract the subdomain from a global forecast) are run in serial and could often just be run from the command line as `executable_name param.file`, however due to "quirks" in the way the package is written (the library-like parts call `MPI_Init`) combined with the architecture of the supercomputer I'm using (any code using MPI must run only on compute nodes) mean they have to be launched as MPI programs using a single task.

My requirement is that I am able to factor out the MPI parameters from the CWL file. I want to use the same CWL spec on different machines with different size jobs, e.g. run a v small 32 core job on our test cluster: `mpirun -n 32 weather testjob.params` and production scale jobs on e.g. ARCHER (UK national machine) `aprun -n 768 -N 24 weather bigjob.params`

Obviously I will have to supply this information somehow and I've changed my approach somewhat from the one I used in this PR. I've figured out how to compile NodeJS and used inline JS to do something sensible. I've almost got my workflow going - I'll get it on GH later today and drop a link in here. (Basically I have an MPI input which holds non-default params and constructs the command line for me: 
```
arguments:
  - position: 0
    valueFrom: $(mpi.run("PREP_REAL_CASE"))
```
)

I'd be very happy to join your call tomorrow, but I will have to leave at 15:00 UTC to join another one. If it would be OK to put this item near the top of the agenda that would be

mr-c:
@rupertnash Sure, I put "MPI" as the first thing on the agenda for tomorrow: https://docs.google.com/document/d/1DnW9R3GjrLStVxJtYP51S7-VNM8cZRHeoi-XaeZ9kNY/edit?usp=sharing

Talk to you then!

rupertnash:
Due to scheduling I will have to join you next week at https://www.timeanddate.com/worldclock/fixedtime.html?&iso=20200331T1500

Since there is a bit more time, is there anything that would be useful for me to flesh out in advance?

mr-c:
No worries, @rupertnash ; looking forward to the conversation next week

>  is there anything that would be useful for me to flesh out in advance?

1. What does this mean for portability, especially for non-HPC execution?

2. Does it need its own `MPIRequirement` extension?

mcpherson:
Can I briefly hijack this discussion? Please slap me down if not, but I think my questions are related (at least to @tetron entry above).

I believe @mr-c is familiar with what we're trying to do at Los Alamos: use CWL to specify HPC (including lots of MPI) workflows, load a graph database with the parsed workflow, use our own *runner* to execute the workflow on HPC platforms with something like Slurm (to start).

We're finding that parsing the CWL file to a graph is very hard and will consume lots of resources. We are back to searching for an alternative. The @tetron entry has me thinking of `cwltool` again. Perhaps we could *wrap* `cwltool` to load our database instead of actually executing the job. This is just a thought, but perhaps you can point me to the best venue for discussions with CWL/`cwltool` experts. Maybe one of these teleconferences you mention?

We also want to extend CWL to use Python for scripting (vs. JavaScript), but that's for another day.

Thanks for any consideration.

tetron:
Hi @mcpherson 

The best places are the gitter channel and the weekly video chat.

Here are notes from the discussion with @rupertnash about MPI support in CWL.

https://docs.google.com/document/d/1StJS0f5KHjKWShki3aUbnpCsUPiPZEaYrvArVJIq5HA/edit?usp=drivesdk

As far as what you are doing at Los Alamos HPC, I would need to learn more about the particular infrastructure problems you are trying to solve before I can make any suggestions.

rupertnash:
Closing - follow up in #1276

PR #1258, opened 2020-03-05
WIP: Allow fractional cores

bogdang989:
Initial implementation for common-workflow-language/common-workflow-language#883


tetron:
This is incorporated in #1262

PR #1263, opened 2020-03-17
Pass loadingContext to pack().

tetron:
Ensures that all the document loading options get passed through.


tetron:
It would be fair to say I am hoist by my own petard of cwltool document loading complexity.  But supporting different CWL versions is also really hard!

Issue #221, opened 2016-11-08
enable use inside Jupyter notebooks


mr-c:
Tricky due to Jupyter's IO not supporting fileno/subprocess:
``` python
import cwltool.factory
f = cwltool.factory.Factory()
echo = f.make("v1.0/v1.0/echo-tool.cwl")
out = echo(in="foo")
```
```
  File "/home/michael/src/jupyterenv/local/lib/python2.7/site-packages/cwltool/job.py", line 221, in run
    cwd=self.outdir)
  File "/usr/lib/python2.7/subprocess.py", line 703, in __init__
    errread, errwrite), to_close = self._get_handles(stdin, stdout, stderr)
  File "/usr/lib/python2.7/subprocess.py", line 1153, in _get_handles
    errwrite = stderr.fileno()
  File "/home/michael/src/jupyterenv/local/lib/python2.7/site-packages/ipykernel/iostream.py", line 307, in fileno
    raise UnsupportedOperation("IOStream has no fileno.")
UnsupportedOperation: IOStream has no fileno.
```

r4space:
I see this and the linked issue are both closed - was the feature added or the effort disbanded?

(googling Jupyter and CWL pretty much only returns these)

PR #1272, opened 2020-04-15
enable use inside Jupyter notebooks

giannisdoukas:
I am creating a kernel for the JN. Because I had that [issue](https://github.com/common-workflow-language/cwltool/issues/221) I add the option on the RuntimeContext to replace the default stdin/stdout/stderr.


cwltool/job.py
```diff
@@ -903,19 +906,25 @@ def _job_popen(
...
     monitor_function=None,  # type: Optional[Callable[[subprocess.Popen[str]], None]]
+    default_stdin: Optional[Union[IO[Any], int]] = None,
+    default_stdout: Optional[IO[Any]] = None,
+    default_stderr: Optional[IO[Any]] = None,
 ) -> int:
 
     if job_script_contents is None and not FORCE_SHELLED_POPEN:
 
-        stdin = subprocess.PIPE  # type: Union[IO[Any], int]
+        stdin = default_stdin if default_stdin is not None else subprocess.PIPE # type: Union[IO[Any], int]
```
tetron, line 916: I don't think this makes sense, users are not supposed to ever interact directly with running programs

giannisdoukas: I did it to not have [issue](https://github.com/common-workflow-language/cwltool/issues/221) with the process. To be able to use the runtimeContext like this:

runtime_context.default_stdin = subprocess.DEVNULL
runtime_context.default_stdout = subprocess.DEVNULL
runtime_context.default_stderr = subprocess.DEVNULL

https://github.com/giannisdoukas/CWLJNIKernel/blob/master/cwlkernel/CoreExecutor.py#L47

tetron: Maybe I'm misunderstanding the problem, but what the original code is doing is that stdin for the subprocess is always a pipe, it never reads from sys.stdin.  You don't want it to read from sys.stdin either, and the existing code already does that.  I'm suggesting that `default_stdin` is unnecessary -- `default_stdout` and `default_stderr` are fine.

giannisdoukas: Yes, it was my fault because I did not include any test to illustrate how I want to use that feature. To be honest I wasn't sure if I had to do it for the stdin but I wanted for sure for the stdout/stderr.

tetron: So I think you want to revert the stdin changes, and add a test case for capturing stdout/stderr.

mr-c:
Thank you @giannisdoukas !

PR #1276, opened 2020-04-27
Add MPIRequirement

rupertnash:
Following discussion in #1253 and the telco, I have implemented support for tools that need to be launched in parallel with MPI.

I have split the configuration required into 2 parts. An environment configuration (how to run MPI programs on this particular machine) which is given the cwltool via a simple YAML file by the `--mpi-config-file` command line option. The second part, the number of processes to start, is configured per-tool, eg. in the CWL file:
```
requirements:
  MPIRequirement:
    processes: $(inputs.num_processes)
```

As with my previous PR, this is really just starting the conversation. I hope to be able to spend a bit more time writing code in the next week or two. I would also be pleased to join the CWL call if this is happening on Tuesday 28th?

- [x] Figure out whether we need to pass through environment vars to mpirun 
 - Yes, we do, in at least one case I've identified
- [x] If so, add code to specify these to platform configuration


cwltool/process.py
```diff
@@ -107,6 +107,7 @@ def filter(self, record):  # type: (logging.LogRecord) -> bool
     "NetworkAccess",
     "InplaceUpdateRequirement",
     "LoadListingRequirement",
+    "MPIRequirement",
```
mr-c, line 110: Should be prefixed like those below
```suggestion
    "http://commonwl.org/cwltool#MPIRequirement",
```

tetron:
How does this interact with HardwareRequirement?  Since you are calling the parallel scheduler you probably also need to request CPUs and RAM?

I had a thought about generalizing the construction of the MPI call.  What if it was a CommandLineTool with a defined interface?  

Something like:

```
class: CommandLineTool
inputs:
  numprocs: int
  ram: int
  cores: int
  workdir: Directory,
  command: string[]
baseCommand: my_mpi_runner
arguments: [
  --numprocs, $(inputs.numprocs),
  --ram, $(inputs.ram),
  --cores, $(inputs.cores),
  --workdir, $(inputs.workdir),
  $(inputs.command)
]
```

So cwltool would build the command line of the underlying command, and then pass *that* with all the runtime information to the MPI runner tool to construct the MPI command line.

rupertnash:
Hi Peter,


> How does this interact with HardwareRequirement? Since you are calling the parallel scheduler you probably also need to request CPUs and RAM?
We are not, here, calling the parallel scheduler (I have not seen that term in CWL? Are you perhaps mixing your paradigms?) nor specifying the number of cores to use for this tool invocation, just the number of MPI processes to start (these may often have the same value, but one often has a hybrid parallel code with say one MPI process per NUMA region and use a threading approach (often OpenMP) between those cores). That’s not to say that you couldn’t use e.g. $(requirements.HardwareRequirement.cores) as the value of the MPIRequirement.processes!

So I would suggest that this requirement is in the general case orthogonal to the HardwareRequirement- those have to be satisfied at an earlier stage (allocating hardware) than here (last thing before invocation).
> I had a thought about generalizing the construction of the MPI call. What if it was a CommandLineTool with a defined interface?
> 
> Something like:
> 
> class: CommandLineTool
> inputs:
>   numprocs: int
>   ram: int
>   cores: int
>   workdir: Directory,
>   command: string[]
> baseCommand: my_mpi_runner
> arguments: [
>   --numprocs, $(inputs.numprocs),
>   --ram, $(inputs.ram),
>   --cores, $(inputs.cores),
>   --workdir, $(inputs.workdir),
>   $(inputs.command)
> ]
> So cwltool would build the command line of the underlying command, and then pass that with all the runtime information to the MPI runner tool to construct the MPI command line.
> 
This is a really interesting idea! So instead of using my simple but ad hoc YAML file to specify how to do “mpirun(processes, *args)” you provide a CWL which can both formalise the interface AND be more flexible behind it.

This might also work for other runtimes (containers even?) 

I’m not really sure how to implement it however?

Thanks!
Rupert

tetron:
I see.  I was thinking that it would submit your reservation and then block until you get it, and then your program would launch.  I remember now from the discussion we had a few weeks ago, the assumption here is that the nodes have already been reserved, so you only need the special MPI launcher.

Generalizing the definition of the launcher would be powerful but not necessary to merge a first pass at this feature.

I'll have to take a closer look at the code a little later but in the mean time your mpi unit tests are failing on the CI infrastructure.

cwltool/process.py
rupertnash: Hi Michael - I have tried this and ended up going down a bit of a rabbit hole of confusion about the Schema Salad definitions... I think I've found a bug in SALAD that I'll PR over on that repo (will require a small patch in cwltool also).

Then I noticed that all the requirements with a `http://commonwl.org/cwltool#` prefix also appear with it. Is this what you mean:
```
supportedProcessRequirements = [
    # snip...
    "LoadListingRequirement",
    "MPIRequirement",
    "http://commonwl.org/cwltool#MPIRequirement",
    # snip
    "http://commonwl.org/cwltool#LoadListingRequirement",
    "http://commonwl.org/cwltool#InplaceUpdateRequirement",
]
```
?

mr-c: The prefixes are necessary for requirements that are not (or were not at the time) part of the CWL standard. For now, MPIRequirement must be prefixed only

@tetron can comment on if we should remove the other prefixed requirements that have since moved into a released version of the standard. Shouldn't the upgrader be internally rewriting these, yes?

rupertnash: Thanks for the steer - however that takes me back to schema issues... If I make your change, then the tool loading fails in `process.checkRequirements`, because as the tool definition is parsed as:
```
'$namespaces': ordereddict([('cwl', 'http://commonwl.org/cwltool#')]),
 'arguments': ['-c', 'import os; print(os.getpid())'],
 'baseCommand': 'python',
 'class': 'CommandLineTool',
 'cwlVersion': 'v1.2.0-dev3',
 'doc': 'Trivial MPI test that prints the process IDs of each of the parallel\n'
        'processes. Requires Python (but you have cwltool running, right?)\n'
        'and an MPI implementation.\n',
 'http://commonwl.org/cwltool#original_cwlVersion': 'v1.2.0-dev2',
 'id': 'file:///Users/rnash2/work/vestec/cwltool/tests/wf/mpi_simple.cwl',
 'inputs': [],
 'outputs': [ordereddict([('type', 'File'), ('id', 'file:///Users/rnash2/work/vestec/cwltool/tests/wf/mpi_simple.cwl#pids'), ('outputBinding', ordereddict([('glob', 'b32630f55cb7498ca7c8277cb00b49d1856a88dd')]))])],
 'requirements': [ordereddict([('processes', 2), ('class', 'MPIRequirement')])],
 'stdout': 'b32630f55cb7498ca7c8277cb00b49d1856a88dd'}
```

Note that in the .cwl, requirements is:
```
requirements:
  cwl:MPIRequirement:
    processes: 2
```
so the `cwl:` prefix/namespace is getting stripped off somehow. Changing my schema to have `inVocab: false` gives an enormous error during schema validation. At it's root the issue is that the `Names` object passed to the `UnionSchema` initialiser has a default namespace of "http://commonwl", which is correct as an Avro name but not helpful here. This motivates my ticket https://github.com/common-workflow-language/schema_salad/issues/325

rupertnash:
- [ ] Figure out whether we need to pass through environment vars to mpirun 
- [ ] If so, add code to specify these to platform configuration

README.rst
```diff
@@ -558,6 +558,70 @@ For this example, grab the test.json (and input file) from https://github.com/Ca
...
+|                |                  |          | environment. Those that match|
+|                |                  |          | will be passed through.      |
++----------------+------------------+----------+------------------------------+
+| env_set        | Mapping[str,str] | {}       | A dictionary whose keys are  |
+|                |                  |          | the environment variables set|
+|                |                  |          | and the values being the     |
+|                |                  |          | values.                      |
++----------------+------------------+----------+------------------------------+
+
+
```
mr-c, line 624: Say something about how MPIRequirement does or does not interact with `DockerRequirement`?

cwltool/argparser.py
```diff
@@ -608,6 +608,13 @@ def arg_parser() -> argparse.ArgumentParser:
         "listed targets (can provide more than once).",
     )
 
+    parser.add_argument(
+        "--mpi-config-file",
+        type=str,
+        default=None,
+        help="Platform specific configuration for MPI (parallel launcher, its flag etc).",
```
mr-c, line 615: Mention where this file format is defined?

cwltool/extensions-v1.1.yml
```diff
@@ -44,3 +44,23 @@ $graph:
         subscope: run
       doc: |
         Specifies the process to run.
+
+- name: MPIRequirement
```
mr-c, line 48: I think this extension could work against `v1.0` files as well ; is there a particular reason you chose v1.1 as the starting point?

rupertnash:
A collaborator has now confirmed that this works on an HPC system that needs a whole load of env vars set in order to launch jobs. He also requests improvements to docs...

mr-c:
> He also requests improvements to docs...

@rupertnash The price of success! Thanks again :-)

README.rst
rupertnash: Done. I've also mentioned Singularity, which does work for simple cases.

cwltool/extensions-v1.1.yml
rupertnash: Cos it's newer and I only wanted to write the thing once while it was under development? Added to 1.0 extensions also.

README.rst
```diff
@@ -582,6 +582,15 @@ in an integer). For example::
...
       type: int
 
+Interaction with containers: the MPIRequirement currently prepends its
+commands to the front of the command line that is constructed. If you
+wish to run a containerised application in parallel, for simple use
+cases this does work with Singularity, depending upon the platform
+setup. However this combination should be considered "alpha" -- please
+do report any issues you have! This does not work with Docker at the
+moment. (More precisely, you get `n` copies of the same single process
+image run at the same time that cannot communicate with each other.)
```
mr-c, line 592: Thanks for the information. If it doesn't work with Docker then we should not try to run if `DockerRequirement` is under `requirements` and we are not in `--singularity` mode. Likewise if `DockerRequirement` is under `hints` and we have neither `--no-containers` nor `--singularity`.

Did you test with udocker?

rupertnash: Due to an ongoing issue with our cluster, I've not been able to test with udocker, however documentation suggests that it *should* work with MPI. Someone else testing this might be a useful exercise (of this and the wider MPIRequirement feature).

I have added the necessary code (and some tests) to stop if one has both Docker and MPI in `requirements` (same for `hints` as no obvious tie-breaker). If one is hinted and the other required, then the hint is dropped (and warned)

PR #1281, opened 2020-05-01
Use schema_salad.avro.schema.Name according to docs

rupertnash:
Apropos my PR on the schema_salad repo (https://github.com/common-workflow-language/schema_salad/pull/324) make cwltool supply `Optional[str]` args as `None` when they are absent instead of `""`.

This change does NOT require a version bump of schema_salad in requirements.txt


rupertnash:
Thanks for updating this after merging the salad PR

PR #1301, opened 2020-06-02
cwltool will not try to convert to file:// uri, in case the URI already contains a schema

kmavrommatis:
`cwltool` converts all the file strings to a file:// URI
In cases cwltool is wrapped to support other types of URI schemas, e.g. s3:// it converts them to a meaningless URI of the form file://s3://....
The suggested changes leave file strings that already contain a schema unchanged.
Thanks in advance for the consideration.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,34 @@
+name: "Code scanning - action"
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 10 * * 2'
+
+jobs:
+  CodeQL-Build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        # We must fetch at least the immediate parents so that if this is
+        # a pull request then we can checkout the head.
+        fetch-depth: 2
+
+    # If this run was triggered by a pull request event, then checkout
+    # the head of the pull request instead of the merge commit.
+    - run: git checkout HEAD^2
+      if: ${{ github.event_name == 'pull_request' }}
+      
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      with:
+        languages: python
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v1


mr-c:
The tests pass on MS Windows but now the Linux tests fail :-D https://travis-ci.org/github/common-workflow-language/cwltool/jobs/698794202#L433

.project
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mr-c, line 1: This file doesn't belong in this PR

kmavrommatis: I will restore the files, 
thanks for catching this

kmavrommatis:
@mr-c .
I have resolved the previous coding issues, and travis and appveyor go through without errors but there are codecov failures.
tbh I am not codecov savvy and have not been able to find information about the fails in logs.
Can you provide some guidance or resources ?
Thanks

mr-c:
Hello @kmavrommatis. Take a look at https://codecov.io/gh/common-workflow-language/cwltool/compare/520acbfb82455c4bdabd5f2ea24842804e1c9f58...859b03c4080c49c0dd4656d58430668adbd8f4f7/diff

That shows lines of code changed in this PR that aren't executed by any of our tests. Can you add tests for your changes so we don't accidently re-introduce the problem later?

cwltool/process.py
```diff
@@ -407,13 +407,20 @@ def _realpath(
...
+
     outfiles = list(_collectDirEntries(outputObj))
     visit_class(outfiles, ("File", "Directory"), _realpath)
     pm = path_mapper(outfiles, "", destination_path, separateDirs=False)
     stage_files(pm, stage_func=_relocate, symlink=False, fix_conflicts=True)
 
     def _check_adjust(a_file: CWLObjectType) -> CWLObjectType:
-        a_file["location"] = file_uri(pm.mapper(cast(str, a_file["location"]))[1])
+        location = cast(str, a_file["location"])
+        _logger.debug("KM:: for file {} location is {}".format( a_file, location ))
```
mr-c, line 418: When you're ready for review, please remove this logging

kmavrommatis:
@mr-c 
> That shows lines of code changed in this PR that aren't executed by any of our tests. Can you add tests for your changes so we don't accidently re-introduce the problem later?

Thanks for the response, this is very helpful and makes sense.
The modifications of the code I introduced aim to allow the tool to work with files that are not on the file system (e.g. on s3, google storage etc). How would you suggest to create a test for this? wouldn't a simple cwl workflow that requires an s3 object fail if it cannot find it? 

Thanks

cwltool/process.py
kmavrommatis: The logging line is removed

mr-c:
@kmavrommatis Might be simpler to directly call the changed functions via new unit tests ; the existing tests (of all sizes: unit, integration, and end-to-end) are at https://github.com/common-workflow-language/cwltool/tree/master/tests

kmavrommatis:
Unable to resolve problems with CI tools. 
I will reopen a pull request later

PR #1302, opened 2020-06-04
change hardcoded path sep to os.sep

gungorbudak:
This is a super simple fix to the issue on Windows 10 for collecting secondary files on outputs I described here: https://cwl.discourse.group/t/secondary-files-cannot-be-collected-by-cwltool-on-windows-10/158


mr-c:
Thank you @gugorbudak! Our Microsoft Windows CI provider, appveyor, doesn't support Linux containers under MS Windows so these sort of bugs pop up as we can't automatically run the entire test suite there.

If anyone wants to run the `cwltool` test suite manually under MS Windows (not in WSL) and fix bugs, that would be very appreciated!

Also, if there is a CI provider that supports testing programs that launch Linux containers under MS Windows, I would very much like to hear about that!

PR #1310, opened 2020-07-01
Don't map file to local filesystem if the scheme is not file://

kmavrommatis:
The intention is to avoid mapping s3 or gs objects to the local
filesystem. Instead leave them at their cloud location.

See of https://github.com/common-workflow-language/cwltool/pull/1301 for a previous version of this PR


cwltool/process.py
```diff
@@ -415,7 +415,15 @@ def _realpath(
     stage_files(pm, stage_func=_relocate, symlink=False, fix_conflicts=True)
 
     def _check_adjust(a_file: CWLObjectType) -> CWLObjectType:
-        a_file["location"] = file_uri(pm.mapper(cast(str, a_file["location"]))[1])
+        location = cast(str, a_file["location"])
+        if urllib.parse.urlparse(os.path.splitdrive(a_file["location"])[1] ).scheme == "" :
```
mr-c, line 419:
```suggestion
        if urllib.parse.urlparse(os.path.splitdrive(cast(str, a_file["location"]))[1] ).scheme == "" :
```

kmavrommatis:
@mr-c 
As I was looking around the code to resolve some of the unit test failures, I am not sure that the place where the proposed change is the best to apply it. 
The whole reason behind the PR is to make sure that if the location of a file is on s3 or gs (cases that are handled e.g. by cwl-tes) the output files are not reported as URIs of the form : `file://s3://blah/blah`
This modification of the URI seems to be coming from the `schema.salad.ref_resolver.file_uri` function. Does it make sense to modify that instead of `cwltool.process` ? 
i.e. let that function return the existing URI (e.g. s3://.... , gs:// etc) if it detects one instead of altering it?
I don't know what this may affect downstream.
any thoughts about this?

Thanks for your help.

tetron:
I'm sorry I haven't been watching this too closely, but I think you might be doing this the wrong way.  The code that's turning your s3:// URLs into file paths is probably the `urljoin` method of `Fetcher`:

https://github.com/common-workflow-language/schema_salad/blob/main/schema_salad/fetcher.py#L15

You need to override the `supported_schemes` method, for example:

https://github.com/arvados/arvados/blob/master/sdk/cwl/arvados_cwl/fsaccess.py#L219

You can inject your customer fetcher by setting `fetcher_constructor` in `LoadingContext`.

mr-c:
@kmavrommatis looks like you are adopting the advice from @tetron in https://github.com/ohsu-comp-bio/cwl-tes/pull/41/commits/b7f3e323531b466e87be15b72980902abf93705e , so I'm going to close this PR for now

PR #1342, opened 2020-09-09
Fix algorithm for generating output names when output names clash

pvanheus:
When a scatter step in a workflow is chosen as a step in an output, the output names will naturally clash. Here is an example workflow:

```
#!/usr/bin/env cwl-runner
cwlVersion: v1.1
class: Workflow

inputs:
  range:
    type: int
outputs:
  output:
    type: File[]
    outputSource:
      generate_files/output

steps:
  generate_list:
    requirements:
    run:
      class: ExpressionTool
      inputs:
        max: 
          type: int
          default: 100
      outputs:
        numbers:
          type: int[]
      expression: |
        ${
          var numberList = Array.apply(null, Array(inputs.max)).map(function(_, i) { return i});
          return { "numbers": numberList } 
         }
    in:
      max: range
    out:
      - numbers
  generate_files:
    requirements:
      - class: ScatterFeatureRequirement
    scatter: number
    run:
      class: CommandLineTool
      inputs:
        number:
          type: int
          inputBinding:
            position: 10
      baseCommand: [ echo ]
      stdout: output.txt
      outputs:
        output:
          type: stdout
    in:
      number: generate_list/numbers
    out:
      - output
```

The current cwltool logic for this case is [here](https://github.com/common-workflow-language/cwltool/blob/main/cwltool/process.py#L281) and it results in names in a list like `output.txt`, `output.txt_2`, `output.txt_2_3` as the logic appends a number to the proposed target name instead of incrementing the number. Eventually cwltool crashes with a `File name too long` OSError exception. This patch changes the logic to always target the original target name (`entry.target`) and increment the number appended to the end so that names become `output.txt`, `output.txt_2`, `output.txt_3` etc. It also adds a little comment explaining the logic.

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -1261,3 +1261,23 @@ def test_v1_0_arg_empty_prefix_separate_false() -> None:
     )
     assert "completed success" in stderr
     assert error_code == 0
+
+
+def test_scatter_output_filenames(tmpdir: py.path.local) -> None:
+    """If a scatter step produces identically named output then confirm that the final output is renamed correctly."""
+    cwd = tmpdir.chdir()
+    rtc = RuntimeContext()
+    rtc.outdir = str(cwd)
+    factory = cwltool.factory.Factory(runtime_context=rtc)
+    output_names = ['output.txt', 'output.txt_2', 'output.txt_3']
+    scatter_workflow = factory.make(get_data("tests/scatter_numbers.cwl"))
+    result = scatter_workflow(range=3)
+    assert 'output' in result
+    
+    locations = sorted([element['location'] for element in result['output']])
+
+    assert(
+        locations[0].endswith('output.txt') and
+        locations[1].endswith('output.txt_2') and
+        locations[2].endswith('output.txt_3')
+    ), "Locations {} do not end with {}".format(locations, output_names)
--- a/cwltool/process.py
+++ b/cwltool/process.py
@@ -276,12 +276,13 @@ def stage_files(
             targets[entry.target] = entry
         elif targets[entry.target].resolved != entry.resolved:
             if fix_conflicts:
-                tgt = entry.target
+                # find first key that does not clash with an existing entry in targets
+                # start with entry.target + '_' + 2 and then keep incrementing the number till there is no clash
                 i = 2
-                tgt = "%s_%s" % (tgt, i)
+                tgt = "%s_%s" % (entry.target, i)
                 while tgt in targets:
                     i += 1
-                    tgt = "%s_%s" % (tgt, i)
+                    tgt = "%s_%s" % (entry.target, i)
                 targets[tgt] = pathmapper.update(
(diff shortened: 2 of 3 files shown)


pvanheus:
Some help would be appreciated interpreting the test failure - I don't see how it is related to the change that this PR introduces.

tetron:
@pvanheus that error is pretty obscure.  I have restarted the build in case it was a transient failure: https://travis-ci.org/github/common-workflow-language/cwltool/jobs/725533973

pvanheus:
Thanks @tetron  - hit the same error again. I really don't get what this error means.

tetron:
I am going to try running it locally and see what happens but I haven't gotten to it yet.

pvanheus:
@tetron I tried to run the tests locally as per the README but `py.test` complained about the `----pyarg` flag. Is there a way to perhaps do the testing in a Docker container?

mr-c:
Thank you @pvanheus for this! I think I fixed the build on our side.

Can you add your example as a test case? `range: 100` seems to run fast enough while still triggering the error without this fix

tests/test_examples.py
```diff
@@ -1261,3 +1261,30 @@ def test_v1_0_arg_empty_prefix_separate_false() -> None:
     )
     assert "completed success" in stderr
     assert error_code == 0
+
+
+def test_scatter_output_filenames(tmpdir: py.path.local) -> None:
+    """Check that when a scatter step produces same named output the final output is renamed correctly."""
```
mr-c, line 1267:
```suggestion
    """If a scatter step produces identically named output then confirm that the final output is renamed correctly."""
```
Small wording tweak; thanks for this test!

tests/test_examples.py
```diff
@@ -1261,3 +1261,30 @@ def test_v1_0_arg_empty_prefix_separate_false() -> None:
...
+        assert(
+            locations[0].endswith('output.txt') and
+            locations[1].endswith('output.txt_2') and
+            locations[2].endswith('output.txt_3'),
+            "Locations {} do not end with {}".format(locations, output_names)
+        )
+    finally:
+        for filename in output_names:
+            if os.path.exists(filename):
+                os.remove(filename)
```
mr-c, line 1290: I think pytest cleans up automatically if you use the `tmpdir` fixin?

https://docs.pytest.org/en/latest/tmpdir.html#the-tmpdir-fixture

pvanheus: Perhaps. Other tests do `tmpdir.remove(ignore_errors=True)` e.g. https://github.com/common-workflow-language/cwltool/blob/main/tests/test_examples.py#L952

mr-c: I think that was a work around for a bug that was [fixed](https://github.com/pytest-dev/pytest/pull/6944) in Pytest 6.0.0 ; I'll make a separate PR to remove those

mr-c (approved):

PR #1343, opened 2020-09-28
podman driver changes

nyue:
To use podman, the following changes are required.
However, these changes are not docker compatible
so there is a need for a maintainable way to support
the different flags/arguments required for different
drivers

### DIFF
--- a/cwltool/docker.py
+++ b/cwltool/docker.py
@@ -259,7 +259,7 @@ class DockerCommandLineJob(ContainerCommandLineJob):
             "target=" + target,
         ]
         if not writable:
-            options.append("readonly")
+            options.append("ro")
         output = StringIO()
         csv.writer(output).writerow(options)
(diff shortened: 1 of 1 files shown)


cwltool/docker.py
```diff
@@ -392,7 +392,7 @@ def create_runtime(
                 runtime.append("--net=none")
 
             if self.stdout is not None:
-                runtime.append("--log-driver=none")
+                runtime.append("--log-driver=k8s-file")
```
mr-c, line 395: I'd accept a flag and/or auto detection for this one

mr-c (commented):
Interesting!

nyue:
I have since tested podman 2.x with cwltool and found that they have made podman command line options more compatible with docker. Hence, this pull request is redundant.

mr-c:
@nyue Good news, thanks!

PR #1350, opened 2020-10-12
Provide Docker for CWL workflows that use Docker in Kubernetes batch system

jonathanxu18:


jonathanxu18:
@mr-c I edited the error message when checking for --userns support in Singularity since I think it was checking for the wrong one. I also took out --pid when using --userns since it would try to mount /proc which was preventing CWL workflows that use Docker from running in Kubernetes. Please let me know what you think!

cwltool/singularity.py
```diff
@@ -47,7 +47,7 @@ def _singularity_supports_userns() -> bool:
                 stdout=DEVNULL,
                 universal_newlines=True,
             ).communicate(timeout=60)[1]
-            _USERNS = "No valid /bin/sh" in result
+            _USERNS = "/bin/sh doesn't exist in container" in result
```
mr-c, line 50: Can we get a test for this?

mr-c: Thanks for the test. In which version of Singularity did the error message change?

We should support both error messages.

mr-c (changes requested):
There is a `.DS_Store` file in this PR that shouldn't be here. Also, we should continue to support the old error message as well as the new one.

Thanks again for this!

cwltool/singularity.py
```diff
@@ -47,7 +47,7 @@ def _singularity_supports_userns() -> bool:
                 stdout=DEVNULL,
                 universal_newlines=True,
             ).communicate(timeout=60)[1]
-            _USERNS = "No valid /bin/sh" in result
+            _USERNS = "No valid /bin/sh" in result or "/bin/sh doesn't exist in container" in result
```
jonathanxu18, line 50: @mr-c I was having some trouble figuring out which version the error message changed but I tweaked the check to support both error messages. Would this be sufficient or would you still like to know which version it changed on. Thank you!

jonathanxu18: Also, I removed the `.DS_Store` from the previous commit.

cwltool/singularity.py
jonathanxu18: @mr-c I see that I'm currently failing the Travis build test because of permission denials for the test I added. Do I need to edit this? I tested it on my local machine and it seemed to be working fine.

mr-c (approved):

mr-c:
@jonathanxu18 I think the tests are failing due to something I'm fixing over at https://github.com/common-workflow-language/cwltool/pull/1353

So I'll merge that fix in here once that PR is complete

tests/test_singularity.py
```diff
@@ -127,3 +129,20 @@ def test_singularity_docker_image_id_in_tool(tmp_path: Path) -> None:
         ["--singularity", get_data("tests/debian_image_id.cwl"), "--message", "hello"]
     )
     assert result_code1 == 0
+
+
+@needs_singularity_3_or_newer # type: ignore
+def test_singularity_userns_error() -> None:
```
mr-c, line 135:
```suggestion
def test_singularity_userns_error() -> None:
    """
    Confirm the behaviour of Singularity 3.x --userns in Travis CI.
    
    Travis CI Xenial is not yet running a kernel with userspace namespaces enabled yet.
    """
```

to fix `D103: Missing docstring in public function`

Now that we have documented it, we see that this is not the best test. Our Travis CI setup runs Singularity 2.x, so this function is never run. The test function targets an external program, not the code that was introduced in this PR. And different users will have different experiences based upon their kernel version and configuration.

Since the changes are straight forward, and at least privately tested, I am okay if you remove this test.

jonathanxu18: Got it, just removed the test. Thanks again for the feedback!

PR #1363, opened 2020-11-09
support opaque resources

mr-c:
A bit brutal. Maybe at some point we should assume there are no opaque resource specifications?

### DIFF
--- a/cwltool/builder.py
+++ b/cwltool/builder.py
@@ -158,7 +158,7 @@ class Builder(HasReqsHints):
         names: Names,
         requirements: List[CWLObjectType],
         hints: List[CWLObjectType],
-        resources: Dict[str, Union[int, float]],
+        resources: Dict[str, Union[int, float, str]],
         mutation_manager: Optional[MutationManager],
         formatgraph: Optional[Graph],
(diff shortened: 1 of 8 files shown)


mvdbeek:
That does fix the failing unit test in https://github.com/common-workflow-language/galaxy, thanks @mr-c!

jmchilton:
Amazing, I will run the whole suite of green tests and see what I come up with.

> Maybe at some point we should assume there are no opaque resource specifications?

I think letting the DRM pick the resources is an important use case. If you want to make it realistic that you can generate command-lines to qsub, etc... from CWL without assuming a ton of plumbing inside the job I think this is an important capacity to allow. That said, I can appreciate this is indeed brutal.

mr-c:
> I think letting the DRM pick the resources is an important use case. If you want to make it realistic that you can generate command-lines to qsub, etc... from CWL without assuming a ton of plumbing inside the job I think this is an important capacity to allow. That said, I can appreciate this is indeed brutal.

Agreed! But it seems that at some point we have to round the number of cores and do other modifications. When in the execution process do we know what the resource manager picked for Galaxy-CWL execution?

Issue #888, opened 2018-08-29
File handles are leaked with --parallel


sersorrel:
cwltool fails with tracebacks like this on very large workflows when using `--parallel`:

```pytb
Failed to evaluate expression:
Expression evaluation error:
[Errno 24] Too many open files
Traceback (most recent call last):
  File "/home/ubuntu/cwltool/cwltool/expression.py", line 293, in do_eval
    strip_whitespace=strip_whitespace)
  File "/home/ubuntu/cwltool/cwltool/expression.py", line 231, in interpolate
    debug=debug, js_console=js_console)
  File "/home/ubuntu/cwltool/cwltool/expression.py", line 196, in evaluator
    debug=debug, js_console=js_console)
  File "/home/ubuntu/cwltool/cwltool/sandboxjs.py", line 334, in execjs
    fn, timeout=timeout, js_console=js_console, force_docker_pull=force_docker_pull, debug=debug)
  File "/home/ubuntu/cwltool/cwltool/sandboxjs.py", line 161, in exec_js_process
    new_proc = new_js_proc(js_engine_code, force_docker_pull=force_docker_pull)
  File "/home/ubuntu/cwltool/cwltool/sandboxjs.py", line 86, in new_js_proc
    dockerimgs = subprocess.check_output(["docker", "images", "-q", nodeimg]).decode('utf-8')
  File "/usr/lib/python3.6/subprocess.py", line 336, in check_output
    **kwargs).stdout
  File "/usr/lib/python3.6/subprocess.py", line 403, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.6/subprocess.py", line 667, in __init__
    errread, errwrite) = self._get_handles(stdin, stdout, stderr)
  File "/usr/lib/python3.6/subprocess.py", line 1177, in _get_handles
    c2pread, c2pwrite = os.pipe()
OSError: [Errno 24] Too many open files

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ubuntu/cwltool/cwltool/command_line_tool.py", line 91, in run
    ev = self.builder.do_eval(self.script)
  File "/home/ubuntu/cwltool/cwltool/builder.py", line 424, in do_eval
    strip_whitespace=strip_whitespace)
  File "/home/ubuntu/cwltool/cwltool/expression.py", line 296, in do_eval
    raise WorkflowException("Expression evaluation error:\n%s" % e)
[...]

PR #1367, opened 2020-11-12
Parallel execution uses thread pool to avoid leaking resources.

tetron:
Use task_queue so that threads are reused and don't leak resources.

refs #888 

Note: doesn't clean up node processes produced by sandboxjs.  Normally this won't be a problem since it only runs through the  `MultithreadedJobExecutor.run_jobs()` once.  However, if you're using `cwltool` from Python code and calling `run_jobs()` more than once, you might want to call `_terminate_processes()` in between.


mr-c (changes requested):
Two conformance tests failed:

Test 138 timed out: /home/travis/build/common-workflow-language/cwltool/common-workflow-language-main/cwltool_with_cov3 --parallel --outdir=/tmp/tmphs7__ko5 --quiet v1.0/count-lines18-wf.cwl v1.0/count-lines3-job.json

Test simple scatter over an embedded subworkflow

Terminating lingering process

Test 139 timed out: /home/travis/build/common-workflow-language/cwltool/common-workflow-language-main/cwltool_with_cov3 --parallel --outdir=/tmp/tmp_kszpo39 --quiet v1.0/count-lines14-wf.cwl v1.0/count-lines4-job.json

Test simple multiple input scatter over an embedded subworkflow

Terminating lingering process

mr-c:
Thank you @tetron for this contribution from the Arvados project!

I'm not yet able to reproduce these timeouts on Travis locally.

cwltool/task_queue.py
```diff
@@ -0,0 +1,119 @@
...
+            task()
+            return
+
+        while True:
+            try:
+                if unlock is not None:
+                    unlock.release()
+                if check_done is not None:
+                    if check_done.is_set():
+                        return
```
mr-c, line 97: Shouldn't `self.in_flight` be decremented in this case?

Does Arvados use `check_done`? If not, I would prefer to remove it from here.

cwltool/task_queue.py
```diff
@@ -0,0 +1,119 @@
...
+                self.task_queue.put(task, block=True, timeout=3)
```
mr-c, line 98: I'm surprised that you can't put more tasks into the queue than `thread_count` was set to.

What's the reasoning?

cwltool/task_queue.py
```diff
@@ -0,0 +1,119 @@
...
+        self.lock = lock
+        self.in_flight = 0
+        self.error: Optional[Exception] = None
+
+        for _r in range(0, self.thread_count):
+            t = threading.Thread(target=self._task_queue_func)
+            self.task_queue_threads.append(t)
+            t.start()
+
+    def _task_queue_func(self) -> None:
```
mr-c, line 52: I made this function private, as I don't see any reason for anyone outside this class to care about it.

cwltool/task_queue.py
```diff
@@ -0,0 +1,119 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+"""TaskQueue."""
+
+import queue
+import threading
+
+from typing import Callable, Optional
+
+from .loghandler import _logger
```
mr-c, line 11: I switched to the CWL logger, from the Arvados logger

cwltool/task_queue.py
tetron: Yes.  What it is used for is when (arvados-cwl-runner) catches an interrupt signal it want to gracefully shut down.  Part of that is telling the task queue to stop if it is blocked.

cwltool/task_queue.py
tetron: If you have N threads that are not doing anything yet, you can put up to N things in the queue and then each thread will have something to do.

If you have more than N things, you are going to end up blocking eventually because all the threads are busy.  This just means you have N things running and up to another N things queued before you block.  Making the queue unbounded size is unlikely to have any advantage in throughput, but it would mean you could end up using a bit more RAM if you created a  backlog of hundreds of pending jobs.

cwltool/task_queue.py
tetron: Sounds good to me.

cwltool/task_queue.py
tetron: Thanks

tetron:
Seems likely there's an infrequent race condition deadlock.  Not totally surprising, this stuff is hard.  Will give it another look.

tetron:
It is the same two tests failing on all three conformance suites, so there's definitely something wrong.  However I haven't been able to reproduce it locally.

```
Test timed out: /home/travis/build/common-workflow-language/cwltool/common-workflow-language-main/cwltool_with_cov3 --parallel --outdir=/tmp/tmplkc_pab0 --quiet v1.0/count-lines18-wf.cwl v1.0/count-lines3-job.json

Test simple scatter over an embedded subworkflow

Terminating lingering process

Test  timed out: /home/travis/build/common-workflow-language/cwltool/common-workflow-language-main/cwltool_with_cov3 --parallel --outdir=/tmp/tmp0f5lgsgn --quiet v1.0/count-lines14-wf.cwl v1.0/count-lines4-job.json

Test simple multiple input scatter over an embedded subworkflow

Terminating lingering process
```

mr-c:
Yeah, I couldn't reproduce it either. Maybe run it in docker with `--cpus` set to 1 or 2?

tetron:
With `queue size = 2` I was able to reproduce the deadlock (so I guess that means the travis environment reports 2 logical CPUs).  

It needed to release the `workflow_eval_lock` lock while blocking on the queue, otherwise the job runner can't finish, and since we need a running job to finish in order to free up a queue slot, this results in a deadlock.

Hopefully everything passes now!

cwltool/task_queue.py
tetron: You're right about decrementing `in_flight` when exiting `add()` before the item has gone into the queue.

mr-c (approved):

PR #1372, opened 2020-11-15
Fix cwltool-in-docker.sh to accept DOCKER_HOST

tom-tan:
This request makes `cwltool-in-docker.sh` to check the existence of the `DOCKER_HOST` environment variable in addition to the existence of `/var/run/docker.sock`.
This script is used in the [cwltool container image](https://hub.docker.com/r/commonworkflowlanguage/cwltool).

In the current implementation, it checks `/var/run/docker.sock` and aborts if it does not exist.
However, it is too restrictive because using `DOCKER_HOST` can achieve the same objective.
This request removes this restriction.


tom-tan:
Thanks!

Issue #1377, opened 2020-11-26
SoftwareRequirement: environment modules changes to env not propagating to tool


rupertnash:
## Expected Behavior

If I have a CWL file with a SoftwareRequirement that is resolved by an environment module, then I would expect that the environment variables set propagate to the tool execution.

## Actual Behavior
They do not: only `PATH` is modified by the environment module

## Workflow Code
Note this depends on having environment modules set up. Modify the cwl to load a module you have available:

env.cwl:
```
#!/usr/bin/env cwl-runner
# -*- mode: yaml; -*-

cwlVersion: v1.1
class: CommandLineTool
$namespaces:
  cwltool: http://commonwl.org/cwltool#

hints:
  SoftwareRequirement:
    packages:
      ncl:
        version: ["6.6.2"]

inputs: []

baseCommand: env

outputs:
  output:
    type: stdout
```
modules.yml:
```
- type: modules
  modulecmd: /usr/bin/modulecmd
```

cwltool runs OK:
```
 cwltool --beta-dependency-resolvers-configuration modules.yml env.cwl 
INFO /lustre/home/shared/dc118/sw/cwl/3.0.20201117141248/bin/cwltool 3.0.20201117141248
INFO Resolved 'env.cwl' to 'file:///lustre/home/dc118/nashvest/env.cwl'
INFO [job env.cwl] /tmp/7u0vetk_$ env > /tmp/7u0vetk_/d5514f0d054e63af38037aba76940ead71fd7324
Autoloading gcc/6.3.0
Autoloading gsl/2.6-gcc6
INFO [job env.cwl] completed success
{
    "output": {
        "location": "file:///lustre/home/dc118/nashvest/d5514f0d054e63af38037aba76940ead71fd7324",
        "basename": "d5514f0d054e63af38037aba76940ead71fd7324",
        "class": "File",
        "checksum": "sha1$488689425380b273e1e59a0dd3fdc884a57a842c",
        "size": 471,
        "path": "/lustre/home/dc118/nashvest/d5514f0d054e63af38037aba76940ead71fd7324"
    }
}
INFO Final process status is success
```

Module that should have been loaded:
```
$ module show ncl
-------------------------------------------------------------------
/lustre/sw/modulefiles/ncl/6.6.2:

conflict        ncl
Autoloading gcc/6.3.0
module load     gcc/6.3.0
Autoloading gsl/2.6-gcc6
module load     gsl/2.6-gcc6
[...]

PR #1378, opened 2020-11-27
WIP: fix interaction of software requirements and environments

rupertnash:
A very first attempt at the fix I proposed in #1377 

This now works for a the single test case on that issue. I'd like feedback before trying to do it more properly.

Thanks!

TODO:

- [x]  document the current behaviour

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -123,9 +123,15 @@ jobs:
     strategy:
       matrix:
         cwl-version: [v1.0, v1.1, v1.2]
+        container: [docker, singularity]
 
     steps:
       - uses: actions/checkout@v2
+
+      - name: Set up Singularity
+        uses: eWaterCycle/setup-singularity@v6
+        with:
+          singularity-version: ${{ env.singularity_version }}
 
       - name: Set up Python
         uses: actions/setup-python@v2
@@ -141,6 +147,8 @@ jobs:
       - name: Run CWL conformance tests ${{ matrix.cwl-version }}
         env:
           version: ${{ matrix.cwl-version }}
+          container: ${{ matrix.container }}
+          spec_branch: main
         run: ./conformance-test.sh
 
   release_test:
--- a/README.rst
+++ b/README.rst
@@ -296,6 +296,30 @@ CWL documents can be expressed as RDF triple graphs.
   cwltool --print-rdf --rdf-serializer=turtle mywf.cwl
 
 
+Environment Variables in cwltool
+--------------------------------
+
+This reference implementation supports several ways of setting
+enviroment variables for tools, in addition to the standard
+``EnvVarRequirement``. The sequence of steps applied to create the
+enviroment is:
+
+0. If the ``--preserve-entire-environment`` flag is present, then begin with the current
+   environment, else begin with an empty environment.
+
+1. Add any variables specified by ``--preserve-environment`` option(s).
+
+2. Set ``TMPDIR`` and ``HOME`` per `the CWL v1.0+ CommandLineTool specification <https://www.commonwl.org/v1.0/CommandLineTool.html#Runtime_environment>`_.
+
+3. Apply any ``EnvVarRequirement`` from the ``CommandLineTool`` description.
+
+4. Apply any manipulations required by any ``cwltool:MPIRequirement`` extensions.
+
+5. Substitute any secrets required by ``Secrets`` extension.
+
+6. Modify the environment in response to ``SoftwareRequirement`` (see below).
+
+
 Leveraging SoftwareRequirements (Beta)
 --------------------------------------
 
@@ -318,10 +342,11 @@ plugins to enable to "resolve" ``SoftwareRequirement`` dependencies.
 
 Using these hints will allow cwltool to modify the environment in
 which your tool runs, for example by loading one or more Environment
-Modules. The environment is constructed as normal (i.e. standard CWL
-runtime enviroment, ``EnvVarRequirement``, ``--preserve-environment``
-are applied), then the environment may modified by the selected tool resolver.  This currently means that you
-cannot override any environment variables set by the selected tool resolver.
+Modules. The environment is constructed as above, then the environment
+may modified by the selected tool resolver.  This currently means that
+you cannot override any environment variables set by the selected tool
+resolver. Note that the enviroment given to the configured dependency
+resolver has the variable `_CWLTOOL` set to `1` to allow introspection.
 
 To discuss some of these plugins and how to configure them, first consider the
 following ``hint`` definition for an example CWL tool.
--- a/README.rst
+++ b/README.rst
@@ -316,6 +316,13 @@ This option allows one to specify a dependency resolver's configuration file.
 This file may be specified as either XML or YAML and very simply describes various
 plugins to enable to "resolve" ``SoftwareRequirement`` dependencies.
 
+Using these hints will allow cwltool to modify the environment in
+which your tool runs, for example by loading one or more environment
+modules. The enviroment is constructed as normal (i.e. standard CWL
+runtime enviroment, ``EnvVarRequirement``, ``--preserve-environment``
+are applied), then this is modified. This currently means that you
+cannot override variables set in this way.
+
 To discuss some of these plugins and how to configure them, first consider the
 following ``hint`` definition for an example CWL tool.
 
@@ -427,7 +434,7 @@ So consider the resolvers configuration file
     base_path: ./tests/test_deps_env
     mapping_files: ./tests/test_deps_mapping.yml
 
-And the corresponding mapping configuraiton file (`tests/test_deps_mapping.yml`):
+And the corresponding mapping configuration file (`tests/test_deps_mapping.yml`):
 
 .. code:: yaml
 
--- a/README.rst
+++ b/README.rst
@@ -317,11 +317,11 @@ This file may be specified as either XML or YAML and very simply describes vario
 plugins to enable to "resolve" ``SoftwareRequirement`` dependencies.
 
 Using these hints will allow cwltool to modify the environment in
-which your tool runs, for example by loading one or more environment
-modules. The enviroment is constructed as normal (i.e. standard CWL
+which your tool runs, for example by loading one or more Environment
+Modules. The environment is constructed as normal (i.e. standard CWL
 runtime enviroment, ``EnvVarRequirement``, ``--preserve-environment``
-are applied), then this is modified. This currently means that you
-cannot override variables set in this way.
+are applied), then the environment may modified by the selected tool resolver.  This currently means that you
+cannot override any environment variables set by the selected tool resolver.
 
 To discuss some of these plugins and how to configure them, first consider the
 following ``hint`` definition for an example CWL tool.
--- a/conformance-test.sh
+++ b/conformance-test.sh
@@ -1,45 +1,70 @@
 #!/bin/bash
 venv() {
-        if ! test -d "$1" ; then
-                virtualenv -p python3 "$1"
-        fi
-	# shellcheck source=/dev/null
-        source "$1"/bin/activate
+    if ! test -d "$1" ; then
+	if command -v virtualenv > /dev/null; then
+            virtualenv -p python3 "$1"
+	else
+	    python3 -m venv "$1"
+	fi
+    fi
+    # shellcheck source=/dev/null
+    source "$1"/bin/activate
 }
+
+# Set these environment variables when running the script, e.g.:
+# version=v1.1 spec_branch=new_test container=docker ./conformance_test.sh
+
+# Version of the standard to test against
+# Current options: v1.0, v1.1, v1.2
 version=${version:-v1.0}
+
+# Which branch of the standard's repo to use.
+# This can be useful when adding new features
+spec_branch=${spec_branch:-main}
+
+# Which container runtime to use
+# Valid options: docker, singularity
+container=${container:-docker}
+
+set -e
+set -x
+
 if [[ "$version" = "v1.0" ]] ; then
-   wget https://github.com/common-workflow-language/common-workflow-language/archive/main.tar.gz
-   tar xzf main.tar.gz && rm main.tar.gz
+    repo=common-workflow-language
 else
     # shellcheck disable=SC2001
-    repo=$(echo "$version" | sed 's/\(v[0-9]*\.\)\([0-9]*\).*/\1\2/')  
-    wget https://github.com/common-workflow-language/cwl-"${repo}"/archive/main.tar.gz
-    tar xzf main.tar.gz && rm main.tar.gz
+    repo=cwl-$(echo "$version" | sed 's/\(v[0-9]*\.\)\([0-9]*\).*/\1\2/')
 fi
 
-docker pull node:slim
+if [ ! -d "${repo}-${spec_branch}" ]; then
+    if [ ! -f "${repo}-${spec_branch}.tar.gz" ]; then
+	wget "https://github.com/common-workflow-language/${repo}/archive/${spec_branch}.tar.gz"
+    fi
+    tar xzf "${spec_branch}.tar.gz"
+fi
+
+if [ "${container}" == "docker" ]; then
+    docker pull node:slim
+fi
+
+venv cwltool-venv3
+pip3 install -U setuptools wheel pip
+pip3 uninstall -y cwltool
+pip3 install -e .
+pip3 install codecov cwltest>=2.1
+pushd "${repo}-${spec_branch}" || exit 1
 
 # shellcheck disable=SC2043
-for CONTAINER in docker
-# for CONTAINER in docker singularity
-# singularity having issues on ci.commonwl.org; tests pass with https://gist.github.com/mr-c/0ec90d717617d074017c0cb38b72d1a4
-do
-	venv cwltool-venv3
-	pip3 install -U setuptools wheel pip
-	pip3 uninstall -y cwltool
-	pip3 install -e .
-	pip3 install "cwltest>=1.0.20180518074130" codecov
-	if [[ "$version" = "v1.0" ]]
-	then
-		DRAFT="DRAFT=v1.0"
-		pushd common-workflow-language-main || exit 1
-	else
-		pushd cwl-"${repo}"-main || exit 1
-	fi
-	rm -f .coverage* coverage.xml
-	source=$(realpath ../cwltool)
-	COVERAGE_RC=${PWD}/.coveragerc
-	cat > "${COVERAGE_RC}" <<EOF
+if [[ "$version" = "v1.0" ]]; then
+    DRAFT="DRAFT=v1.0"
+fi
+# Clean up all cov data
+find . -name '.coverage*' -print0 | xargs -0 rm -f
+rm -f coverage.xml
+
+source=$(realpath ../cwltool)
+COVERAGE_RC=${PWD}/.coveragerc
+cat > "${COVERAGE_RC}" <<EOF
 [run]
 branch = True
 source = ${source}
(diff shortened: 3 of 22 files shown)


mr-c (commented):
I like this! Wasn't as bad as I expected. Guess we'll have to add testing for other types of dependency resolvers, but this should work, yeah?

mr-c (approved):

mr-c:
@rupertnash Question: who wins, an environment variable set by `EnvVarRequirement` or an identical variable set by the dependency script?

I suspect that it is currently the later when it should be the former. Tricky because at this point `EnvVarRequirement` and the https://www.commonwl.org/v1.0/CommandLineTool.html#Runtime_environment variables are blended together. I guess I could add a check that favors `cwl_env`'s original values except for `HOME`, `TMPDIR`, and `PATH`. 

Speaking of, I think we should not allow `HOME` or `TMPDIR` to be overridden by the dependency script at all.

rupertnash:
Agreed about `HOME` and `TMPDIR`, I'll add them to the exceptions

Who wins? Hard to say. We first configure the environment as normal and then allow the module to modify it, potentially overriding ones set by `EnvVarRequirement`. However this could also be a desired modification e.g. of `LD_LIBRARY_PATH`.

I see that §5.6 in the spec it says a `SoftwareRequirement` is "A list of software packages that should be configured in the environment of the defined process." This clearly allows us to modify the environment and does not place limits on this. I can see no "hierarchy" of how the environment is created so with my pedant's hat on either behaviour is acceptable.

There are multiple sources of environment now:
- Core runtime environment vars (HOME, TMPDIR, PATH)
- Variables defined by `EnvVarRequirement`
- The default environment of the container, such as when using `DockerRequirement`
- Variables set/modified by `SoftwareRequirement`s
- Ones preserved by user (`--preserve-environment=VAR`/`--preserve-entire-environment`)
I think a clear description of how these interact would help us.

On the immediate problem, I'm torn and can definitely see cases when you'll want one behaviour not the other. E.g. if you want to `--preserve-environment LD_LIBRARY_PATH` and allow a module to append its own paths to that you want the current behaviour. On the other hand you may wish to override something set by a modulefile, eg some default.

mr-c (changes requested):
Okay, this needs a test and it will be good enough to merge!

mr-c:
> Who wins?

Whatever we decide, it needs documentation, yep

rupertnash:
TODO:
- [ ] document the current behaviour (next week)

What about this ordering? (I realise this might require some significant changes)

0. Start empty
1. Add preserved environment vars (or start with current env if entire env being preserved)
2. Set TMPDIR and HOME
3. Modify the environment in response to software requirements
4. Apply EnvVarRequirements

This will (I think) be more consistent with the way it's done for docker (i.e. the set env vars are applied last, starting from the ones set in the container)

mr-c:
@rupertnash It is a good ordering. I think 1-3 is already happening, but should be confirmed. To implement 4+5 we'll need to split the env list into two ; might not be that bad to implement

README.rst
```diff
@@ -292,6 +292,13 @@ This option allows one to specify a dependency resolver's configuration file.
 This file may be specified as either XML or YAML and very simply describes various
 plugins to enable to "resolve" ``SoftwareRequirement`` dependencies.
 
+Using these hints will allow cwltool to modify the environment in
+which your tool runs, for example by loading one or more environment
+modules. The enviroment is constructed as normal (i.e. standard CWL
```
mr-c, line 297:
```suggestion
which your tool runs, for example by loading one or more `environment modules
<https://en.wikipedia.org/wiki/Environment_Modules_(software)>`_. The environment
is constructed as normal (i.e. standard CWL
```

README.rst
```diff
@@ -292,6 +292,13 @@ This option allows one to specify a dependency resolver's configuration file.
...
+runtime enviroment, ``EnvVarRequirement``, ``--preserve-environment``
+are applied), then this is modified. This currently means that you
```
mr-c, line 299:
```suggestion
are applied), then the environment may modified by the selected tool resolver.  This currently means that you
```

README.rst
```diff
@@ -292,6 +292,13 @@ This option allows one to specify a dependency resolver's configuration file.
...
+cannot override variables set in this way.
```
mr-c, line 300:
```suggestion
cannot override any environment variables set by the selected tool resolver.
```

cwltool/run_job.py
```diff
@@ -0,0 +1,103 @@
+"""Only used when there is a job script or CWLTOOL_FORCE_SHELL_POPEN=1."""
```
mr-c, line 1: We should set this file to be excluded from test coverage

mr-c (changes requested):
https://github.com/common-workflow-language/cwltool/pull/1378/files#diff-08aa5643b92eb585337d3151a28a6bb8905e482a1094f90e0dd90939d8ee4d8cR958 seems to not be covered by any tests?

Lets also mark cwltool/run_job.py as excluded from the test coverage requirements.

rupertnash:
OK - I think I got the coverage to exclude run_job.py.

For your point about not testing that line: in `test_dependencies`, this is being hit twice iff `module` is available on the host. I don't know if this is possible to add to Travis?

I also noticed a bug in the test skip logic: the first 2 tests in that file have probably never run...

mr-c:
> OK - I think I got the coverage to exclude run_job.py.

Confirmed, thanks!

> For your point about not testing that line: in `test_dependencies`, this is being hit twice iff `module` is available on the host. I don't know if this is possible to add to Travis?

http://modules.sourceforge.net/ ?

tests/test_dependencies.py
```diff
@@ -35,14 +41,18 @@ def test_bioconda() -> None:
...
 
+@pytest.mark.skipif(not deps, reason="galaxy-lib is not installed")
 @pytest.mark.skipif(
     not spawn.find_executable("modulecmd"), reason="modulecmd not installed"
 )
-def test_modules() -> None:
+def test_modules(monkeypatch: Any) -> None:
+    """Here we do a basic smoke test using environment modules to
+    satisfy a SoftwareRequirement.
+    """
```
mr-c, line 51:
```suggestion
    """Do a basic smoke test using environment modules to satisfy a SoftwareRequirement."""
```

To fix https://travis-ci.com/github/common-workflow-language/cwltool/jobs/454295930#L4188

tests/test_dependencies.py
```diff
@@ -55,3 +65,43 @@ def test_modules() -> None:
...
+
+@pytest.mark.skipif(not deps, reason="galaxy-lib is not installed")
+@pytest.mark.skipif(
+    not spawn.find_executable("modulecmd"), reason="modulecmd not installed"
+)
+def test_modules_environment(tmp_path: Path, monkeypatch: Any) -> None:
+    """Here check that the environment variables set by a module are
+    being propagated into a tool's environment, by running `env` as the
+    tool and parsing its output.
+    """
```
mr-c, line 78:
```suggestion
    """
    Check that the environment variables set by a module are being propagated correctly.

    Do so by  by running `env` as the tool and parsing its output.
    """
```

To fix https://travis-ci.com/github/common-workflow-language/cwltool/jobs/454295930#L4191

rupertnash:
OK - having just looked at the Travis log, this one isn't my fault: it's because of the tests that never ran due to the bug I fixed in e1c3325 above

Looks like the member vars of `software_requirements.DependencyConfiguration` aren't being set up correctly.

mr-c:
Some notes from today's CWL chat for @rupertnash 

What about `CLASSPATH` set by an `EnvVarRequirement`? I wouldn't want that to overwrite a `CLASSPATH` set by a `SoftwareRequirement`, I would want them appended. So `EnvVarRequirement` should be set before `SoftwareRequirement`s so `module load`, `conda install` can append to env vars that are lists like `CLASSPATH`

As for `HOME` and `TMPDIR`, that should be set early and set again at the end, just in case it got overwritten.

rupertnash:
Following CWL chat just now:
0. Start empty
1. Add preserved environment vars (or start with current env if entire env being preserved)
2. Set `TMPDIR` and `HOME` as per https://www.commonwl.org/v1.2/CommandLineTool.html#Runtime_environment
3. Set `_CWLTOOL=1` (to allow modules etc be able to know when running under CWL)
3. Apply `EnvVarRequirements`
4. Modify the environment in response to software requirements
5. Ensure `HOME` is set as above, remove `_CWLTOOL`

rupertnash:
OK - I've rebased this onto main. Going to look at the todo list now

README.rst
```diff
@@ -298,6 +298,30 @@ CWL documents can be expressed as RDF triple graphs.
...
 
+Environment Variables in cwltool
+--------------------------------
+
+This reference implementation supports several ways of setting
+enviroment variables for tools, in addition to the standard
+``EnvVarRequirement``. The sequence of steps applied to create the
+enviroment is:
+
+0. If ``--preserve-entire-environment`` is set, begin with the current
```
mr-c, line 309:
```suggestion
0. If the ``--preserve-entire-environment`` flag is present, then begin with the current
```

README.rst
```diff
@@ -298,6 +298,30 @@ CWL documents can be expressed as RDF triple graphs.
...
+   environment, else begin with an empty environment.
+
+1. Add any variables specified by ``--preserve-environment`` option(s).
+
+2. Set ``TMPDIR`` and ``HOME`` per the CWL specification.
```
mr-c, line 314:
```suggestion
2. Set ``TMPDIR`` and ``HOME`` per `the CWL v1.0+ CommandLineTool specification <https://www.commonwl.org/v1.0/CommandLineTool.html#Runtime_environment>`_.
```

README.rst
```diff
@@ -298,6 +298,30 @@ CWL documents can be expressed as RDF triple graphs.
...
+
+3. Apply ``EnvVarRequirement`` from tool description.
```
mr-c, line 316:
```suggestion
3. Apply any ``EnvVarRequirement`` from the ``CommandLineTool`` description.
```

README.rst
```diff
@@ -298,6 +298,30 @@ CWL documents can be expressed as RDF triple graphs.
...
+
+4. Apply any manipulations required by ``MPIRequirement`` extension.
```
mr-c, line 318:
```suggestion
4. Apply any manipulations required by any ``cwltool:MPIRequirement`` extensions.
```

README.rst
```diff
@@ -298,6 +298,30 @@ CWL documents can be expressed as RDF triple graphs.
...
+
+5. Substitute any secrets required by ``Secrets`` extension.
+
+6. Modify the environment in response to ``SoftwareRequirement``.
```
mr-c, line 322:
```suggestion
5. Substitute any secrets required by ``Secrets`` extension.

6. Modify the environment in response to ``SoftwareRequirement`` (see below).
```

cwltool/run_job.py
```diff
@@ -0,0 +1,106 @@
...
+        else:
+            stdout = sys.stderr
+        if stderr_path is not None:
+            stderr: Union[BinaryIO, TextIO] = open(stderr_path, "wb")
+        else:
+            stderr = sys.stderr
+        if os.name == "nt":
+            close_fds = False
+            for key, value in env.items():
+                env[key] = str(value)
```
mr-c, line 72: This case can be dropped; see https://github.com/common-workflow-language/cwltool/pull/1423 which is about to be merged

cwltool/singularity.py
```diff
@@ -89,6 +89,8 @@ def _normalize_sif_id(string: str) -> str:
 
 
 class SingularityCommandLineJob(ContainerCommandLineJob):
+    CONTAINER_TMPDIR: str = "/tmp"  # nosec
```
mr-c, line 92: Should `CONTAINER_TMPDIR` move into `ContainerCommandLineJob`?

rupertnash: maybe?

cwltool/run_job.py
rupertnash: Does this mean all the `if Windows` checks can be removed?

mr-c: Yep! Sorry about all the pending merge conflicts..

cwltool/singularity.py
rupertnash: Yes it can; I will update

cwltool/job.py
```diff
@@ -613,6 +617,19 @@ def add_writable_directory_volume(
     ) -> None:
         """Append a writable directory mapping to the runtime option list."""
 
+    def _preserve_environment_on_containers_warning(
+        self, varname: Optional[str] = None
+    ) -> None:
+        """When running in a container, issue a warning."""
+        if varname is None:
+            _logger.warning(
+                "You have specified `--preserve-entire-environment` while running a container"
```
mr-c, line 626: We should say why this results in a warning (could hamper reproducibility / portability)

cwltool/singularity.py
```diff
@@ -370,6 +370,12 @@ def add_writable_directory_volume(
                 self.append_volume(runtime, source, volume.target, writable=True)
                 ensure_writable(source)
 
+    def _required_env(self) -> Dict[str, str]:
+        return {
+            "TMPDIR": self.CONTAINER_TMPDIR,
+            "HOME": self.builder.outdir,
```
mr-c, line 376: `WARNING: Overriding HOME environment variable with SINGULARITYENV_HOME is not permitted`

https://github.com/common-workflow-language/cwltool/pull/1378/checks?check_run_id=2237763816#step:9:100

```suggestion
            "SINGULARITYENV_HOME"": self.builder.outdir,
```

rupertnash:
Ugh, need to handle env vars containing new lines by using this instead of `env`
```
import json
import os
print(json.dumps(dict(os.environ), indent=4))
```

rupertnash:
I thought I was?

rupertnash:
Right, that looks like it's working now!

I need to give the docs a once over.

I have run the conformance tests under Singularity (3.7) and all have passed except 173.
```
Test 173 failed: /lustre/home/dc118/nashvest/cwltool/common-workflow-language-main/cwltool_with_cov3 --parallel --singularity --outdir=/tmp/tmpg_fy7936 --quiet v1.0/docker-run-cmd.cwl v1.0/empty.json
Test Docker ENTRYPOINT usage
Returned non-zero
INFO:    Using cached SIF image
FATAL:   "-c": executable file not found in $PATH
ERROR [job docker-run-cmd.cwl] Job error:
("Error collecting output for parameter 'cow':\nv1.0/docker-run-cmd.cwl:14:7: Did not find output file with glob pattern: '['cow']'", {})
WARNING [job docker-run-cmd.cwl] completed permanentFail
WARNING Final process status is permanentFail

196 tests passed, 1 failures, 0 unsupported features
```

This test is problematic for Singularity. The test runs the tool "v1.0/docker-run-cmd.cwl", who's definition is:
```yaml
class: CommandLineTool
cwlVersion: v1.0
hints:
  ResourceRequirement:
    ramMin: 8
requirements:
  DockerRequirement:
    dockerPull: bash:4.4.12
inputs: []
outputs:
  cow:
    type: File
    outputBinding:
      glob: cow
baseCommand: ["-c", "echo 'moo' > cow"]
```

Singularity has two variants 
- `singularity exec [<exec options>] <container> <command> [<args>]` to run a user-specified command in a container, and
- `singularity run [<options>] <container> [<args>]` to run the image's default command in a container

While docker is, according to docs `docker run image [command] [args]`  but this isn't the whole story because `ENTRYPOINT` and `CMD` mess around with this, both of which are set in this container to make it act a lot like just executing `bash` on your terminal. 

Reading the CWL spec (https://www.commonwl.org/v1.0/CommandLineTool.html#DockerRequirement):

> When running a tool contained in Docker, the workflow platform must not assume anything about the contents of the Docker container, such as the presence or absence of specific software, except to assume that the *generated command line represents a valid command within the runtime environment of the container*.

I conclude that the behaviour of this test is contrary to the wording of the standard since `["-c", "echo 'moo' > cow"]` is not a valid *command* in either shell or docker language.

mr-c:
> Right, that looks like it's working now!

Huzzah!

> I have run the conformance tests under Singularity (3.7) and all have passed except 173.

Yeah, this is the one that never passed for me. Here's the discussion I had with the Singularity devs about it https://github.com/hpcng/singularity/issues/3673#issuecomment-517691733

In short: can you restore the running of the conformance tests with Singularity, but skipping test 173? And then we can open another issue to follow this further.

Then I can do another review here and hopefully merge 😄

rupertnash:
> In short: can you restore the running of the conformance tests with Singularity, but skipping test 173? 
Yes, just trying to figure our how to skip tests with `cwltest` - any clues?

mr-c:
> > In short: can you restore the running of the conformance tests with Singularity, but skipping test 173? 
> 
> Yes, just trying to figure our how to skip tests with `cwltest` - any clues?
> 

`cwltest -n1-172,174-${last_test_number}` or fix https://github.com/common-workflow-language/cwltest/issues/111 😄

tetron:
v1.0 didn't specify what to do in the presence of ENTRYPOINT.  It was addressed in v1.1:

https://www.commonwl.org/v1.1/CommandLineTool.html#DockerRequirement

> A container image may specify an ENTRYPOINT and/or CMD. Command line arguments will be appended after all elements of ENTRYPOINT, and will override all elements specified using CMD (in other words, CMD is only used when the CommandLineTool definition produces an empty command line).

Unfortunately this is a situation where we were codifying existing de-facto behavior.

How does does the behavior of `singularity run` differ from Docker ENTRYPONT?

rupertnash:
Not significantly

tetron:
> Not significantly

So in that case, the correct behavior is probably to use `singularity run` not `singularity exec` ?  Is it possible to have an null default script in which case the fallback case for `run` is the same as `exec`?

conformance-test.sh
```diff
@@ -6,36 +6,33 @@ venv() {
...
 	pip3 install -e .
-	pip3 install "cwltest>=1.0.20180518074130" codecov
-	if [[ "$version" = "v1.0" ]]
-	then
-		DRAFT="DRAFT=v1.0"
-		pushd common-workflow-language-main || exit 1
-	else
-		pushd cwl-"${repo}"-main || exit 1
+	pip3 install codecov
+	pip3 install git+https://github.com/common-workflow-language/cwltest.git@main#cwltest
```
mr-c, line 31: `cwltest` version 2.1 is out with your changes now :-) https://pypi.org/project/cwltest/2.1.20210423130452/

rupertnash:
@mr-c the coverage part of conformance-test.sh is a bit shonky. I've mostly fixed this running locally but what about the upload part? This errors out with:
```
+ codecov --file coverage.xml

      _____          _
     / ____|        | |
    | |     ___   __| | ___  ___ _____   __
    | |    / _ \ / _  |/ _ \/ __/ _ \ \ / /
    | |___| (_) | (_| |  __/ (_| (_) \ V /
     \_____\___/ \____|\___|\___\___/ \_/
                                    v2.1.11

==> Detecting CI provider
  -> Got branch from git/hg
  -> Got sha from git/hg
==> Preparing upload
    Error running `['hg', 'locate']`: returncode=None, output=[Errno 2] No such file or directory: 'hg'
    Error running `['hg', 'locate']`: returncode=None, output=[Errno 2] No such file or directory: 'hg'
==> Processing gcov (disable by -X gcov)
==> Collecting reports
    Targeting specific files
    + coverage.xml bytes=407851
==> Uploading
    .url https://codecov.io
    .query commit=a76fed55b5f9fb726b1a2c8144c656743849e160&branch=deps-env&package=py2.1.11
    Gzipping contents..
    Compressed contents to 29094 bytes
    Pinging Codecov...
Error: Could not determine repo and owner
```

.github/workflows/ci-tests.yml
```diff
@@ -141,6 +146,7 @@ jobs:
       - name: Run CWL conformance tests ${{ matrix.cwl-version }}
         env:
           version: ${{ matrix.cwl-version }}
+          spec_branch: exclude
```
mr-c, line 149: This branch was merged into `main`
```suggestion
```

rupertnash: Thanks - I'll do PRs for the other spec repos then can replace this.

rupertnash: done

rupertnash:
OK, I'm very confused about the failing test in 1.1 (199). It runs `wc` on whale.txt but the output mismatches.

If I run a shell in the container via singularity we can see different behaviour:
```
Singularity> wc cwl-v1.1-main/tests/whale.txt 
  16  198 1111 cwl-v1.1-main/tests/whale.txt
Singularity> wc < cwl-v1.1-main/tests/whale.txt 
  16  198 1111
Singularity> cat cwl-v1.1-main/tests/whale.txt | wc
     16     198    1111
Singularity> cat cwl-v1.1-main/tests/whale.txt | wc | sha1sum 
2936abd60952276ffbf91d709b19c6c48c263947  -
```
The last gives the expected answer (and the one that docker gives).

There must be some difference in the way singularity is passing file handles to the process in the container vs docker.

rupertnash:
I'm just going to disable for now, see #1440

tetron:
@rupertnash argh, this has come up before, the `wc` program differs slightly in its output, either across implementations, or depending on whether you're writing to a PTY or a file.  I there's a longstanding bug report to do something about that.

mr-c (commented):
Just one final Singularity + CWL v1.2 test failure: "Test directory literal containing a real file"

https://github.com/common-workflow-language/cwltool/pull/1378/checks?check_run_id=2421911722#step:6:880

I think it is unrelated and can be moved to another issue

mr-c (approved):

mr-c:
@rupertnash Do you want to clean up the git history for a simple rebase onto `main` ; or shall I squash down to a single mega-commit?

rupertnash:
> @rupertnash Do you want to clean up the git history for a simple rebase onto `main` ; or shall I squash down to a single mega-commit?

I am not a bit fan of mega commits. I've streamlined the history in what I just forced (and rebased onto main). Please merge if you're happy!!

PR #1380, opened 2020-11-30
Use existing location if a secondary file has one.

DailyDreaming:
This allows Toil to pass v1.2 conformance test 307.

In Toil, we pass the following (example) inputs to the executor:

```
{
    "inputWithSecondary": {
        "class": "File",
        "location": "toilfs:0:files/no-job/file-1120f3b05d62444083a9ff5c60c7f772/secondary_file_test.txt",
        "size": 0,
        "basename": "secondary_file_test.txt",
        "nameroot": "secondary_file_test",
        "nameext": ".txt",
        "secondaryFiles": [
            {
                "location": "toilfs:0:files/no-job/file-ccead684521d44f98ff81b118f37159b/secondary_file_test.txt.accessory",
                "class": "File",
                "basename": "secondary_file_test.accessory",
                "nameroot": "secondary_file_test",
                "nameext": ".accessory",
                "size": 0
            }
        ]
    }
}
```

When cwltool makes its builder, it swaps:

toilfs:0:files/no-job/**file-ccead684521d44f98ff81b118f37159b**/secondary_file_test.txt.accessory

for

toilfs:0:files/no-job/**file-1120f3b05d62444083a9ff5c60c7f772**/secondary_file_test.txt.accessory

The bolded portion is the same basedir as the primary file, which creates a path to a file that does not exist.

It does this here: https://github.com/common-workflow-language/cwltool/blob/78fe9d41ee5a44f8725dfbd7028e4a5ee42949cf/cwltool/builder.py#L428

This seems to be because cwltool:
1. Assumes that primary and secondary files are in the same directory and thus have the same interchangeable basedir in the `toilfs:` schema.
1. Does not check to see if secondaryfiles have been resolved yet.  (the list has resolved secondary files in toil because we process with a builder already before it runs a builder again in cwltool)

Two possible proposed solutions:
 - The easier solution imo (this PR) is to check for an already processed secondary file before fabricating a new secondary file's entry.
 - An alternative solution might be to force toil to place secondary files at the same basedir as the primary file if the first solution isn't preferable.

### DIFF
--- a/cwltool/builder.py
+++ b/cwltool/builder.py
@@ -465,7 +465,7 @@ class Builder(HasReqsHints):
                                             + sfname
                                         )
                                     else:
-                                        sf_location = d_location + sfname
+                                        sf_location = f'{d_location}{sfname}'
                                 sfbasename = sfname
                             elif isinstance(sfname, MutableMapping):
                                 sf_location = sfname["location"]
--- a/cwltool/builder.py
+++ b/cwltool/builder.py
@@ -1,6 +1,7 @@
 import copy
 import logging
 import math
+import os
 from typing import (
     IO,
     Any,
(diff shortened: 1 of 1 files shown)


mr-c (changes requested):
Can you run `make mypy` address the issues?

https://travis-ci.com/github/common-workflow-language/cwltool/jobs/453580948#L628

DailyDreaming:
@mr-c The linting should be addressed (although the checking is perhaps a bit verbose; please advise if I you'd prefer I handle it differently).  One test is failing because `run_test.sh` in `cwl-v1.2` no longer has executable permissions.  I made a PR here and tests should pass once it's merged: https://github.com/common-workflow-language/cwl-v1.2/pull/74

mr-c:
We are out of Travis CI coins, sorry @DailyDreaming ; If you have Travis CI coins in your personal account, you could fork the repo and run them there

DailyDreaming:
@mr-c Sorry I used your coins.  Will be much more conservative about running tests in the future.

All tests passed when run earlier except for the cwl-v1.2 tests, which passed locally for me after the execution bit fix.  I'll just wait on Travis here then.

mr-c:
@DailyDreaming Don't worry about it! There are a lot of projects that pull from the same bucket of coins.

DailyDreaming:
Seems like this may have been fixed already.  Updating cwltool in Toil seems to have resolved this issue.  Closing!

PR #1384, opened 2020-12-03
No useless dirs

tetron:
* Fix long standing bug creating useless empty temp directories, cluttering up user's filesystem
* Fix correctness issue getting `loadListing` from `outputBinding`.

### DIFF
--- a/cwltool/command_line_tool.py
+++ b/cwltool/command_line_tool.py
@@ -1214,7 +1214,7 @@ class CommandLineTool(Process):
                     rfile = files.copy()
                     revmap(rfile)
                     if files["class"] == "Directory":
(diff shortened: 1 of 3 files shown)


mr-c (changes requested):
Oh, I was looking at something similar; this makes a lot more sense. Can we get some tests as well?

tetron:
The `loadListing` bugfix definitely needs a conformance test.

This business with creating useless output directories is a little harder to test for.  Maybe it could run a workflow and then check that there are no unexpected directories left over.

tetron:
https://github.com/common-workflow-language/cwl-v1.2/pull/69

mr-c:
Oops, guess I need to tweak the mergify config

PR #1386, opened 2020-12-05
singularity > 3: use SINGULARITY_BIND for longer bind lists

mr-c:
- [ ] install Singularity v3 in Travis?

### DIFF
--- a/cwltool/docker.py
+++ b/cwltool/docker.py
@@ -246,9 +246,8 @@ class DockerCommandLineJob(ContainerCommandLineJob):
             return cast(Optional[str], r["dockerImageId"])
         raise WorkflowException("Docker image %s not found" % r["dockerImageId"])
 
-    @staticmethod
     def append_volume(
-        runtime: List[str], source: str, target: str, writable: bool = False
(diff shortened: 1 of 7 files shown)


tetron:
This might not work.  From my research, memory used to pass arguments is shared by the command line and the environment.  Switching over to an environment variable may use marginally less memory because it will be one big string and not an array of strings (so we reclaim the space used by the pointers into the arguments array) but if we are over by a lot it won't magically fix it.

The maximum memory used for arguments is determined by `ARG_MAX`, which you can look up with `getconf ARG_MAX`.  On Linux this is 1/4 of the stack limit.  On my system the stack limit is 8 MiB so ARG_MAX is 2 MiB.

So an alternate/complementary solution would be to catch the OSError from subprocess and check for E2BIG (Argument list too long) and try doubling the stack limit with [resource.setrlimit](https://docs.python.org/3/library/resource.html?highlight=setrlimit#resource.setrlimit).

Although, setting thousands of individual binds is still going to be kind of inefficient.  A better solution overall would be to create a hardlink tree so that only the base directory needs to be mounted into the container.

PR #1392, opened 2020-12-17
make unit tests run in-process for speed

rupertnash:
As discussed on the call on Tuesday, this commit replaced the call to `subprocess.Popen` with a direct call to `cwltool.main.main` to avoid the overhead of starting a new process and re-importing all the modules.

No idea how robust this is. One obvious thing to try would be to add have both implementations available so tests that NEED to be run in a new process can be.


mr-c:
Travis CI running over at https://travis-ci.com/github/mr-c/cwltool/builds/209417043

mr-c:
https://travis-ci.com/github/mr-c/cwltool/jobs/461892396#L3844

```
=========================== short test summary info ============================
FAILED tests/test_examples.py::test_cid_file_dir_arg_is_file_instead_of_dir[]
FAILED tests/test_examples.py::test_cid_file_dir_arg_is_file_instead_of_dir[--debug]
FAILED tests/test_examples.py::test_cid_file_non_existing_dir[] - SystemExit: 2
FAILED tests/test_examples.py::test_cid_file_non_existing_dir[--debug] - Syst...
FAILED tests/test_singularity.py::test_singularity_pullfolder - AssertionError
```

also  https://travis-ci.com/github/mr-c/cwltool/jobs/461892392#L467

tests/util.py
```diff
@@ -86,19 +88,18 @@ def get_main_output(
...
-        env=env,
-    )
-
-    stdout, stderr = process.communicate()
+    stdout = io.StringIO()
+    stderr = io.StringIO()
+    if env is not None:
+        assert monkeypatch is not None
+        monkeypatch.setattr(os, "environ", env)
+    rc = main(argsl=args, stdout=stdout, stderr=stderr)
```
mr-c, line 98:
```suggestion
    try:
        rc = main(argsl=args, stdout=stdout, stderr=stderr)
    except SystemExit as e:
        rc = e.code
``

File: tests/util.py [hunk 2]
```diff
@@ -86,19 +88,22 @@ def get_main_output(
...
+    stdout = io.StringIO()
+    stderr = io.StringIO()
+    if env is not None:
+        assert monkeypatch is not None
+        monkeypatch.setattr(os, "environ", env)
+    try:
+        rc = main(argsl=args, stdout=stdout, stderr=stderr)
+    except SystemExit as e:
+        rc = e.code
+``
```
mr-c, line 102:
```suggestion
```

rupertnash:
Instead of this approach, we could consider the below (add to tests/conftest.py) which uses multiprocessing in forkserver mode?

(Test cases would have to use this fixture instead of importing `util.get_main_output`, but it's only the function signatures that would have to change)

```
@pytest.fixture(scope="session")
def forkserver_pool():
    import multiprocessing
    ctx = multiprocessing.get_context("forkserver")
    ctx.set_forkserver_preload("cwltool.main")
    pool = ctx.Pool(processes=1, maxtasksperchild=1)
    yield pool
    pool.close()
    pool.join()

class MainOutputGetter:
    def __init__(self, pool):
        self.pool = pool

    def __call__(
        self,
        args: List[str],
        env: Union[
            Mapping[bytes, Union[bytes, str]], Mapping[str, Union[bytes, str]], None
        ] = None,
    ) -> Tuple[Optional[int], str, str]:
        return self.pool.apply(gmo_impl, (args, env))

@pytest.fixture
def get_main_output(forkserver_pool):
    return MainOutputGetter(forkserver_pool)
```

And have a helper module with approximately this branch's implementation of `get_main_output`

mr-c:
> Instead of this approach, we could consider the below (add to tests/conftest.py) which uses multiprocessing in forkserver mode?

What's the perceived benefit of this other method?

rupertnash:
Gives the benefit of isolating child processes from each other while mitigating much of module import time

mr-c:
Only one test is failing now, and I don't think it is due to lack of process isolation: https://github.com/common-workflow-language/cwltool/runs/2078730738#step:9:163

rupertnash:
"get main output implementation" a function that is basically this branch's version of `get_main_output`

This was just a vague thought in response to seeing your activity on this PR. I recently saw a tweet about using forkserver with preloaded modules and wrote the above. It's not been run....

mr-c:
> This was just a vague thought in response to seeing your activity on this PR. I recently saw a tweet about using forkserver with preloaded modules and wrote the above. It's not been run....

Lets finish the approach implemented in this PR first and then we can evaluate further enhancements :-)

rupertnash:
Right - the test is failing because pytest is capturing the logged warning about content types. When run in a separate process that all gets put to the actual stderr FD and captured by `get_main_output`. Here, I guess the logging module is writing to `sys.stderr` which pytest is grabbing for us per-test.

I see that pytest has some fixtures to help with this - https://docs.pytest.org/en/latest/capture.html?highlight=stderr

rupertnash:
Re. `test_content_type.py`: why I have 2 questions:

1. Why `import pydot` at the top?
2. Why the loop `for test_file in ("js_output.cwl", "js_output_workflow.cwl")` when `test_file` is never referenced?

mr-c:
> Re. `test_content_type.py`: why I have 2 questions:
> 
>     1. Why `import pydot` at the top?
> 
>     2. Why the loop `for test_file in ("js_output.cwl", "js_output_workflow.cwl")` when `test_file` is never referenced?

Good questions, lets punt that to another issue for another day?

Is this PR ready for review & merging?

mr-c (approved):

PR #1398, opened 2021-01-24
upgrade acceptable rdflib version

mr-c:


mr-c:
TravisCI passed in https://github.com/mr-c/cwltool/pull/3, so force merging here.

PR #1412, opened 2021-03-08
CI: only create a local branch for PRs

rupertnash:
Hopefully resolve why this passes:
https://github.com/common-workflow-language/cwltool/runs/2058705861?check_suite_focus=true
and this fails:
https://github.com/common-workflow-language/cwltool/runs/2058861887?check_suite_focus=true


rupertnash:
Might want to manually push this commit to main to check that it works, rather than merging here?

PR #1415, opened 2021-03-09
--single-step to extract/run a single step

### LINKED ISSUE #1407: Feature request: Add option to execute only a single workflow step
I would like to execute just a single step of a workflow.

For example, call cwltool with `--target wf/step!` and have only that step executed, not the whole workflow starting from that step.

Update: I think it would be even better to generalize the interface, so that one could specify a range of steps to run. So, for example `--target wf/stepB-stepB` would run just step B and `--target wf/stepB-stepD` would run steps B, C, and D.

mr-c:
Should fix #1407


jdidion (approved):
Great, looks like it works!

PR #1416, opened 2021-03-09
Secondaryfile deps

tetron:


cwltool/process.py
```diff
@@ -1153,7 +1153,15 @@ def scandeps(
                 if doc["class"] == "Directory" and "listing" in doc:
                     deps["listing"] = doc["listing"]
                 if doc["class"] == "File" and "secondaryFiles" in doc:
-                    deps["secondaryFiles"] = doc["secondaryFiles"]
+                    deps["secondaryFiles"] = scandeps(
```
mr-c, line 1156: needs a `cast` here

mr-c: https://github.com/common-workflow-language/cwltool/runs/2076333148?check_suite_focus=true#step:9:26

PR #1446, opened 2021-05-18
Fix erroneous URL encoding of final output filenames

### LINKED ISSUE #1445: CommandLineTool output filenames are URL encoded
## Expected Behavior
CommandLineTool outputs whose filenames contain special characters should retain those special characters when copied to the final output directory. For example, if a CommandLineTool writes an output file with the name "(parentheses).txt", then the filename should remain "(parentheses).txt" when copied/moved to the final output directory.

## Actual Behavior
Filenames of CommandLineTool outputs are URL encoded after being copied/moved to the final output directory. For example, an output filename of "(parentheses).txt" becomes "%28parentheses%29.txt" in `command_line_tool.collect_output()`. 

## Cursory Analysis
This happens because `fs_access.glob()` returns a file URI, which is itself URL encoded, and `command_line_tool.collect_output()` properly calls `urllib.parse.unquote()` to obtain the URL decoded **path** but fails to do so for the **basename** and **nameroot** fields. This record's **basename** is then used as the destination name for outputs copied/moved to final output directory.

### url_encode_test.cwl
```
class: CommandLineTool
cwlVersion: v1.0

requirements: 

baseCommand: touch

inputs:
    inp:
        type: string
        inputBinding:
            position: 0
outputs:
    out_file:
        type: File
        outputBinding:
            glob: $(inputs.inp)
```

### job.json
`{"inp": "(parentheses).txt"}`

## Full Traceback
N/A

## Your Environment
* cwltool version: 3.1.20210426140515

AlexTate:
Fixes #1445; command_line_tool.collect_output() now URL decodes output file `basename`s, and uses the decoded result to determine the `nameroot` and `nameext` fields

### DIFF
--- a/tests/test_path_checks.py
+++ b/tests/test_path_checks.py
@@ -2,15 +2,15 @@ import pytest
 import urllib.parse
 
 from pathlib import Path
-from typing import cast, MutableMapping
-from ruamel.yaml.comments import CommentedMap, CommentedSeq
+from typing import cast
+from ruamel.yaml.comments import CommentedMap
 from schema_salad.sourceline import cmap
 
 from cwltool.main import main
 from cwltool.command_line_tool import CommandLineTool
 from cwltool.context import LoadingContext, RuntimeContext
 from cwltool.update import INTERNAL_VERSION
-from cwltool.utils import CWLObjectType, CWLOutputType
+from cwltool.utils import CWLObjectType
 from .util import needs_docker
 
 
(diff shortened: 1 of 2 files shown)


mr-c (changes requested):
Thank you @AlexTate for this PR. Can you add a test to demonstrate the problem and the fix?

mr-c (changes requested):
Please also run `make format` :-)

mr-c (changes requested):
https://github.com/common-workflow-language/cwltool/pull/1446/files shows a couple lines without test coverage; can you investigate?

tests/test_path_checks.py
```diff
@@ -95,3 +103,66 @@ def test_unicode_in_output_files(tmp_path: Path, filename: str) -> None:
         filename,
     ]
     assert main(params) == 0
+
+
+def test_clt_returns_specialchar_names(tmp_path: Path) -> None:
```
mr-c, line 108: This needs a docstring: https://github.com/common-workflow-language/cwltool/pull/1446/checks?check_run_id=2668401511#step:7:26

AlexTate:
Of course. I can address both of these requests but it will be a little over a week before I have a chance to take a look.

Issue #1466, opened 2021-06-28
cwltool is not packing `Operation` definitions


jmfernandez:
I have been trying [cwltool 3.1.20210616134059](https://github.com/common-workflow-language/cwltool/releases/tag/3.1.20210616134059) with a very naïve, single `Operation` step workflow. Both validation and graph generation work, but when I packed it

```bash
cwltool --pack abstract-cosifer-workflow_fail1.cwl > abstract-cosifer-workflow_fail1.packed.cwl
```
packed file fails on validation

```bash
cwltool --validate abstract-cosifer-workflow_fail1.packed.cwl
```
```
INFO /home/jmfernandez/projects/iPC/ipc_workflows/cosifer/cwl_abstract/.p/bin/cwltool 3.1.20210616134059
INFO Resolved 'abstract-cosifer-workflow_fail1.packed.cwl' to 'file:///home/jmfernandez/projects/iPC/ipc_workflows/cosifer/cwl_abstract/a/abstract-cosifer-workflow_fail1.packed.cwl'
ERROR Tool definition failed initialization:
Tool definition file:///home/jmfernandez/projects/iPC/ipc_workflows/cosifer/cwl_abstract/a/abstract-cosifer-workflow_fail1.packed.cwl#abstract-cosifer.cwl failed validation:
  Reference `#abstract-cosifer.cwl` not found in file `file:///home/jmfernandez/projects/iPC/ipc_workflows/cosifer/cwl_abstract/a/abstract-cosifer-workflow_fail1.packed.cwl`.
```

because the `Operation` defintion was not included. I evolved that workflow, in order to have two parallel steps: a `CommandLineTool` and an `Operation`. And I got the very same result: the Operation definition was not included.

Having a look at cwltool code, I have found that it is skipping Operation definitions:

https://github.com/common-workflow-language/cwltool/blob/8ec1eb78e4c70638737656ecbcce02a4f0041c49/cwltool/pack.py#L254-L255

In order to ease reproducibility of this issue, I'm providing you a zip file with the workflows [operation_not_included.zip](https://github.com/common-workflow-language/cwltool/files/6725435/operation_not_included.zip)

mr-c:
Dear @jmfernandez ; thanks for your ideal issue report!

I don't think this was intentional. Can you send a PR to add `"Operation"` to this list, and a simplified test case? Thanks!

mr-c:
Fixed in https://github.com/common-workflow-language/cwltool/pull/1467

PR #1467, opened 2021-06-28
Allow packing Operation records

jmfernandez:
CWL 1.2 introduced [Operation](https://www.commonwl.org/v1.2/Workflow.html#Operation) in order to model abstract workflow steps, needed to use CWL to describe a workflow which could not exist or be implemented in a different way than CWL.

As `cwltool --pack` was ignoring records of class `Operation`, packed workflows were not correct (they did not validate with `cwltool --validate`). The issue was explained at https://github.com/common-workflow-language/cwltool/issues/1466.

These commit and test fix the issue which arises while packing workflows having `Operation` records, and its work.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -7,6 +7,7 @@ include tests/*
 include tests/tmp1/tmp2/tmp3/.gitkeep
 include tests/tmp4/alpha/*
 include tests/wf/*
+include tests/wf/operation/*
 include tests/override/*
 include tests/checker_wf/*
 include tests/subgraph/*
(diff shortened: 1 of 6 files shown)


tetron (approved):
operation-single.cwl needs to be added to MANIFEST.in, the rest looks ok

mr-c (approved):
Perfect, thanks! I'm going to make one small change and then I'll merge this. Afterwards I'll make a new release of `cwltool`.

jmfernandez:
> operation-single.cwl needs to be added to MANIFEST.in, the rest looks ok

I have just added the path pointing to the new data test directory to MANIFEST.in (see https://github.com/jmfernandez/cwltool/commit/55c0d2a5890a6e93a72613c1f6f0251fb1a6566c) and merged latest changes from main

PR #1469, opened 2021-07-01
Allow files to be named pipes

### LINKED ISSUE #1468: Request for cwltool to support named pipes
## Expected Behavior
Some runners such as Toil might implement file streaming with named pipes when `streamable` flag is true:
https://github.com/DataBiosphere/toil/issues/3469

## Actual Behavior
An exception is raised because a named pipe is not a regular file:
https://github.com/common-workflow-language/cwltool/blob/ed9dd4c3472e940a52dfe90049895f470bfd7329/cwltool/job.py#L180

## Workflow Code
```
#!/usr/bin cwl-runner

cwlVersion: v1.2
class: CommandLineTool
baseCommand: cat

stdout: $(inputs.output_filename)
arguments: ["-A"]

inputs:
  input_file:
    type: File
    streamable: true
    inputBinding:
      position: 1
  output_filename:
    type: string?
    default: out.txt

outputs:
  count:
    type: stdout

```

## Your Environment
* cwltool version: 
3.0.20201203173111 (Linux)

mhpopescu:
Fixes #1468

### DIFF
--- a/cwltool/context.py
+++ b/cwltool/context.py
@@ -107,6 +107,7 @@ class RuntimeContext(ContextBase):
         self.pull_image = True  # type: bool
         self.rm_container = True  # type: bool
         self.move_outputs = "move"  # type: str
+        self.streaming_allowed = False  # type: bool
 
         self.singularity = False  # type: bool
         self.debug = False  # type: bool
--- a/cwltool/context.py
+++ b/cwltool/context.py
@@ -107,7 +107,7 @@ class RuntimeContext(ContextBase):
         self.pull_image = True  # type: bool
         self.rm_container = True  # type: bool
         self.move_outputs = "move"  # type: str
-        self.streaming_allowed = False  # type: bool
+        self.streaming_allowed: bool = False
 
         self.singularity = False  # type: bool
         self.debug = False  # type: bool
--- a/cwltool/job.py
+++ b/cwltool/job.py
@@ -177,6 +177,8 @@ class JobBase(HasReqsHints, metaclass=ABCMeta):
             os.makedirs(self.outdir)
 
         def is_streamable(file: str) -> bool:
+            if not runtimeContext.streaming_allowed:
+                return False
             for inp in self.joborder.values():
                 if isinstance(inp, dict) and inp.get("location", None) == file:
                     return inp.get("streamable", False)
--- a/cwltool/job.py
+++ b/cwltool/job.py
@@ -176,16 +176,18 @@ class JobBase(HasReqsHints, metaclass=ABCMeta):
         if not os.path.exists(self.outdir):
             os.makedirs(self.outdir)
 
-        def is_streamable(file):
+        def is_streamable(file: str) -> bool:
             for inp in self.joborder.values():
-                if isinstance(inp, dict) and inp.get('location', None) == file:
-                    return inp.get('streamable', False)
+                if isinstance(inp, dict) and inp.get("location", None) == file:
+                    return inp.get("streamable", False)
             return False
 
         for knownfile in self.pathmapper.files():
             p = self.pathmapper.mapper(knownfile)
             if p.type == "File" and not os.path.isfile(p[0]) and p.staged:
-                if not (is_streamable(knownfile) and stat.S_ISFIFO(os.stat(p[0]).st_mode)):
+                if not (
+                    is_streamable(knownfile) and stat.S_ISFIFO(os.stat(p[0]).st_mode)
+                ):
                     raise WorkflowException(
                         "Input file %s (at %s) not found or is not a regular "
                         "file." % (knownfile, self.pathmapper.mapper(knownfile)[0])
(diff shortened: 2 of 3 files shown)


mr-c:
Can you add a unit test to prevent regressions?

We had also talked about only allowing named pipes in certain circumstances, so we need a way to only allow them when we are doing streaming, and forbid them otherwise

mhpopescu:
> We had also talked about only allowing named pipes in certain circumstances, so we need a way to only allow them when we are doing streaming, and forbid them otherwise

For this check I see the following options:
1) `type` is currently one of the following: `{File, Directory, CreateFile, WritableFile, CreateWritableFile}` and it could be extended with something like `StreamingFile`
2) `MapperEnt` adds a `streamable` attribute
3) have a convention that says that p[3] is the `streamable` flag which may or may not be present

mr-c:
> > We had also talked about only allowing named pipes in certain circumstances, so we need a way to only allow them when we are doing streaming, and forbid them otherwise
> 
> For this check I see the following options:
> 
>     1. `type` is currently one of the following: `{File, Directory, CreateFile, WritableFile, CreateWritableFile}` and it could be extended with something like `StreamingFile`
> 
>     2. `MapperEnt` adds a `streamable` attribute
> 
>     3. have a convention that says that p[3] is the `streamable` flag which may or may not be present
> 
>     4. search in `self.joborder` for the file and get the `streamable` attribute from there

@tetron any preferences here?

mr-c:
> 4) search in `self.joborder` for the file and get the `streamable` attribute from there

How terrible would this be to implement? Is self.joborder available where you made this change in this PR? I guess it could be a bit slow to search for this information if there are many inputs...

mr-c (approved):
LGTM; make the checks happy (which you can see locally by running `tox`) and I'll merge this (or I can wait until we get more testing with Toil)

cwltool/job.py
```diff
@@ -175,13 +176,22 @@ def _setup(self, runtimeContext: RuntimeContext) -> None:
         if not os.path.exists(self.outdir):
             os.makedirs(self.outdir)
 
+        def is_streamable(file: str) -> bool:
```
mr-c, line 179: `make diff_pydocstyle_report`?

mr-c (changes requested):
Needs a test!

mr-c (changes requested):
Also, please add a "streaming_allowed" boolean flag to the RuntimeContext and then adjust the toil-cwl-runner PR to set that flag to `True` only when we are in a streaming compatible context.

tests/test_streaming.py
```diff
@@ -0,0 +1,108 @@
...
+    path = tmp_path / "tmp"
+    os.mkfifo(path)
+
+    joborder = {
+        "inp": {
+            "class": "File",
+            "location": str(path),
+            "streamable": streamable,
+        }
+    }  # type: CWLObjectType
```
mr-c, line 99: Since this is a Python 3.6+ project we can use more concise type hint syntax:

```suggestion
    joborder: CWLObjectType = {
        "inp": {
            "class": "File",
            "location": str(path),
            "streamable": streamable,
        }
    }
```

cwltool/context.py
```diff
@@ -107,6 +107,7 @@ def __init__(self, kwargs: Optional[Dict[str, Any]] = None) -> None:
         self.pull_image = True  # type: bool
         self.rm_container = True  # type: bool
         self.move_outputs = "move"  # type: str
+        self.streaming_allowed = False  # type: bool
```
mr-c, line 110: Please use the new style everywhere :-) (Sorry to be annoying about this!)

cwltool/job.py
```diff
@@ -175,13 +176,24 @@ def _setup(self, runtimeContext: RuntimeContext) -> None:
...
+            if not runtimeContext.streaming_allowed:
+                return False
+            for inp in self.joborder.values():
+                if isinstance(inp, dict) and inp.get("location", None) == file:
+                    return inp.get("streamable", False)
+            return False
```
mr-c, line 185: This line needs testing

mr-c (changes requested):

cwltool/job.py
mhpopescu: I think this is dead code. Maybe I should add an "It should not get here" comment? I no not know an use-case when the file is not in self.joborder and that line would be reached. Another way of writing this function is to remove the first 2 lines and instead of `False` return `runtimeContext.streaming_allowed`. But doing this would unnecessarily iterate through all the files even when streaming is not allowed.

mr-c: A `File` from an input `Directory` could be staged in using `InitialWorkDirRequirement` and it itself would not appear in the top level of the joborder, and so would (incorrectly) reach this line. Maybe throw an exception here and run all the conformance tests and regular on a version of the code with `runtimeContext.streaming_allowed` set to `True` to see if there are other scenarios you are missing.

mhpopescu: If it returns False an exception is thrown here:
https://github.com/common-workflow-language/cwltool/blob/5c4464ea4670e30fcd1f22e46b6dc98ab17d2fe1/cwltool/job.py#L193

mr-c: Okay, then temporarily set `runtimeContext.streaming_allowed` to always be `True` and run the pytest and conformance tests again

mr-c (approved):

PR #1471, opened 2021-07-09
Pin networkx so we don't install numpy, scipy and pandas by default.

tetron:
The dependency chain is

cwltool -> prov -> networkx -> numpy

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,7 @@ rdflib>=4.2.2,<5.1
 shellescape>=3.4.1,<3.9
 schema-salad>=8.1.20210627200047,<9
 prov==1.5.1
+networkx==2.5.1
 bagit==1.8.1
 mypy-extensions
 psutil>=5.6.6
--- a/.snyk
+++ b/.snyk
@@ -1,0 +1,10 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.19.0
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore:
+  SNYK-PYTHON-NETWORKX-1062709:
+    - '*':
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -4,6 +4,7 @@ rdflib>=4.2.2,<5.1
 shellescape>=3.4.1,<3.9
 schema-salad>=8.1.20210627200047,<9
 prov==1.5.1
+networkx==2.5.1
```
mr-c, line 7: From  https://app.snyk.io/org/mr-c/project/0ffa07f0-c57e-4587-863c-6ce25cf63660/pr-check/d387790b-91bb-47b6-819b-a96d8feab7c0

> Affected versions of this package are vulnerable to Deserialization of Untrusted Data. This package is vulnerable to arbitrary code execution via insecure YAML deserialization due to the use of a known vulnerable function load() in yaml, which is called in read_yaml() in networkx/readwrite/nx_yaml.py.

> networkx/readwrite/nx_yaml.py is deprecated and scheduled to be removed in the next release of Networkx. Users should avoid using this function completely.

> Fixed in: networkx@2.6

mr-c (changes requested):

requirements.txt
mr-c: Does `prov` use this code path? If not, then we can ignore the warning

mr-c: Seems like we are safe: https://github.com/trungdong/prov/search?q=nx_yaml

mr-c (approved):

requirements.txt
tetron: :man_facepalming: 

I have no idea what prov does.  I did file an issue with networkx https://github.com/networkx/networkx/issues/4968

Issue #1364, opened 2020-11-10
Workflow with InitialWorkdirRequirement causes output to be symlinked from /tmp


hcraT:
## Expected Behavior

The workflow posted below produces the following tree structure:

```bash
out_dir/
  step1.txt
```

`step1.txt` is a normal file.

## Actual Behavior

`step1.txt` is a symbolic link to the `/tmp` folder where results are computed.

## Workflow Code

```yaml
cwlVersion: v1.1
class: Workflow

requirements:
  SubworkflowFeatureRequirement: {}
  InlineJavascriptRequirement: {}
  StepInputExpressionRequirement: {}

inputs: []

outputs:
  mmm:
    type: Directory
    outputSource: merge_output/output

steps:
  step1:
    in: []
    out: [output]
    run:
      class: CommandLineTool
      baseCommand: [echo]
      stdout: "step1.txt"
      inputs: []
      outputs:
        output:
          type: stdout
      arguments:
        - valueFrom: "step1"
  merge_output:
    in:
      in1: step1/output
    out: [output]
    run:
      requirements:
        InitialWorkDirRequirement:
          listing: |
            ${
              return [
                {
                  class: "Directory",
                  basename: "out_dir",
                  listing: [ inputs.in1 ]
                }
              ];
            }
      class: CommandLineTool
      baseCommand: ["true"]
      inputs:
        in1:
          type: File
      outputs:
        output:
          type: Directory
          outputBinding:
            glob: "out_dir"
```

If I change the merge_output step to:

```yaml
  merge_output:
    in:
      in1: step1/output
    out: [output]
    run:
      class: ExpressionTool
      inputs:
        in1: File
      outputs:
        output: Directory
      expression: |
        ${
          return {
            output: {
              class: "Directory",
              basename: "out_dir",
              listing: [ inputs.in1 ]
            }
          };
        }
```

the output is correct.

## Full Traceback

No exception happens but I thought that some useful information
could be part of the log.

```pytb
INFO /shares/CIBIO-Storage/CO/scratch/sharedCO/tools/executables/cwltool/current/installed/bin/__cwltool_orig__ 3.0
INFO Resolved 'bug_report.cwl' to 'file:///shares/CIBIO-Storage/CO/elaborazioni/sharedCO/dev/pipeline/cwl_my_examples/bug_report.cwl'
DEBUG Parsed job order from command line: {
    "id": "bug_report.cwl"
}
[...]

PR #1478, opened 2021-08-05
Check python3

tetron:
"make release-test" tweaks.

Check that "python" is actually python 3.  Also add "pip install -e ." before running tests on HEAD.


mr-c (approved):

tetron:
I really don't understand why these tests are failing.

tetron:
Seems like this might be detecting a legitimate bug, but I don't yet know why the test passes in some environments and fails in other environments.

tetron:
So this is definitely failing due to an underlying bug which is probably the same one described in https://github.com/common-workflow-language/cwltool/issues/1364

The problem is when moving a directory containing a symlink to another file in the temp directory, it doesn't update the symlink to point to the new location.

I don't understand why it is happening on this branch when (a) the changes are so trivial (b) I can't reproduce it except by changing the code to skip a step which mitigates the crash (but not the underlying error).

I either way, I will extend the test to check for expected behavior more thoroughly, and then fix the underlying bug.

Issue #1480, opened 2021-08-13
Setting `writable` of an `InitialWorkDirRequirement` entry does not make the root dir writable


jdidion:
I am running the [`recursive-input-directory.cwl`](https://github.com/common-workflow-language/cwl-v1.2/blob/main/tests/recursive-input-directory.cwl) conformance test, which specifies an `InitialWorkDirRequirement` with a writable entry containing all the contents of the input directory.

I have placed the input files in a read-only directory `inputdir` (file permissions `0500`). So my basedir looks like:

```
basedir
|-- inputdir
|   |-- a
|   |-- b
|   `-- c
|      `-- d
|-- recursive-input-directory.cwl
`-- recursive-input-directory_input.json
```

 I run:

`cwltool --basedir . --move-outputs recursive-input-directory.cwl recursive-input-directory_input.json`

where the input file is

```
{
  "input_dir": {
    "class": "Directory",
    "location": "inputdir"
  }
}
```

## Expected Behavior

cwltool creates a new writable directory (`work_dir`), copies all the files from `inputdir` into `work_dir`, and successfully runs the tool.

## Actual Behavior

cwltool fails with a permission error. On further investigation, I find that `work_dir` is being created by the `shutil.copytree` call [here](https://github.com/common-workflow-language/cwltool/blob/23acadf4b5553ca28d24b77e1116df28d99fc052/cwltool/process.py#L306). `copytree` copies all the permissions of the source directory, so `work_dir` and all of the files/directories under it initially have `500` permissions. Then, `ensure_writable` adds write permission to all of the files/directories under `work_dir`, but it does *not* set `work_dir` to writable.

I am happy to submit a PR to fix this, but I need guidance. One solution would be to modify `ensure_writable` to always set the root directory to be writable. Another solution would be to treat this as a special case and add write permission to `entry.target`.

## Full Traceback

```pytb
Traceback (most recent call last):
  File "/Users/jdidion/.pyenv/versions/3.6.14/envs/dxc/lib/python3.6/site-packages/cwltool/main.py", line 1249, in main
    tool, initialized_job_order_object, runtimeContext, logger=_logger
  File "/Users/jdidion/.pyenv/versions/3.6.14/envs/dxc/lib/python3.6/site-packages/cwltool/executors.py", line 59, in __call__
    return self.execute(process, job_order_object, runtime_context, logger)
[...]

PR #1481, opened 2021-08-13
bug/1480: make writable workdir actually writable

jdidion:
Fix for #1480 

* add include_root option to set root dir writable in ensure_writable
* set include_root=True when making workdir writable


mr-c (changes requested):
Thanks for the diagnosis and fix! This will need a test case where the inputs are marked read-only.

mr-c (approved):

PR #1485, opened 2021-08-18
build and publish containers to quay.io

mr-c:


tom-tan:
Great!
It would be nice if README mentions the new container image.

PR #1506, opened 2021-09-04
pack: add $namespaces to the (effective) root only

mr-c:
As per schema-salad, `$namespaces` are only valid in the root document:

https://www.commonwl.org/v1.0/SchemaSalad.html#Explicit_context
https://www.commonwl.org/v1.1/SchemaSalad.html#Explicit_context (same as v1.0)
https://www.commonwl.org/v1.2/SchemaSalad.html#Explicit_context (same as v1.0, v1.1)

Inspired by (and partial fix for) https://github.com/common-workflow-language/cwl-utils/issues/71#issuecomment-912887777

### DIFF
--- a/cwltool/pack.py
+++ b/cwltool/pack.py
@@ -268,6 +268,8 @@ def pack(
 
     if schemas:
         packed["$schemas"] = list(schemas)
+    if namespaces:
+        packed["$namespaces"] = namespaces
 
     for r in list(rewrite.keys()):
         v = rewrite[r]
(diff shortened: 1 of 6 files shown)


simleo:
If anyone has an old packed workflow that you can't read with cwl-utils due to this, here's a workaround:

```python
import json
from pathlib import Path
from cwl_utils.parser import load_document_by_yaml

wf_path = Path("packed.cwl")
with open(wf_path, "rt") as f:
    json_wf = json.load(f)
graph = json_wf.get("$graph", [json_wf])  # support single tool
for n in graph:
    ns = n.pop("$namespaces", {})
    if ns:
        json_wf.setdefault("$namespaces", {}).update(ns)
doc = load_document_by_yaml(json_wf, wf_path.absolute().as_uri())
```

PR #1529, opened 2021-09-22
trust given secondaryFiles if !discover_secondaryFiles

mr-c:
This is needed for toil-cwl-runner (and others) as some secondaryFile
expressions can modify other entries in a way that would cause
them to not pass validatation a second time

Example conformance test:
command_input_file_expression: Confirm CommandInputParameter expression can receive a File object


mr-c:
Why now? The last release tightened the processing secondaryFiles (and other expressions) in https://github.com/common-workflow-language/cwltool/commit/b67a5e0fc6b1e0dbd2277a112012bbe19a9dcc91

PR #1541, opened 2021-10-03
Remove temporary directories when cache dir is empty string

kinow:
For issue reported on Discourse: https://cwl.discourse.group/t/cwl-runner-tmpdir-not-deleted/401

When debugging the example from https://www.commonwl.org/user_guide/02-1st-example/index.html, with the command line `cwltool --tmpdir-prefix /tmp/cwl/ /tmp/1st-tool.cwl /tmp/echo-job.yml`, I noticed the `output_dirs` being set to an empty list.

The issue was because the `runtime_context.cachedir` was not `None`, but it was an empty string. So the value of `output_dirs` was incorrectly initialized.

Pending a unit/functional test as I have to learn how to write those for `cwltool` :grimacing: any pointers?

image

Thanks

### DIFF
--- a/cwltool/executors.py
+++ b/cwltool/executors.py
@@ -158,7 +158,7 @@ class JobExecutor(metaclass=ABCMeta):
             )
 
         if runtime_context.rm_tmpdir:
-            if runtime_context.cachedir is None:
+            if not runtime_context.cachedir:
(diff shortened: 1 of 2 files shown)


kinow:
I think I managed to add a test to `test_tmpdir.py`. Fails with the code from `main`:

image

And passes on this branch.

PR #1560, opened 2021-11-03
Call fsaccess.realpath before comparing paths in _relocate.

tetron:


mr-c (changes requested):
Cool. Can we get a test-case demonstrating the problem and that it is fixed?

tetron:
It took me way too long to write a test because the original case was pretty weird, it involves having the output directory also be part of the inputs but also symlinked.

mr-c (approved):

PR #1566, opened 2021-11-11
WIP fix #1330

### LINKED ISSUE #1330: workflow level ResourceRequirements unable to access workflow level inputs
``` cwl
#!/usr/bin/env cwl-runner
cwlVersion: v1.0
class: Workflow

requirements:
  InlineJavascriptRequirement: {}  # shouldn't need this, but makes the error more interesting

hints:
  ResourceRequirement:
    coresMax: $(inputs.threads_max)

inputs:
  threads_max:
    type: int
    default: 4

steps:
  one:
    in: []
    run:
      class: CommandLineTool
      inputs: []
      baseCommand: echo
      arguments: [ $(runtime.cores) ]
      outputs: []
    out: []

outputs: []
```

```
$ cwltool workflow-dyn-res.cwl
INFO /home/michael/cwltool/env3.8/bin/cwltool 3.0.20200724083110
INFO Resolved 'workflow-dyn-res.cwl' to 'file:///home/michael/cwltool/workflow-dyn-res.cwl'
INFO [workflow ] start
INFO [workflow ] starting step one
INFO [step one] start
ERROR Expecting value: line 1 column 1 (char 0)
script was:
01 "use strict";
02 var inputs = {};
03 var self = null;
04 var runtime = {
05     "tmpdir": "/tmp/tmpcb9aom_8",
06     "outdir": "/tmp/nx2jmuji"
07 };
08 (function(){return ((inputs.threads_max));})()
stdout was: 'undefined'
stderr was: ''

  File "/home/michael/cwltool/cwltool/sandboxjs.py", line 411, in execjs
    return cast(CWLOutputType, json.loads(stdout))
  File "/usr/lib/python3.8/json/__init__.py", line 357, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.8/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.8/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

The above exception was the direct cause of the following exception:

  File "cwltool/expression.py", line 377, in do_eval
    return interpolate(
  File "cwltool/expression.py", line 293, in interpolate
    e = evaluator(
  File "cwltool/expression.py", line 250, in evaluator
    return execjs(
  File "/home/michael/cwltool/cwltool/sandboxjs.py", line 413, in execjs
    raise JavascriptException(
cwltool.sandboxjs.JavascriptException: Expecting value: line 1 column 1 (char 0)
script was:
01 "use strict";
02 var inputs = {};
03 var self = null;
04 var runtime = {
05     "tmpdir": "/tmp/tmpcb9aom_8",
06     "outdir": "/tmp/nx2jmuji"
07 };
08 (function(){return ((inputs.threads_max));})()
stdout was: 'undefined'
stderr was: ''

ERROR Exception on step 'one'
ERROR [step one] Cannot make job: Expression evaluation error:
Expecting value: line 1 column 1 (char 0)
script was:
01 "use strict";
02 var inputs = {};
03 var self = null;
04 var runtime = {
05     "tmpdir": "/tmp/tmpcb9aom_8",
06     "outdir": "/tmp/nx2jmuji"
07 };
08 (function(){return ((inputs.threads_max));})()
stdout was: 'undefined'
stderr was: ''

INFO [workflow ] completed permanentFail
{}
```

manabuishii:
#1330

At least workflow-dyn-res.cwl does not create error.

But this PR fail some tests.

```
(venv) ➜  cwltool git:(fix_issue_1330) ✗ python3 -m cwltool workflow-dyn-res.cwl         
INFO /Users/manabu/ghq/github.com/common-workflow-language/cwltool/cwltool/__main__.py 3.1.20211107152837
INFO Resolved 'workflow-dyn-res.cwl' to 'file:///Users/manabu/ghq/github.com/common-workflow-language/cwltool/workflow-dyn-res.cwl'
INFO [workflow ] start
INFO [workflow ] starting step one
INFO [step one] start
INFO [job one] /private/tmp/docker_tmpduee6x2k$ echo \
    4
4
INFO [job one] completed success
INFO [step one] completed success
INFO [workflow ] completed success
{
    "file:///Users/manabu/ghq/github.com/common-workflow-language/cwltool/workflow-dyn-res.cwl#threads_max": 4
}
INFO Final process status is success
```

### DIFF
--- a/cwltool/process.py
+++ b/cwltool/process.py
@@ -775,6 +775,10 @@ class Process(HasReqsHints, metaclass=abc.ABCMeta):
             )
 
         job = copy.deepcopy(joborder)
+        if len(job.keys()) > 0:
+            for jo in joborder:
+                field = shortname(jo)
+                job[field] = joborder[jo]
 
         make_fs_access = getdefault(runtime_context.make_fs_access, StdFsAccess)
         fs_access = make_fs_access(runtime_context.basedir)
--- a/cwltool/workflow_job.py
+++ b/cwltool/workflow_job.py
@@ -386,6 +386,8 @@ def object_from_state(
     incomplete: bool = False,
 ) -> Optional[CWLObjectType]:
     inputobj = {}  # type: CWLObjectType
+    for s in state.keys():
+        inputobj[s] = state[s].value
     for inp in parms:
         iid = original_id = cast(str, inp["id"])
         if frag_only:
--- a/cwltool/workflow.py
+++ b/cwltool/workflow.py
@@ -440,10 +440,11 @@ class WorkflowStep(Process):
             )
 
         step_input = {}
-        for inp in self.tool["inputs"]:
-            field = shortname(inp["id"])
-            if not inp.get("not_connected"):
-                step_input[field] = job_order[inp["id"]]
+        for inp in job_order:
+            field = shortname(inp)
+            step_input[field] = job_order[inp]
+            # if not inp.get("not_connected"):
+            #    step_input[field] = job_order[field]
 
         try:
             yield from self.embedded_tool.job(


cwltool/workflow_job.py
```diff
@@ -386,6 +386,8 @@ def object_from_state(
     incomplete: bool = False,
 ) -> Optional[CWLObjectType]:
     inputobj = {}  # type: CWLObjectType
+    for s in state.keys():
+        inputobj[s] = state[s].value
```
kinow, line 390: I think we can avoid accessing keys & values separately. Something like:

```python
for k, v in state.items():
  inputobj[k] = v
```

Or start `inputobj` by deep-cloning `state`.

kinow (commented):
@manabuishii impressive work tracking down where you had to add the state/job info :clap: 

I think what this change is doing is making the job info from the parent Job available to children Jobs. Not sure if there's any downsides to this.

**Maybe** another approach could be to modify `Process#evalResources` and rewrite the Resource value, evaluating the expression to `4`. This way when children Job are created they wouldn't have to try to resolve the variables/values.

cwltool/process.py
```diff
@@ -775,6 +775,10 @@ def _init_job(
             )
 
         job = copy.deepcopy(joborder)
+        if len(job.keys()) > 0:
+            for jo in joborder:
+                field = shortname(jo)
+                job[field] = joborder[jo]
```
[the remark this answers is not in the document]
kinow: Or maybe it's used later by another child job? Maybe better to leave this one?

kinow:
>Maybe another approach could be to modify Process#evalResources and rewrite the Resource value, evaluating the expression to 4. This way when children Job are created they wouldn't have to try to resolve the variables/values.

Hmm, interesting. Tried that, but realized that `process.py` intentionally skips evaluating the resources for the top level job (the Workflow itself is the parent job). :man_shrugging: there must be a good reason for doing so, so I think your approach should be the right fix @manabuishii :+1: let's see what's wrong with the build...

mr-c:
@manabuishii Can you fix the merge conflict so that the CI can run?

cwltool/process.py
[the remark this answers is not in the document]
manabuishii: @kinow Thanks, comment.
I tested without this `cwltool/process.py` , there is no problem.
So I removed it from PullRequest.

cwltool/workflow_job.py
manabuishii: Current commited version is 
https://github.com/common-workflow-language/cwltool/pull/1566/files#diff-ed87d61e2237460638e7f3ddb21bafc9969d8c142c804d79f94801142342f3d6R389-R391

My local version is following .
But I can not pass `py39-mypy`

```python
    for s in state.keys():
        if hasattr(state[s], "value"):
            inputobj[s] = state[s].value
```

I replace it to your code.

```
    for k,v  in state.items():
        if type(v) is WorkflowStateItem:
            if hasattr(v, "value"):
                inputobj[k] = v.value
```

cwltool/workflow_job.py
```diff
@@ -551,6 +554,23 @@ def do_output_callback(self, final_output_callback: OutputCallbackType) -> None:
...
+                        paramskey.append(v)
+                        paramskeys.append(shortname(v))
+
+        newwo = {}
+        for k, v in wo.items():
+            if k in paramskey:
+                newwo[k] = v
+            elif k in paramskeys:
+                newwo[k] = v
+        wo = newwo
```
mr-c, line 573:
```suggestion
        if wo is not None:
            new_wo = {}
            for k, v in wo.items():
                if k in paramskey:
                    new_wo[k] = v
                elif k in paramskeys:
                    new_wo[k] = v
            wo = new_wo
```

kinow:
Getting closer to a green build! :rocket:

manabuishii:
@kinow @mr-c 

This commit is done by me, but account is not current my account .

https://github.com/common-workflow-language/cwltool/pull/1566/commits/a7a66cf340fd221a3a7e591b39e0647c6229e791

Is there any good way to rewrite my commit ?
Or Can I leave it as it is?

kinow:
@manabuishii the commit will be squashed when we merge the PR (it has 6 commits, ideally a single commit is best, or grouped commits if they are helpful for future devs.)

Or if you prefer to edit the commit now and change the author, you can use interactive rebase or amend (if the last commit), e.g. https://stackoverflow.com/a/3042512

manabuishii:
@kinow Thanks !!
I choose to squash when sending pull requests .

manabuishii:
@kinow @mr-c 
How do I execute following test ( that fails GitHub Actions)

```text
Continuous integration tests / CWL spec conformance tests (v1.0, podman) (pull_request) 
```

Is this correct ?

```text
source venv/bin/activate
container=podman spec_branch=main version=v1.0 ./conformance-test.sh
```

mr-c:
You can run just the test that failed

https://github.com/common-workflow-language/cwltool/runs/5765165834?check_suite_focus=true#step:5:490

`cwltool --parallel --podman --debug https://github.com/common-workflow-language/common-workflow-language/raw/main/v1.0/v1.0/fail-unconnected.cwl`

Note that this test should fail: https://github.com/common-workflow-language/common-workflow-language/blob/a62f91e3d75036e7d89dbd454fc4e1f72caf901f/v1.0/conformance_test_v1.0.yaml#L1863

manabuishii:
My current strategy is such that I make everything a global variable, so
It appears that the scope of the variable is not properly controlled.

Therefore, the

`workflow-dyn-res.cwl` also does not deal with `input.cores`.
It is looking at `inputs.threads_max`, which it was forced to look at as a global variable.

Essentialy , `inputs.threads_max` is first set to `coresMax`
And the `coresMax` is reflected in `runtime.cores`

I think `runtime.cores` is necessary to put it in `arguments` of `echo`.

So this is the reason why `cwltool --parallel --podman --debug https://github.com/common-workflow-language/common-workflow-language/raw/main/v1.0/v1.0/fail-unconnected.cwl` is not failed, I think.

I need to change strategy.

manabuishii:
Memo:
`threads_max` is required at `evalResources` in `process.py` to make `coresMax` and `runtime.core`
After using `threads_max` in `evalResources` of `process.py`, I test  builder.job(in command_line_tool.py) which is copy of job_order , workflow-dyn-res.cwl and fail-unconnected.cwl is behaved expected.
Of course other unit tests are broken.

manabuishii:
@mr-c I think this PullRequest fix #1330.
All tests are passed.

Now I want to check that there is no unwanted behavior .

I want to ask questions related to `fail-unconnected.cwl`

Question 1: What kind of output is expected for `ExpressionTool` instead of `CommandLineTool` ?

`fail-unconnected-expressiontool.cwl` is following

```cwl
class: Workflow
cwlVersion: v1.1
inputs:
  inp1:
    type: string
    default: hello inp1
  inp2:
    type: string
    default: hello inp2
outputs:
  out:
    type: string
    outputSource: step1/out
steps:
  step1:
    in:
      in: inp1
      in2: inp2
    out: [out]
    run: fail-unspecified-input-expressiontool.cwl
```

and

`fail-unspecified-input-expressiontool.cwl` is following

```cwl
cwlVersion: v1.0
class: ExpressionTool

requirements:
  InlineJavascriptRequirement: {} 

inputs:
  in:
    type: string

outputs:
  out:
    type: string

expression: |
  ${ 
    return {"out": inputs.in +" "+inputs.in2};
  }
```

Both
 current HEAD (7a058fed2987bc0046ffbc9a2b8a39b5b4fec6e1)
and this PullRequest return same

Is this result correct ?

```console
python3 -m cwltool fail-unconnected-expressiontool.cwl
INFO /home/manabu/ghq/github.com/common-workflow-language/cwltool/cwltool/__main__.py 3.1.20220325085616
INFO Resolved 'fail-unconnected-expressiontool.cwl' to 'file:///home/manabu/ghq/github.com/common-workflow-language/cwltool/fail-unconnected-expressiontool.cwl'
WARNING Workflow checker warning:
fail-unconnected-expressiontool.cwl:18:7: 'in2' is not an input parameter of
                                          file:///home/manabu/ghq/github.com/common-workflow-language/cwltool/fail-unspecified-input-expressiontool.cwl,
                                          expected in
INFO [workflow ] start
INFO [workflow ] starting step step1
INFO [step step1] start
INFO [step step1] completed success
INFO [workflow ] completed success
{
    "out": "hello inp1 undefined"
}
INFO Final process status is success
```

As WARNING messages indicates , `fail-unconnected-expressiontool.cwl` does not have `in2` as input parameter.
Then result indiates `inputs.in2` is evaluated to `undefined`.

When `CommandLineTool` case is error, but `ExpressionTool` case is warning .
So I want to ask correct behavior.

Question2: What kind of output is expected for `Operation` instead of `CommandLineTool` ?
I think CWL version 1.2 has `Operation` .
But I do not know how to write it for tests.

tom-tan:
I guess it must fail.

The spec of [WorkflowStep.in](https://www.commonwl.org/v1.2/Workflow.html#WorkflowStep) says:
> Input parameters include a schema for each parameter which is used to validate the input object.

But `in2` has no corresponding schema to be used for validation.

cwltool/schemas/v1.1/conformance_tests.yaml
```diff
@@ -2598,6 +2598,17 @@
...
+- job: tests/empty.json
+  tool: tests/fail-unconnected-expressiontool.cwl
+  label: wf_step_access_undeclared_param_expressiontool
+  id: 198
+  doc: >-
+    Test that parameters that don't appear in the `run` process
+    inputs are not present in the input object used to run the expressiontool.
+  should_fail: true
+  tags: [ required, workflow ]
+
```
mr-c, line 2611: The CI doesn't run this copy of the tests, it makes a new checkout from GitHub https://github.com/common-workflow-language/cwltool/blob/7a058fed2987bc0046ffbc9a2b8a39b5b4fec6e1/conformance-test.sh#L41

manabuishii: @mr-c Thanks

I will create a PullRequest about this test case, 

Repo:
https://github.com/common-workflow-language/cwl-v1.2
Branch:
`1.2.1_proposed`

mr-c: Thanks, you can also keep it in this PR, just move the test scripts to the the test directory and add an entry to test_examples.py

mr-c: Can you move this file to the `tests` directory, or a sub-directory?

manabuishii:
I improve `ExpressionTool` inputs handling.

This does not cause error.
It becomes `undefined`.

First, currently

`in` is resolved as `inp1`
`in2` is not resolved.

I think this is expected behavior for following `class ExpressionTool`

https://github.com/common-workflow-language/cwltool/pull/1566/files#diff-3d8236842572888947227e8925a930cebe33d4af9c6bb8abc61282eec6d105c1R17

Just before `exec_js_process`

fn is following.
`inputs` member has only one `in`.

```javascript
'"use strict";
var inputs = {
    "in": "hello inp1"
};
var self = null;
var runtime = {
    "cores": 1,
    "ram": 1024,
    "tmpdirSize": 1024,
    "outdirSize": 1024,
    "tmpdir": null,
    "outdir": null
};
(function(){ 
  return {"out": inputs.in +" "+inputs.in2};
})()'
```

function executed, 
`inputs.in` is `hello inp1`
`inputs.in2` is `undefined` because `inputs.in2` is undefined.

Output is here

```console
$ python3 -m cwltool tests/fail-unconnected-expressiontool.cwl 
INFO /Users/manabu/ghq/github.com/common-workflow-language/cwltool/cwltool/__main__.py 3.1.20211111031309
INFO Resolved 'tests/fail-unconnected-expressiontool.cwl' to 'file:///Users/manabu/ghq/github.com/common-workflow-language/cwltool/tests/fail-unconnected-expressiontool.cwl'
WARNING Workflow checker warning:
tests/fail-unconnected-expressiontool.cwl:18:7: 'in2' is not an input parameter of
                                                file:///Users/manabu/ghq/github.com/common-workflow-language/cwltool/tests/fail-unspecified-input-expressiontool.cwl,
                                                expected in
INFO [workflow ] start
INFO [workflow ] starting step step1
INFO [step step1] start
INFO [step step1] completed success
INFO [workflow ] completed success
{
    "out": "hello inp1 undefined"
}
INFO Final process status is success
```

I investigate related JS behavior.

When `inputs` is declared as dictionary.
`inputs.in2` returns `undefined`.
Just `in2` is cause error because `in2` is not defined.

```
> inputs={}
> inputs.in = "hello inp1"
> inputs.in2
undefined
> in2
VM1960:1 Uncaught ReferenceError: in2 is not defined
    at <anonymous>:1:1
```

I think  that `in2` is resolved as `undefined` and it is correct behavior at least some Javascript Implementation.

manabuishii:
previous version https://github.com/common-workflow-language/cwltool/pull/1566/commits/14a2dedf63b240c4bb3f455292e7fda861b0ee03
`in2` is resolved as `hello inp2`

so output is following.
At least this result is wrong.

```console
$ python3 -m cwltool tests/fail-unconnected-expressiontool.cwl
INFO /Users/manabu/ghq/github.com/common-workflow-language/cwltool/cwltool/__main__.py 3.1.20211111031309
INFO Resolved 'tests/fail-unconnected-expressiontool.cwl' to 'file:///Users/manabu/ghq/github.com/common-workflow-language/cwltool/tests/fail-unconnected-expressiontool.cwl'
WARNING Workflow checker warning:
tests/fail-unconnected-expressiontool.cwl:18:7: 'in2' is not an input parameter of
                                                file:///Users/manabu/ghq/github.com/common-workflow-language/cwltool/tests/fail-unspecified-input-expressiontool.cwl,
                                                expected in
INFO [workflow ] start
INFO [workflow ] starting step step1
INFO [step step1] start
INFO [step step1] completed success
INFO [workflow ] completed success
{
    "out": "hello inp1 hello inp2"
}
INFO Final process status is success
```

manabuishii:
I think current this PR approach is Lazy evaluation.
But implementation is difficult.
In my opinion, Eager evaluation will be more clear and easy to understand and easy to implement.

How do you think ?

kinow:
>In my opinion, Eager evaluation will be more clear and easy to understand and easy to implement.

I haven't dug really deep in the issue as @manabuishii , but from what I could see, I also think that eager evaluation in this case would be better for `cwltool`.

As for specification, I think we can leave it as-is, or maybe add a comment that eager or lazy evaluation is up to implementers, but not enforced by the specification.

manabuishii:
@kinow Thanks !!

> As for specification, I think we can leave it as-is, or maybe add a comment that eager or lazy evaluation is up to implementers, but not enforced by the specification.

I consider about multi cloud situation (including on-premise cloud and some on-premise cluster) sometimes lazy evaluations require variable or file which is not accessible  from some environment.
Maybe it is related to acquire for the machine resources to distributed multi computing environments (multi-cloud).

To consider such a situation , I think we keep discuss and investigate about pros, cons and reasonable assumption.
But during the discussion I agree that "we can leave it as-is".
Mainly I care about the difference between implementations if the specification is said nothing.(Sorry not so much deeply about "as-is" but I feel it's okay because it is current state.)

One more option is that spec said nothing but implementation is accept variable about eager or lazy evaluation.

But I want to fix #1330 😄

tom-tan:
I agree with @kinow that we leave the spec as is (at least in v1.2.0) because clarifying the evaluation strategy may break existing platforms.
IMO it is better to fix it in the later release of CWL by clarifying the default evaluation strategy and by extending the spec to support use cases for eager and lazy strategies.

---
I prefer the eager strategy by default because it makes much easier to implement platforms.
I proposed a tentative extension to cover the use cases for lazy strategy in [another issue](https://github.com/common-workflow-language/cwl-v1.2/issues/157#issuecomment-1144424810).

PR #1567, opened 2021-11-12
WSL2 clarification

mr-c:


kinow (commented):
Clicked on "View File" here on GitHub, and the rendered output looks a bit odd, but maybe because it's in the PR?

image

Text looks good though!

ALuesink:
Similar to @kinow, I'm not sure if the image will show or if it remains a link. Otherwise it looks good.

PR #1577, opened 2021-12-21
_terminate_processes for non-docker processes, like nodejs

mr-c:
Fix for the `cwltool` side of https://github.com/DataBiosphere/toil/issues/3965

### DIFF
--- a/cwltool/main.py
+++ b/cwltool/main.py
@@ -120,7 +120,7 @@ def _terminate_processes() -> None:
         else:
             args = [process.args]
         cidfile = [str(arg).split("=")[1] for arg in args if "--cidfile" in str(arg)]
-        if cidfile:
+        if cidfile:  # Try to be nice
             try:
                 with open(cidfile[0]) as inp_stream:
                     p = subprocess.Popen(  # nosec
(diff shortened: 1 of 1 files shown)


mr-c:
@EtiennePer and @w-gao Let me know if this helps

cwltool/main.py
```diff
@@ -132,6 +132,10 @@ def _terminate_processes() -> None:
                         p.kill()
             except FileNotFoundError:
                 pass
+        if process.stdin:
+            process.stdin.close()
+        process.wait(10)
```
w-gao, line 137: My concern is that this wait may raise a [`TimeoutExpired`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait) exception if the process doesn't end within 10s for some reason, which would interrupt the rest of the cleanup.

w-gao (commented):
Thank you for fixing this, @mr-c! I tested this with `cwltool` and the nodejs processes are cleaned up now. The next step would be to actually call this function in Toil :).

cwltool/main.py
mr-c: Good point, I'll add a try/except

PR #1579, opened 2021-12-22
allow additional pathname characters in strict mode

mr-c:
specifically , : @ ] ^
(comma, colon, at-sign, end-square-bracket, and caret)

In response to https://github.com/common-workflow-language/cwl-v1.2/pull/144


cwltool/command_line_tool.py
```diff
@@ -89,14 +89,51 @@
...
+
+    # Note: the following characters are intentionally not included:
+    #
+    # reserved words in POSIX:
+    # ! { }
+    #
+    # POSIX metacharacters
+    # | & ; < > ( ) $ ` " ' <space> <tab> <newline>
+    # (In accordance with
+    # https://www.commonwl.org/v1.0/CommandLineTool.html#File under "path"
```
kinow, line 124: missing `)`?

kinow (commented):
The colon may give some trouble to users that either copy the files to Windows, or that run `cwltool` on Windows linux subsystem. I think in Windows `:` is reserved for file system drive names, and cannot (or could not last I checked) be used in file names, unless we escape it.

I had a quick test manually, and spaces are not allowed :heavy_check_mark: , commas accepted :heavy_check_mark: , colon accepted :heavy_check_mark: . Accents in words used in Portuguese (ã, é, etc) passed the regex too :heavy_check_mark: .Then tested with a couple Japanese characters :heavy_check_mark: , and found an issue with common symbols used in Japanese, but not sure if important.

The `\w` in the regex appears to match words in Japanese in any of its three writing systems (hiragana, katakana, kanji), not sure about other Chinese characters, but shouldn't be a problem :heavy_check_mark: 

Other Japanese [symbols](http://www.localizingjapan.com/blog/2012/01/20/regular-expressions-for-japanese-text/) that are valid in file names, and are not spaces, are not supported, such as the parentheses "「」" or the Japanese version for comma/dot "、。". 

These symbols can be used in normal file names in Windows/Linux, but I think we cannot guarantee every symbol of every language will be supported, so that should be fine I think.

Thanks! :neckbeard:

mr-c:
@kinow Thanks for the thoughtful review! I also tested using https://regex101.com/r/J8mA2g/1

> The colon may give some trouble to users that either copy the files to Windows, or that run `cwltool` on Windows linux subsystem. I think in Windows `:` is reserved for file system drive names, and cannot (or could not last I checked) be used in file names, unless we escape it.

Sure, but for `cwltool` we only support WSL 2 + Docker, so we only have to care about POSIX. Users can rename results before copying them to other systems if there is an issue.


> Other Japanese [symbols](http://www.localizingjapan.com/blog/2012/01/20/regular-expressions-for-japanese-text/) that are valid in file names, and are not spaces, are not supported, such as the parentheses "「」" or the Japanese version for comma/dot "、。".
> 
> These symbols can be used in normal file names in Windows/Linux, but I think we cannot guarantee every symbol of every language will be supported, so that should be fine I think.

Good to know. If we switch to https://pypi.org/project/regex/  then we can easily include all unicode characters classified as `symbols` and explicitly exclude `Separator` and `Other` characters

PR #1581, opened 2021-12-22
Initial support CUDARequirement extension

tetron:
Express that a CommandLineTool requires NVIDA CUDA GPU support.  Includes support for Docker and singularity.  Supports requesting multiple GPUs, but currently does not attempt to allocate GPUs among multiple concurrent processes.

Example command line tool:

```
cwlVersion: v1.2
class: CommandLineTool
baseCommand: nvidia-smi
inputs: []
outputs: []
$namespaces:
  cwltool: http://commonwl.org/cwltool#
hints:
  cwltool:CUDARequirement:
    cudaVersionMin: "11.4"
    cudaComputeCapabilityMin: "3.0"
    deviceCountMin: 1
    deviceCountMax: 8
  DockerRequirement:
    dockerPull: nvidia/cuda:11.4.0-base-ubuntu20.04
```


cwltool/cuda.py
```diff
@@ -0,0 +1,41 @@
+import subprocess  # nosec
+import xml.dom.minidom  # nosec
```
mr-c, line 2:
```suggestion
import defusedxml.dom.minidom
```
Please and thank you ☺️

tetron: I saw that in the documentation but didn't want to do it, it means adding another package dependency for questionable security benefit, this is already running an external program that could do lots of things that are more dangerous than returning a malicious XML file.

mr-c (changes requested):
Huzzah! Can we have some tests? I'll try to find an CUDA box someplace to run this

mr-c (approved):
The singularity version of the tests passed for me on the BAZIS cluster: https://libguides.vu.nl/rdm/high-performance-computing

tetron:
> Maybe add a comment that nvidia-smi returns non-zero if there is no GPU detected?

I'm not entirely sure what it does, but it seems reasonable to assume that either it'll exit non-zero, or it will return XML with  `<attached_gpus>0</attached_gpus>` and in either case it'll be detected as CUDA not available.

mr-c:
> > Maybe add a comment that nvidia-smi returns non-zero if there is no GPU detected?
> 
> I'm not entirely sure what it does, but it seems reasonable to assume that either it'll exit non-zero, or it will return XML with `<attached_gpus>0</attached_gpus>` and in either case it'll be detected as CUDA not available.

Yes. I was referring to the `nvidia-smi` invocation within the tests themselves: https://github.com/common-workflow-language/cwltool/pull/1581/files#diff-3dac8debea16928fd49923af94fa8e1af78788a1c44d86eb3b6112bbc7516993R11

tetron:
Oh, I see what you mean.  Yes, I'm trying to confirm what it does if it has the libraries but doesn't find any devices.

tetron:
It's annoying because it turns out `nvidia-smi` is one of the things that gets injected from the host system into the container, so the standard CUDA runtime container doesn't actually have it.

mr-c:
FYI, the proposal here was updated in https://github.com/common-workflow-language/cwltool/pull/1629

The differences being 

1. `cudaComputeCapabilityMin` is now `cudaComputeCapability`: a single value is a minimum capability, or if a list of values is specifed then only GPUs with compute capabilities in the list should be considered.
2. `deviceCountMin` was renamed `cudaDeviceCountMin` and can also be an expression in addition to an `int`. Likewise for `deviceCountMax` becoming `cudaDeviceCountMax`

PR #1587, opened 2022-01-06
remove an Any type hint

mr-c:


kinow (approved):
:tada: (CI still running, so approving assuming mypy is happy with this change)

PR #1589, opened 2022-01-10
GA: simplify pip caching

mr-c:


.mergify.yml
```diff
@@ -1,13 +1,19 @@
+queue_rules:
+  - name: default
+    conditions:
+      - -draft
```
kinow, line 4: Extra dash? I think, https://docs.mergify.com/conditions/?highlight=draft#attributes

kinow (approved):
pip caching is looking good. Mergify failed again

```


The configuration uses the deprecated strict mode of the merge action.
A brownout is planned for the whole December 6th, 2021 day.
This option will be removed on January 10th, 2022.
For more information: https://blog.mergify.com/strict-mode-deprecation/

smart+fasttrack is invalid for dictionary value @ pull_request_rules → item 0 → actions → merge → strict
```

Approving as PR is for pip caching :+1:

.mergify.yml
mr-c:
```suggestion
      - draft
```

PR #1599, opened 2022-01-20
Process generator docs

tetron:
Add page describing the ProcessGenerator extension.


mr-c (changes requested):
`docs/processgen.rst: WARNING: document isn't included in any toctree` :-)

tetron:
I didn't dig too far into trying to generate the docs, but it doesn't seem like we have a `make doc-dev` or something that installs sphinx and all the correct plugins.

Does it get auto-built somewhere?  Long term we should be migrating to a proper doc tree but right now it is very incomplete, including how to add to it.

mr-c:
> I didn't dig too far into trying to generate the docs, but it doesn't seem like we have a `make doc-dev` or something that installs sphinx and all the correct plugins.

Okay, I've added a Makefile target so that `make install-doc-dep docs` works

> Does it get auto-built somewhere?

There is a Read the Docs autobuilder enabled for each PR

The current build is https://cwltool--1599.org.readthedocs.build/en/1599/

Though it doesn't show the log unless it fails. To find the log I had to go to https://readthedocs.org/projects/cwltool/builds/ → https://readthedocs.org/projects/cwltool/builds/15866894/ → "view raw" https://readthedocs.org/api/v2/build/15866894.txt

>  Long term we should be migrating to a proper doc tree but right now it is very incomplete, including how to add to it.

`processgen` is the first piece of non-autogenerated doc :-) I've added it to the existing table of contents

mr-c:
https://cwltool--1599.org.readthedocs.build/en/1599/processgen.html

docs/processgen.rst
```diff
@@ -0,0 +1,230 @@
+=================
+Process generator
+=================
+
+Experimental feature.
```
mr-c, line 5:
```suggestion
Experimental feature and unofficial extension to the CWL standards.
```

mr-c (changes requested):
Otherwise, LGTM!

tetron:
Ah, this is good, actually.

To bootstrap proper documentation, it probably wouldn't be too hard to break up the current README into sections.  That's another ticket, though.

mr-c (approved):

PR #1600, opened 2022-01-21
Issue/real time logging

kannon92:
The goal of this PR is to enable logging of plugins while a workflow runs.  

1) Added a CLI option --log-dir 
2) If you specify log-dir, it will generate a unique id for stdout/stderr (Helps keep logs unique for all instances of plugin)
3) Logs will be written to the directory you specify.
4) In the CLT, you still need to specify stdout and/or stderr to actually log the plugins stdout and stderr, respectively.

Caveats:

If you want to capture logs as outputs, this will not work.


mr-c:
hmm.. this causes failures for me for any CWL document that expects the `stdout` or `stderr` to have a particular filename. Maybe the uuid should be for a subdirectory, and not the filename itself?

To test yourself, checkout https://github.com/common-workflow-language/cwl-v1.2/tree/v1.2.0 `pip install cwltest` and run `./run_test.sh RUNNER=cwltool EXTRA="--log-dir $PWD/logs"`

kannon92:
> hmm.. this causes failures for me for any CWL document that expects the `stdout` or `stderr` to have a particular filename. Maybe the uuid should be for a subdirectory, and not the filename itself?

Good call.  

One issue I find with specying stdout/stderr is that every call of the tool will write out the same file name.  If you include the same tool in a workflow multiple times, the log files will conflict with each other.  

I got around this but just including stdout/stderr as an output so the name is auto generated.

I'd like to keep the same behavior.  Generate a random name for the log file so every invocation of the tool produces a unqiue log.  Open to suggestions on this?

> To test yourself, checkout https://github.com/common-workflow-language/cwl-v1.2/tree/v1.2.0 `pip install cwltest` and run `./run_test.sh RUNNER=cwltool EXTRA="--log-dir $PWD/logs"`

I am having some trouble with this call.  If I remove EXTRA,  

```
./run_test.sh RUNNER=cwltool

importlib.metadata.PackageNotFoundError: cwltool

37 tests passed, 299 failures, 0 unsupported features

```

mr-c:
> I am having some trouble with this call. 

I presume you have a virtualenv active with both `cwltest` and your branch installed? `pip install -e path_to_branch`

kannon92:
> > I am having some trouble with this call.
> 
> I presume you have a virtualenv active with both `cwltest` and your branch installed? `pip install -e path_to_branch`

Thank you!  I was missing the `pip install -e path_to_branch`. Thought I could just activate my cwltool branch and install cwltest into that.  But that didn't seem to work.  

They are running now.

kannon92:
@mr-c I ran these tests and I realize that they are failing.  I notice that if you specify the log-dir parameter, any staging of the output fails because that code tries to find the output in the container.  I was thinking that if you specify the log-dir command, you should not have stdout/stderr captured as outputs.  

So I think that outputs for stdout/stderr is incompatible with saying you want to capture logs outside of the container.

kannon92:
> Hmm.. what is a user supposed to do if their workflow has some steps with `stdout` and other steps they want live logging from? (Or they want live logging from all steps)

Good point.  I realize now that having it as an cli command would mean that you wouldn't be able to capture stdout/stderr as outputs for any steps of your workflow.  ie that parameter would mean that outputs can not be used to capture stdout/stderr.  

For our cloud/hpc plugins that we are building, we realize that relying on stdout/stderr as outputs is difficult as stdout/stderr are typically expected to be logs that are streamed as the plugin runs.  So in my case, I was thinking that every plugin/workflow would have live logging turned on and the only thing that you would put in outputs is the actual output of the plugin, not logs.  

Selectively turning on logs means that we may want to add this on a step level basis?  Maybe this would be a future specification for workflows?

kannon92:
> > Selectively turning on logs means that we may want to add this on a step level basis?
> 
> I see a couple options:
> 
> 1. Change the code so that `--log-dir` only affects `CommandLineTool`s without `stdout`/`stderr`
> 
> or
> 
> 2. Fix the code to still work in the above cases
> 
> I feel like a step-level hint would be rather obnoxious and that the above would be less work

Let me walk through the expected behavior just so I understand what you are saying. 

If a user says they want outputs from stdout/stderr, then we should ignore the log-dir field.

If a  user  specifies stdout/stderr fields which will tell the tool to log stdout/stderr and give the log file a name.  This would work with the log-dir parameter.  If you specify a log-dir parameter, we generate a uuid under the log directory and stick the named log files there. 

If a user does not put stdout/stderr in the CLT definition and no stdout/stderr outputs, then I won't log, ignoring the log-dir, because we aren't telling the tool that we want logs.  

The problem arises if someone specifies stdout/stderr and output fields.  The output looks in the container and tries to stage to host system.  

So if outputs and stdout/stderr are not being captured, if someone uses log-dir, then we assume that they want to capture logs for every tool.  And we just generate a random uuid for the log files?

kannon92:
Forget what I wrote above.  

Reading through the code.  I think the best thing to do is to override the staging code (which copies outputs from container to outdir) to use my new log path.   If a user wants the logs as part of the output, we will move them from their log-dir directory to the outdir.  Should be same behavior as before.

tests/test_stdout_stderr_log_dir.py
```diff
@@ -0,0 +1,24 @@
...
+    _, stdout, stderr = get_main_output(
+        ["--log-dir", "logs", "tests/echo.cwl", "--inp", "hello"]
+    )
+    assert "completed success" in stderr, stderr
+    assert json.loads(stdout)["out"].strip("\n") == "hello"
+
+
+def test_log_dir_echo_no_output() -> None:
+    _, stdout, stderr = get_main_output(
+        ["--log-dir", "logs", "tests/echo-stdout-log-dir.cwl", "--inp", "hello"]
```
mr-c, line 20:
```suggestion
from .util import get_data, get_main_output, needs_docker


def test_log_dir_echo_output() -> None:
    _, stdout, stderr = get_main_output(
        ["--log-dir", "logs", get_data("tests/echo.cwl"), "--inp", "hello"]
    )
    assert "completed success" in stderr, stderr
    assert json.loads(stdout)["out"].strip("\n") == "hello"


def test_log_dir_echo_no_output() -> None:
    _, stdout, stderr = get_main_output(
        ["--log-dir", "logs", get_data("tests/echo-stdout-log-dir.cwl"), "--inp", "hello"]
```

kannon92:
Hey @mr-c, is there anything else you'd want from me for this PR?  Otherwise, I think its good to go whenever you want.

mr-c:
Lets ask @tetron about this at today's meeting

cwltool/argparser.py
```diff
@@ -37,6 +37,13 @@ def arg_parser() -> argparse.ArgumentParser:
         help="Output directory. The default is the current directory.",
     )
 
+    parser.add_argument(
+        "--log-dir",
+        type=str,
+        default="",
+        help="Log your tools stdout/stderr to this location outside of container",
```
mr-c, line 44: Is `--log-dir` supposed to work for tools that output to stdout/stderr but we don't capture that in the `CommandLineTool` description itself? This help text implies that it does, but my testing says it does not currently do so.

cwltool/job.py
```diff
@@ -278,8 +278,12 @@ def _execute(
...
             ),
             " < %s" % self.stdin if self.stdin else "",
-            " > %s" % os.path.join(self.outdir, self.stdout) if self.stdout else "",
-            " 2> %s" % os.path.join(self.outdir, self.stderr) if self.stderr else "",
+            " > %s" % os.path.join(self.base_path_logs, self.stdout)
+            if self.stdout
+            else "",
+            " 2> %s" % os.path.join(self.base_path_logs, self.stderr)
+            if self.stderr
+            else "",
```
mr-c, line 286: Here you could try to capture stdout/stderr logs `if runtimeContext.log_dir != ""`

tests/test_stdout_stderr_log_dir.py
```diff
@@ -0,0 +1,30 @@
...
+            "--log-dir",
+            "logs",
+            get_data("tests/echo-stdout-log-dir.cwl"),
+            "--inp",
+            "hello",
+        ]
+    )
+    for dir in os.listdir("logs"):
+        for file in os.listdir(f"logs/{dir}"):
+            assert file == "out.txt"
```
mr-c, line 30: and check the contents?

mr-c (changes requested):

cwltool/argparser.py
kannon92: I think you need either a stdout/stderr field in the CommandLineTool description or the logs are captured as an output.

cwltool/job.py
kannon92: I could.  Not sure what is best.  I could capture logs if someone specifes log_dir but if stderr and stdout are undefined, do you think it could be an issue?  

I can always generate a unique id for stdout and stderr but not sure if that could break tools that specifically don't want to capture stdout/stderr.

I went with an opt in approach right now.

mr-c (approved):

PR #1613, opened 2022-02-02
Handle URLs for command line inputs

tetron:
Uses loader's urljoin to construct relative references based on the file uri of the current working directory, which allows other types of URIs to pass through the command line unbothered.

Also handle edge case where input parameter type is wrapped in single-item list.

### DIFF
--- a/cwltool/argparser.py
+++ b/cwltool/argparser.py
@@ -5,6 +5,7 @@ import os
 from typing import (
     Any,
     AnyStr,
+    Callable,
     Dict,
     List,
     MutableMapping,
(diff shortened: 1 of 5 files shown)


mr-c:
@tetron Check your version of `black`; we upgraded to the (20)22 release (which is now out of beta)

cwltool/argparser.py
```diff
@@ -804,27 +825,32 @@ def add_argument(
     # parameter required.
     required = default is None and input_required
     if isinstance(inptype, MutableSequence):
-        if inptype[0] == "null":
+        if len(inptype) == 1:
+            inptype = inptype[0]
+        elif len(inptype) == 2 and inptype[0] == "null":
+            required = False
+            inptype = inptype[1]
```
mr-c, line 832: I would have appreciated test coverage for these lines

PR #1615, opened 2022-02-09
scandeps should not apply mergedirs at every level

tetron:
When getting back a list of dependencies, and using "nestdirs", in the
specific case it there are references to both a directory and files
within that directory, it may report multiple dependencies for the
same directory.  The mergedir method merges those references into one.

### DIFF
--- a/cwltool/process.py
+++ b/cwltool/process.py
@@ -1104,42 +1104,28 @@ def nestdir(base: str, deps: CWLObjectType) -> CWLObjectType:
         sp = s2.split("/")
         sp.pop()
         while sp:
+            loc = dirname+"/".join(sp)
             nx = sp.pop()
-            deps = {"class": "Directory", "basename": nx, "listing": [deps]}
+            deps = {"class": "Directory", "basename": nx, "listing": [deps], "location": loc}
     return deps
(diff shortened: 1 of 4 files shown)


mr-c (changes requested):
Thanks, Peter. Can we get a test case to prevent regressions?

cwltool/process.py
```diff
@@ -1104,42 +1104,37 @@ def nestdir(base: str, deps: CWLObjectType) -> CWLObjectType:
...
     ents = {}  # type: Dict[str, CWLObjectType]
-    collided = set()  # type: Set[str]
     for e in listing:
         basename = cast(str, e["basename"])
         if basename not in ents:
             ents[basename] = e
+        elif e["location"] != ents[basename]["location"]:
+            raise Exception(
+                "Conflict between %s and %s", e["location"], ents[basename]["location"]
+            )
```
mr-c, line 1130: Can we get a test to confirm the conflict detection as well?

mr-c (changes requested):
Thanks, I have manually confirmed that your test fails without these fixes. Looks like we have one more area that needs testing

PR #1620, opened 2022-02-14
Revert "Drop Python 3.6"

tetron:
Ubuntu 18.04 (LTS, supported until April 2023) and Centos 7 are still
on Python 3.6.

I recognize that the reason 3.6 was dropped was that Python upstream doesn't support 3.6 with security updates any more, but in practice people use whatever Python is supported by their Linux distribution.

This reverts commit 780f64a71fd9778d8481ede998c9c3da91bce4aa.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,9 +33,9 @@ Quay.io (Docker): |Quay.io Container|
 .. |Total PyPI Downloads| image:: https://static.pepy.tech/personalized-badge/cwltool?period=total&units=international_system&left_color=black&right_color=orange&left_text=Total%20PyPI%20Downloads
    :target: https://pepy.tech/project/cwltool
 
-.. |Conda Version| image:: https://anaconda.org/conda-forge/cwltool/badges/version.svg 
+.. |Conda Version| image:: https://anaconda.org/conda-forge/cwltool/badges/version.svg
    :target: https://anaconda.org/conda-forge/cwltool
-   
+
 .. |Conda Installs| image:: https://anaconda.org/conda-forge/cwltool/badges/downloads.svg
    :target: https://anaconda.org/conda-forge/cwltool
 
@@ -84,7 +84,7 @@ If you encounter an error, first try to update package information by using
    sudo apt-get update
 
 If you are running macOS X or other UNIXes and you want to use packages prepared by the conda-forge project, then
-please follow the install instructions for `conda-forge <https://conda-forge.org/#about>`_ (if you haven't already) and then 
+please follow the install instructions for `conda-forge <https://conda-forge.org/#about>`_ (if you haven't already) and then
 
 .. code:: bash
 
@@ -142,11 +142,11 @@ system link or `another facility <https://wiki.debian.org/DebianAlternatives>`_.
 Recommended Software
 ^^^^^^^^^^^^^^^^^^^^
 
-You may also want to have the following installed: 
+You may also want to have the following installed:
 - `node.js <https://nodejs.org/en/download/>`_
 - Docker, udocker, or Singularity (optional)
 
-Without these, some examples in the CWL tutorials at http://www.commonwl.org/user_guide/ may not work. 
+Without these, some examples in the CWL tutorials at http://www.commonwl.org/user_guide/ may not work.
 
 Run on the command line
 -----------------------
@@ -694,7 +694,7 @@ List of all environment can be seen using:
 and running a specfic test env using:
 ``tox -e <env name>``
 and additionally run a specific test using this format:
-``tox -e py36-unit -- -v tests/test_examples.py::test_scandeps``
+``tox -e py310-unit -- -v tests/test_examples.py::test_scandeps``
 
 -  Running the entire suite of CWL conformance tests:
 
--- a/README.rst
+++ b/README.rst
@@ -694,7 +694,7 @@ List of all environment can be seen using:
 and running a specfic test env using:
 ``tox -e <env name>``
 and additionally run a specific test using this format:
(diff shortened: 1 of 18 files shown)


tom-tan:
CentOS 7 is supported until [Jun 30th, 2024](https://www.centos.org/download/).
It needs an effort to support Python 3.6 for another two years.

Fortunately we already have [venv](https://docs.python.org/3/library/venv.html) as a workaround of this issue and it is documented in the [README of cwltool](https://github.com/common-workflow-language/cwltool#install).

Are there reasons to directly support OS built-in version of Python rather than encouraging to use `venv`?

mr-c:
> Ubuntu 18.04 (LTS, supported until April 2023) and Centos 7 are still
> on Python 3.6.

Ubuntu 18.04 (Bionic Beaver) ships Python 3.7 and 3.8

https://packages.ubuntu.com/search?suite=bionic&keywords=python3.7

https://packages.ubuntu.com/search?suite=bionic&keywords=python3.8

So `sudo apt install python3.8 && python3.8 -m pip install --user cwltool` should work

Python 3.8 can be added to CentOS from their official repos:

> 1.Install CentOS SCLo RH repository:
>
>    `yum install centos-release-scl-rh`
>
> 2. Install rh-python38-python rpm package:
>
>    `yum install rh-python38-python`

From https://centos.pkgs.org/7/centos-sclo-rh-x86_64/rh-python38-python-3.8.6-1.el7.x86_64.rpm.html

I would be more interested in a PR adding these instructions to the docs. Bonus if there is a variation of https://github.com/common-workflow-language/cwltool/blob/main/cwltool.Dockerfile for Ubuntu 18.04 and CentOS 7 demonstrating that this works.

And speaking of the docs, this is a good time to start breaking up the giant README.rst and move the installation instructions to their own page, like was done for https://cwltool.readthedocs.io/en/latest/processgen.html

README.rst
```diff
@@ -694,7 +694,7 @@ List of all environment can be seen using:
 and running a specfic test env using:
 ``tox -e <env name>``
 and additionally run a specific test using this format:
-``tox -e py310-unit -- -v tests/test_examples.py::test_scandeps``
+``tox -e py36-unit -- -v tests/test_examples.py::test_scandeps``
```
mr-c, line 697: No reason to change this

cwltool/job.py
```diff
@@ -50,8 +50,8 @@
 from .utils import (
     CWLObjectType,
     CWLOutputType,
-    DirectoryType,
     HasReqsHints,
+    DirectoryType,
```
mr-c, line 54: ? Probably best to run `make sort-imports`

setup.py
```diff
@@ -150,10 +151,10 @@
         "Operating System :: POSIX",
         "Operating System :: POSIX :: Linux",
         "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
-        "Programming Language :: Python :: 3.10",
```
mr-c, line 156: We still support Python 3.10

tox.ini
```diff
@@ -16,20 +16,21 @@ testpaths = tests
...
+  3.6: py36
   3.7: py37
   3.8: py38
   3.9: py39
   3.10: py310
 
 [testenv]
 skipsdist =
-  py3{7,8,9,10}-!{unit,mypy,lintreadme} = True
+  py3{5,7,8,9,10}-!{unit,mypy,lintreadme} = True
```
mr-c, line 27: 6 not 5

mr-c (changes requested):
I prefer we document how to install on CentOS 7 and Ubuntu 18.04 using the available systems packages.

tetron:
@mr-c we're looking into building on ubuntu 18.04 and centos 7 packages with a more recent python.  I'm going to keep this PR open until we have a solution.

cure:
I would really like to seem more process around the depreciation of Python versions. This is not the first time we run into this. Changes like these seem extremely arbitrary, and they cause a pile of sudden, urgent work for no good reason.

Depreciations should not happen suddenly and there should be a sufficient amount of warning. 

Obviously, the "upstream stopped supported security updates" argument carries some weight. However, it is not that simple, and it's not true, because distributions do provide security fixes for older Python versions, especially the ones they ship with by default (e.g. Ubuntu 18.04, where Python 3.6 is the default "python3" package).

mr-c:
@cure Sure, I agree a formal policy would be nice. Would you like to start a draft PR that we could send around to the other users of `cwltool` to review?

Here is shared policy of Matplotlib, scikit-learn, IPython, Jupyter, yt, SciPy, NumPy, and scikit-image as a starting point:

https://numpy.org/neps/nep-0029-deprecation_policy.html

FYI, Conda (Anaconda and conda-forge) and Debian have both moved on from Python 3.6, and as I'm the maintainer of the `cwltool` packages for both, it is problematic to carry on additional old version of Python when there are other alternatives.

https://conda.io/en/latest/miniconda.html

tetron:
> Sure, I agree a formal policy would be nice. Would you like to start a draft PR that we could send around to the other users of cwltool to review?

proposed a policy at PR #1621

>  FYI, Conda (Anaconda and conda-forge) and Debian have both moved on from Python 3.6, and as I'm the maintainer of the cwltool packages for both, it is problematic to carry on additional old version of Python when there are other alternatives.

If it happens in the future that an essential dependency is only available for Python 3.7+ and this is a problem for packaging then we will need to examine our options, but I don't think preemptively dropping support for Python 3.6 is the right way to go about it.

mr-c (approved):

PR #1627, opened 2022-02-21
Fix a few help texts

JelleZijlstra:
The continued strings on a separate line just get ignored by the runtime. You need to put them in parentheses to turn them into a single string.


kinow (approved):
Thanks!!!

mr-c (approved):

PR #1637, opened 2022-03-03
less random files left over after tests

mr-c:


tetron:
Where does `tmp_path` come from, is it a pytest thing?

mr-c:
> Where does `tmp_path` come from, is it a pytest thing?

Yep, it is a built-in pytest fixture

https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture

https://docs.pytest.org/en/6.2.x/fixture.html

tetron (approved):

Issue #1639, opened 2022-03-07
Incorrect validation warning  when combining "scatter" and "when"


tetron:
## Expected Behavior
Should not give a warning when scattering with a conditional connects to an array of optional values.

## Actual Behavior
It gaves a warning because the output type is computed incorrectly for the order of operations in a step that has both a scatter and a conditional.

## Workflow Code

```
steps:
  fastq-validator:
    run: tools/fastqValidator.cwl
    in:
      file:
        - fastq1
        - fastq2
    out: [std_out]
    scatter: file
    when: $(inputs.file != null)

  adjust-names:
    in:
      sample_name: sample_name
      fqValidator: 50-fastq-validator/std_out
    out: [fq1, fq2]
    run: adjust-names.cwl
```

```
workflow.cwl:21:11: Source 'std_out' of type ["null", {"type": "array", "items": "File"}] may be
                    incompatible
workflow.cwl:28:7:    with sink 'fqValidator' of type {"type": "array", "items": ["null", "File"]}
                      Source is from conditional step and may produce `null`
```
The checker is using the wrong type for `fastq-validator/std_out`, it should be {"type": "array", "items": ["null", "File"]}.

## Your Environment
cwltool 3.1.20220221215214

PR #1640, opened 2022-03-07
Fix spurious validation error with scatter+conditionals

tetron:
Typing related to conditionals should be part of the computed output
signature of the step, not added in by the checker.

refs #1639

### DIFF
--- a/cwltool/workflow.py
+++ b/cwltool/workflow.py
@@ -359,6 +359,8 @@ class WorkflowStep(Process):
                     "SubworkflowFeatureRequirement not in requirements"
                 )
 
+        conditional = self.tool.get("when")
+
         if "scatter" in self.tool:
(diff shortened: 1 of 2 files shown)


cwltool/workflow.py
```diff
@@ -398,11 +400,22 @@ def __init__(
             else:
                 nesting = 1
 
-            for _ in range(0, nesting):
+            for n in range(0, nesting):
                 for oparam in outputparms:
+                    if conditional and n == 0:
+                        oparam["type"] = ["null"] + cast(
```
mr-c, line 406: Can you add a test case that triggers this line?

mr-c (changes requested):

cwltool/workflow.py
tetron: Yes, I was going to add a test, I just wanted to see what the other code checks had to say about it.

PR #1641, opened 2022-03-12
implement the loop construct extension 

mr-c:
Based upon @GlassOfWhiskey 's work in https://github.com/common-workflow-language/common-workflow-language/issues/495#issuecomment-1064343639
With comments from @tetron @mr-c


cwltool/extensions-v1.2.yml
```diff
@@ -0,0 +1,249 @@
...
+        Minimum number of GPU devices to request.  If not specified,
+        same as `cudaDeviceCountMax`.  If neither are specified,
+        default 1.
+    cudaDeviceCountMax:
+      type: ['null', int, cwl:Expression]
+      doc: |
+        Maximum number of GPU devices to request.  If not specified,
+        same as `cudaDeviceCountMin`.
+
+- name: LoopInput
```
mr-c, line 123: This file is identical to `extensions-v1.1.yml` except from here on out

mr-c (commented):
@GlassOfWhiskey Very exciting! Can you add tests to cover the missing lines of `cwltool/{checker,workflow_job}.py`?

You can check your progress via `make clean diff-cover`

tests/test_loop.py
```diff
@@ -0,0 +1,216 @@
...
+    assert json.loads(stream.getvalue()) == expected
+
+
+def test_loop_value_from_fail_no_requirement() -> None:
+    """Test that a workflow loop fails if a valueFrom directive is specified without StepInputExpressionRequirement."""
+    params = [
+        "--enable-ext",
+        get_data("tests/loop/invalid-value-from-loop-no-requirement.cwl"),
+        get_data("tests/loop/two-vars-loop-job.yml"),
+
```
mr-c, line 139:
```suggestion
```

cwltool/workflow_job.py
```diff
@@ -559,15 +564,31 @@ def receive_output(
...
+        runtimeContext: RuntimeContext,
+        container_engine: str,
         jobout: CWLObjectType,
         processStatus: str,
     ) -> None:
 
+        requirements = {
+            **{h["class"]: h for h in step.tool.get("hints", [])},
+            **{r["class"]: r for r in step.tool.get("requirements", [])},
+        }
```
tetron, line 576: Is there a reason to do this instead of calling `step.tool.get_requirement("http://commonwl.org/cwltool#Loop")` ?

cwltool/workflow_job.py
```diff
@@ -586,6 +607,113 @@ def receive_output(
         else:
             _logger.info("[%s] completed %s", step.name, processStatus)
 
+        if "http://commonwl.org/cwltool#Loop" in requirements:
```
tetron, line 610: There's a lot going on here.

Architecturally, I'd like to see this go into its own class, the way the scatter methods use ReceiveScatterOutput.  I'll call in `LoopStep` for the sake of discussion.

This would let you encapsulate the loop logic into a class instead of adding a large amount of new code to a method that was previously much shorter.

I would also like to see the loop logic go into a function that returns a JobsGenerator, similar to the way scatter works.

So the sketch of the way this would work is:

1. in `try_make_job`, if there's a loop, it calls `jobs = LoopStep().job(step, inputobj, callback, runtimeContext)`  This creates a LoopStep object and calls the job generator method on it.
2. in `LoopStep.job()` it checks the loop condition.
3. LoopStep has a method called loop_callback
4. if the condition passes, it yields the first iteration `step.job(inputobj, self.loop_callback, runtimeContext)`
5. When that finishes, loop_callback is called with the result of the step.  Record the results, update the input object, and set a flag that an iteration has completed
6. Control will periodically come back to `LoopStep.job` which will check that flag.  If there's nothing ready yet, return None
7. If an iteration completed, we go back to the top and check the loop condition.  If the loop condition is still true, yield step.job() with the updated inputobj
8. If the loop condition is false, call the _original_ callback with the final results

tetron (changes requested):

cwltool/extensions-v1.2.yml
```diff
@@ -0,0 +1,241 @@
...
+      jsonldPredicate:
+        _id: "cwltool:loop"
+        mapSubject: id
+        mapPredicate: loopSource
+      doc: |
+        Defines the input parameters of the loop iterations after the first one
+        (inputs of the first iteration are the step input parameters). If no
+        `loop` rule is specified for a given step `in` field, the initial value
+        is kept constant among all iterations.
+    outputMethod:
```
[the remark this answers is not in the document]
mr-c: Fixed, thanks

cwltool/extensions-v1.2.yml
```diff
@@ -0,0 +1,241 @@
...
+        Only run the step while the expression evaluates to `true`.
+        If `false` and no iteration has been performed, the step is skipped.
+
+        A skipped step produces a `null` on each output.
+
+        The `inputs` value in the expression must be the step input object.
+
+        It is an error if this expression returns a value other than `true` or `false`.
+    loop:
+      type: LoopInput[]?
```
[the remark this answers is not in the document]
mr-c: I've made the `loop` field mandatory

cwltool/workflow_job.py
GlassOfWhiskey: No, simply didn't know the existence of that method XD. Fixed by calling `step.step.get_requirement`.

cwltool/workflow_job.py
```diff
@@ -879,3 +883,223 @@ def job(
...
+                    None,
+                    {},
+                    debug=runtimeContext.debug,
+                    js_console=runtimeContext.js_console,
+                    timeout=runtimeContext.eval_timeout,
+                    container_engine=self.container_engine,
+                )
+                if whenval is True:
+                    yield from self.step.job(self.joborder, callback, runtimeContext)
+                    if self.processStatus == "permanentFail":
```
tetron, line 938: When yield returns control to this method, you can't assume the step has actually finished running.  You only know that the step is finished when the callback has been called.

I suggest setting `processStatus = ""` before doing `yield from self.step.job(...)`

```
self.processStatus = ""
yield from self.step.job(...)
while self.processStatus == "":
  yield None
```

The loop callback needs to always set `self.processStatus = processStatus` to indicate the callback was called and the loop should proceed.

cwltool/workflow_job.py
```diff
@@ -879,3 +883,223 @@ def job(
...
+                        output_callback(self.output_buffer, self.processStatus)
```
tetron, line 939: You need a `return` after this

cwltool/workflow_job.py
```diff
@@ -879,3 +883,223 @@ def job(
...
+                else:
+                    raise WorkflowException(
+                        "Loop condition 'loopWhen' must evaluate to 'true' or 'false'"
+                    )
+        except WorkflowException:
+            raise
+        except Exception:
+            _logger.exception("Unhandled exception")
+            self.processStatus = "permanentFail"
+            self.step.completed = True
```
tetron, line 975: Instead of setting `step.completed` this should be calling output_callback().

cwltool/workflow_job.py
```diff
@@ -879,3 +883,223 @@ def job(
...
+                _logger.debug(
+                    "Iteration %i of [%s] produced output %s",
+                    self.iteration,
+                    self.step.name,
+                    json_dumps(jobout, indent=4),
+                )
+
+            if processStatus not in ("success", "skipped"):
+                if self.processStatus != "permanentFail":
+                    self.processStatus = processStatus
```
tetron, line 1025: not quite following the logic here but if you follow my other comment you need to move this out so that `                    self.processStatus = processStatus` is always called

tetron (changes requested):

cwltool/workflow_job.py
GlassOfWhiskey: This is not clear to me. Why do you say that the process could not be finished? There is no fire and forget logic inside the implementation.

tetron: The job() method returns a python generator (iterator protocol).  It does one of three things.

1. yield a Job object (which has a run() method)
2. yield None (there's nothing to do right now, but there might be something to do in the future
3. return and end the iterator (the operation is completed)

The job object is invoked at some point in the future.  In particular, with the multi-threaded executor, it goes into a queue.

Meanwhile, we keep calling the iterator to get more work.  For example, a scatter step is going to produce N jobs, an then idle (return None) until all those jobs have completed.

At the top level of the workflow, we loop over each step and the ones which are ready to run, we call the job() method to ask that step to generate jobs for us.  If that step returns None we'll go on to the next thing, but we'll eventually loop around back to iterate on the step again.

It might be helpful to think of these iterators as channels feeding job objects up to the top level.  Part of the process is going through each iterator and asking it if it has new values to produce or not (sort of like reading a channel, except that we are actually passing control to the iterator which has to yield it back).  

Does this sort of make sense?

docs/loop.rst
```diff
@@ -0,0 +1,139 @@
...
+    in:
+      i1: i1
+    out: [o1]
+    requirements:
+      cwltool:Loop:
+        loopWhen: $(inputs.i1 < 10)
+        loop:
+          i1: o1
+        outputMethod: last
+This loop executes untile the counter ``i1`` reaches the value of 10, and then terminates. Note that if the ``loopWhen`` condition evaluates to ``false`` prior to the first iteration, the loop is skipped. The value assumed by the output fields depends on the specified ``outputMethod``, as described below.
```
tetron, line 33: typo "untile"

tetron (approved):

PR #1655, opened 2022-04-25
Changed cwlviewer.py to make node color dependent on operation class

RenskeW:
Node color/style:
- CommandLineTool: "lightgoldenrodyellow", "filled"
- Workflow: "#F3CEA1", "filled"
- Operation: [no color], "dashed"

CommandLineTool & Workflow now have the same color as [cwlviewer](https://github.com/common-workflow-language/cwlviewer).

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -72,7 +72,7 @@ jobs:
 
       - name: Upload coverage to Codecov
         if: ${{ matrix.step == 'unit' }}
-        uses: codecov/codecov-action@v2.1.0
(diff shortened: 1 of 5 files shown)


mr-c (changes requested):
Super! Can you add a test case that confirms the expected outcome? Perhaps using an workflow with each type of step represented?

Bonus if you paste in a picture of the resulting graph as a comment here 🙂

RenskeW:
Schermafbeelding 2022-05-02 om 07 49 13

mr-c (approved):
Huzzah!

PR #1665, opened 2022-05-06
draft schema for paramter restrictions extension

mr-c:
Requires https://github.com/common-workflow-language/schema_salad/pull/535


tetron:
Is this related to the CTD mapping/translation effort?

mr-c:
> Is this related to the CTD mapping/translation effort?

Yes, this is the beginning of an implementation of [the input value restriction proposal inspired by CTD](https://github.com/common-workflow-language/common-workflow-language/issues/764)

Basically to get the checks working. I'd like to develop a more concise syntax either before the extension is released, or before the proposal to CWL v1.3+ is made

tetron:
Is there documentation you can link to specifically for the CTD value restriction model?

The 2018 proposal seems a little over-complicated with different types of intervals.

I started working on a simpler proposal here but haven't followed up on it:

https://github.com/common-workflow-language/schema_salad/tree/constraints

mr-c:
> Is there documentation you can link to specifically for the CTD value restriction model?

Nothing in their XSD, alas https://github.com/WorkflowConversion/CTDSchema/blob/5ab34f57cc18cf472aac109e9bdde0e40855dda2/Param_1_7_0.xsd#L39

Digging through the CTDConverter & CTDOpts code we learn that the `restrictions` field is either a comma-separated list of allowable values, or a simple numeric range separated by a colon

https://github.com/WorkflowConversion/CTDopts/blob/eec3d874bcf1ff66724cf3d7cda2dd6a70292e97/CTDopts/CTDopts.py#L173

It is the CTD people who asked for more control over intervals in the 2018 proposal; is each end of the interval exclusive or inclusive?

cwltool/process.py
```diff
@@ -804,6 +810,98 @@ def _init_job(
...
+                                    rpattern = constraint["rpattern"]
+                                    quoted_value = json.dumps(value)
+                                    matched = cast(
+                                        bool,
+                                        execjs(
+                                            f"/{rpattern}/.test({quoted_value})",
+                                            "",
+                                            runtime_context.eval_timeout,
+                                            runtime_context.force_docker_pull,
+                                        ),
```
kinow, line 890: Couldn't we use `re.search(rpattern, quoted_value)` or another Python statement here?

kinow (commented):
Looking good!

cwltool/process.py
mr-c: We could, but then we'd have to codify the Python RegEx rules into the CWL standard; it is simpler to use the ECMAScript 5.1 regex functionality as that matches what we already use for CWL Expressions

kinow: Good point, thanks for explaining it :+1:

PR #1676, opened 2022-06-03
Set last modification date of remote file as local file timestamp

RenskeW:
Changed cwltool/utils.py::downloadHttpFile to set Last-Modified date of remote file as timestamp of downloaded file.

### DIFF
--- a/tests/test_http_input.py
+++ b/tests/test_http_input.py
@@ -6,7 +6,7 @@ from datetime import datetime
 from cwltool.pathmapper import PathMapper
 from cwltool.utils import CWLObjectType
 
-from pytest_httpserver import HTTPServer
+from pytest_httpserver import HTTPServer  # type: ignore[import]
 
 
 def test_http_path_mapping(tmp_path: Path) -> None:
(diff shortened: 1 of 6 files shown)


mr-c (changes requested):
Consider using https://pytest-httpserver.readthedocs.io/en/latest/ so we don't have to rely on a remote server?

tests/test_http_input.py
```diff
@@ -25,3 +28,46 @@ def test_http_path_mapping(tmp_path: Path) -> None:
...
+        "Strict-Transport-Security": "max-age=31536000; includeSubDomains",
+    }
+
+    remote_file_name = "testfile.txt"
+
+    with HTTPServer() as httpserver:
+        httpserver.expect_request(f"/{remote_file_name}").respond_with_data(
+            response_data="Hello World", headers=headers
+        )
+        location = httpserver.url_for(f"/{remove_file_name}")
```
mr-c, line 54:
```suggestion
        location = httpserver.url_for(f"/{remote_file_name}")
```

mr-c (approved):

PR #1678, opened 2022-06-07
Propagate input annotations to primary.cwlprov files

RenskeW:
Propagate structured annotations for files and directories (e.g. Schema.org, but also CWL-specific metadata fields) to the provenance graph. In addition, annotations under the `cwlprov:prov` field should be incorporated in the RDF graph as well. 

Step 1: If 'label' is specified in input object, it is added as an annotation to the file entity it corresponds to.

Example:
```
file: 
  class: File
  path: ./file.tsv
  label: file_label
```
This would be represented in the cwlprov provenance like this:

```
id:92fc295f-f6cb-444b-8ffe-0f604b127858 a wf4ever:File,
        wfprov:Artifact,
        prov:Entity ;
    rdfs:label "file_label"^^xsd:string ;
    prov:specializationOf data:86ec86fdbbe70863a78453f69349568f9d1f14a1 ;
    cwlprov:basename "file.tsv"^^xsd:string ;
    cwlprov:nameext ".tsv"^^xsd:string ;
    cwlprov:nameroot "file"^^xsd:string .
```
See [here](https://github.com/RenskeW/cwlprov-provenance/tree/main/prov_data_annotations/example2) for the full example RO.

Step 2: Values of `doc`, `format`, `label`, `intent` are propagated to RDF provenance graph. 

**EXAMPLE RO:** https://github.com/RenskeW/cwlprov-provenance/tree/930cc268da77c5aa3739a2f7f87b94e076b144e6/cwlprov_rdf_examples

Screenshot 2023-11-06 at 12 24 31

Mapping to Schema.org terms:

- `doc`: https://schema.org/description
- `label`: https://schema.org/name
- `format`: https://schema.org/encodingFormat
- `intent`: https://schema.org/featureList

TODO: 
- [x] Propagate annotations for File
- [x] Propagate annotations for Directory
- [x] Add support for annotations of type CommentedSeq
- [ ] Propagate annotations under `cwlprov:prov`
- [ ] Propagate values of `doc`, `intent`, `label`, and `intent` fields to RDF
- [ ] inputs should be copied into `data/` , not be broken symlinks
- [ ] improve code coverage
- [ ] Fix double entries in provenance (optionally)


mr-c:
Thank you for this @RenskeW ! can you `make format` and push the changes?

mr-c:
@RenskeW Thanks!

Checking https://www.commonwl.org/v1.2/CommandLineTool.html#File we see that `label` is not one of the fields present here in the CWL standards. So we'll need to use a namespaced field name, for which I recommend that you choose one of the schema.org properties like https://schema.org/name as it is linked to `rdfs:label` and is modeled as equivalent to `dcterms:title`.

After you make that change, can you add a `s:name` to one of the input objects used in the cwlprov tests? You'll need to add a `$namespaces` dictionary to document the `s` prefix if the corresponding CWL description lacks that. Then you can add a check to one of the cwlprov tests to  look for the `rdfs:label` in the resulting CWLProv RO.

mr-c:
(maybe in CWL v1.3 we will add a label field to `File`)

cwltool/provenance_profile.py
```diff
@@ -350,6 +369,29 @@ def declare_file(self, value: CWLObjectType) -> Tuple[ProvEntity, ProvEntity, st
             file_entity.add_attributes({CWLPROV["nameext"]: value["nameext"]})
         self.document.specializationOf(file_entity, entity)
 
+        
+
+        def _add_nested_annotations(dataset, e: ProvEntity) -> ProvEntity:
+            for annotation in dataset:
+                if isinstance(dataset[annotation], (str, bool, int, float)): # check if these are all allowed types
```
RenskeW, line 376: isinstance(dataset[annotation], list), isinstance(dataset[annotation], MutableMapping)

RenskeW: MutableSequence = list

cwltool/provenance_profile.py
```diff
@@ -296,6 +297,24 @@ def record_process_end(
...
+            annotation_key = SCHEMA[annotation_key.replace(schema2_uri, '')].uri 
+        
+        if not isinstance(annotation_value, (MutableSequence, MutableMapping)):            
+            e.add_attributes({annotation_key: str(annotation_value)})
+        else:
+            nested_id = uuid.uuid4().urn
+            nested_entity = self.document.entity(nested_id)
+            e.add_attributes({annotation_key: nested_entity.identifier})
+            for nested_key in annotation_value:
+                nested_value = annotation_value[nested_key]
```
RenskeW, line 314: This gives an error when annotation_value is a CommentedSeq 
= Input file has array of annotations like this:
```
s:hasPart:
- s:identifier: <identifier1>
- s:identifier: <identifier2>
```

cwltool/provenance_profile.py
```diff
@@ -350,6 +369,21 @@ def declare_file(self, value: CWLObjectType) -> Tuple[ProvEntity, ProvEntity, st
             file_entity.add_attributes({CWLPROV["nameext"]: value["nameext"]})
         self.document.specializationOf(file_entity, entity)
 
+        # Identify all schema annotations
+        schema_annotations = dict([(v, value[v]) for v in value.keys() if 'schema.org' in v])
+
+        # Transfer SCHEMA annotations to provenance
+        for s in schema_annotations:
+            if "additionalType" in s:
+                additional_type = schema_annotations[s].split(sep='/')[-1] # find better method?
```
RenskeW, line 378: Necessary to anticipate `\` instead of `/`?

RenskeW: URL always has / so not necessary

cwltool/provenance_profile.py
RenskeW: Add if CommentedSeq

cwltool/provenance_profile.py
```diff
@@ -247,6 +248,18 @@ def evaluate(
...
+            #     metadata = job_order_object[CWLPROV['prov'].uri] # change uri to CWLPROV['prov'].uri
+            #     for item in metadata:
+            #         # make a new entity with id
+            #         # give it type additionalType value
+            #         # add nested annotations
+            #         # how much of this can we reuse from _add_nested_annotations?
+            #         # how do we identify the correct file to write to? self.workflow_run_uri?
+            #         # 
+            #         pass
+
```
RenskeW, line 261: First step in propagating metadata under `cwlprov:prov` to provenance as well.

PR #1683, opened 2022-06-22
Message about other implementations

tetron:
This was merged by accident and reverted, bringing it back for discussion

* Text might be too "market-y" so looking for suggestions on toning it down
* After seeing it a few times people might find it tiresome so looking for suggestions on how to turn it off


mr-c (changes requested):
I think a comment in the README and `--help` output would be the better place. Otherwise this reads too much like an advertisement

PR #1709, opened 2022-08-16
Implementation of store RDF export of the workflow in CWL Prov RO-Bundle

jjkoehorst:
Keeping as draft while we test proposed enhancements to CWLProv 

Added file size to prov RO-Bundle

Creates a workflow.ttl file in the default folder while running.
Then closes the file and after the provenance folder is created it is copied to the provenance/workflow/ folder.

temp directory creation and removal added


kinow (commented):
GH Actions is showing some linter errors. I think running `make format` should fix the code locally, then you can review if it modified only the files you modified, and push it again (sometimes it modifies more than what I changed, and then CI keeps failing; just ignore if `make format` changes more than what you expected).

jjkoehorst:
> GH Actions is showing some linter errors. I think running `make format` should fix the code locally, then you can review if it modified only the files you modified, and push it again (sometimes it modifies more than what I changed, and then CI keeps failing; just ignore if `make format` changes more than what you expected).

Thanks many jobs now fail as well due to:

```
E           bagit.BagValidationError: Bag validation failed: data/26/260807aefd9032a48a0d88d4ff7110a5bbfaea3d sha1 validation failed: expected="260807aefd9032a48a0d88d4ff7110a5bbfaea3d" found="da39a3ee5e6b4b0d3255bfef95601890afd80709"; data/ff/ff1f8b01ee9e74ec5f9840aa07ea6bc2895a235e sha1 validation failed: expected="ff1f8b01ee9e74ec5f9840aa07ea6bc2895a235e" found="da39a3ee5e6b4b0d3255bfef95601890afd80709"
```

As I am removing the provenance /data files and will turn it into an option later. This way we can use input files again that are ±100gb in size and prevent them from being copied to the data folder

cwltool/main.py
```diff
@@ -8,9 +8,12 @@
 import io
 import logging
 import os
+import parser
```
kinow, line 11: I think is not needed @jjkoehorst ?

I cannot see `parser` being used here in the GitHub UI diff. If I open the complete module, I can see a `parser` variable being instantiated, but nothing imported - https://github.com/common-workflow-language/cwltool/blob/bd61e436150c87af758061130a9c6650a0918ea0/cwltool/main.py#L1022

After removing this line the `cwltool` command works with no issues.

tests/test_provenance.py
```diff
@@ -48,7 +48,7 @@ def cwltool(tmp_path: Path, *args: Any) -> Path:
 def cwltool_no_data(tmp_path: Path, *args: Any) -> Path:
     prov_folder = tmp_path / "provenance"
     prov_folder.mkdir()
-    new_args = ["--no-data", "--provenance", str(prov_folder)]
+    new_args = ["--enable-ext", "--no-data", "--provenance", str(prov_folder)]
```
mr-c, line 51: `--enable-ext` shouldn't be required when using `loadListing` with CWL v1.1+

PR #1721, opened 2022-08-29
Workflow checker: ignore 'null' when comparing arrays of types

mr-c:


cwltool/checker.py
```diff
@@ -81,7 +81,8 @@ def can_assign_src_to_sink(
     src: admissible source types
     sink: admissible sink types
 
-    In non-strict comparison, at least one source type must match one sink type.
+    In non-strict comparison, at least one source type must match one sink type,
+       except for 'null'.
```
tetron, line 85: It would be worth expanding this slightly, I think the intention here is to catch a common error, by emitting a warning or error when [X, null] goes to [Y, null] which is technically compatible because "null" could go to "null" but probably not what you intended?

mr-c: Thanks, with an expanded doc are you other otherwise okay with this?

tetron (approved):

tetron:
If you have ["null", File] to [File] that would previously produce a warning, we should make sure it is doing what we want.

Issue #1728, opened 2022-09-02
Provenance switch to cwl_utils.pack


jjkoehorst:
To improve the provenance of cwltool it was advices to switch to the cwl_utils.pack.

One of the "issues" we have internally is that the workflow files are all combined into a single folder which is ok but the cwl paths in the workflows that point to the individual (step) files are still in an old relative path form.

PR #1730, opened 2022-09-02
drop pack.py in favor of cwl_utils.pack

mr-c:
Only 7 test failures!

Aim to address #1728


mr-c:
Some of the test failures are due to slightly different behavior in cwl_utils.pack (which is derived from sbpack's `cwlpack`)
- it puts step processes inline instead of in a `$graph`.
- therefore the id fields are not prefixed with their original step names

Other test failures might be due to errors in `cwl_utils.pack` itself
- not including all `$schemas` in the parent `$schemas`

PR #1748, opened 2022-10-05
CWL inputs section to schema salad

mr-c:
- [ ] move preamble to static file (or public file that gets `$include`d?)
- [ ] tests


mr-c:
This was a fun experiment. As explained in https://github.com/common-workflow-lab/galaxy/issues/145#issuecomment-1281918942 it is no longer needed

Issue #1770, opened 2022-12-02
Unclear development requirements


sdruskat:
*Backround:* I was cloning this repository to implement podman pull support in cwltool (since [done my @mr-c](https://github.com/common-workflow-language/cwltool/pull/1769) directly :pray:).

## Expected Behavior

- Developing on WSL2 with Ubuntu, I expected that after following the instructions in `CONTRIBUTING.md`, no tests would fail.

- I would expect either 
    - `CONTRIBUTING.md` to include the requirements mentioned below (including how to fix the one test that still fails), **or** 
    - some way of installing the missing test dependencies (docker, udocker, SingularityCE) to be installed automatically through make, and the tests checking for their existence/runtime and failing with a respective error message if requirements aren't met, **or**
    - tests being skipped when the mentioned requirements aren't met (as I wouldn't want to force developers to create this environment).

## Actual Behavior

- After following the instructions in `CONTRIBUTING.md`, tests failed, and the test run hung at a specific test (`test_windows_warning.py`).
- After applying the partial solution below, there is still one test failing (`test_whoami`) with the stacktrace below.

**Partial solution:**

- Investigation showed that there are undocumented requirements for passing tests locally:
    1. A running Docker daemon. Running the daemon (`sudo dockerd`) caused some previously failing tests to pass, and the test run to complete (rather than hang at the above-mentioned test.
    2. A local installation of `udocker`. Installing `udocker` on the test system caused some previously failing tests to pass.
    3. A specific installation of `singularity`. Tests failed with `apptainer` (the Linux Foundation fork of the Singularity codebase that provides the `singularity` command) installed. Installing SingularityCE 3.10.4 (https://github.com/sylabs/singularity/releases/download/v3.10.4/singularity-ce_3.10.4-focal_amd64.deb) caused some previously failing tests to pass.

## Workflow Code

n/a

## Full Traceback
```pytb
(env) stephan@computer:~/src/cwltool$ make test
python --version 2>&1 | grep "Python 3"
Python 3.10.8
python -m pytest -rs
============================================================================================== test session starts ==============================================================================================
platform linux -- Python 3.10.8, pytest-7.2.0, pluggy-1.0.0
rootdir: /home/stephan/src/cwltool, configfile: tox.ini, testpaths: tests
plugins: xdist-3.0.2, mock-3.10.0, cov-4.0.0, httpserver-1.0.6
gw0 [642] / gw1 [642]
[...]

PR #1771, opened 2022-12-02
Workaround for users with an empty fullname

mr-c:
May fix part of #1770


mr-c:
@sdruskat can you give me the output of `python3 -c "import getpass; print(getpass.getuser()); import os; import pwd; print(pwd.getpwuid(os.getuid())[4])"` on your WSL2 system? I'll use that to improve the test

PR #1772, opened 2022-12-03
Move sys.exit

jfennick:
This PR simply moves the sys.exit() call so that users call call run() without exiting. I'm currently copy & pasting the body of run (without the sys.exit call of course), but that isn't ideal.

FYI the reason I'm interested in calling run() is because the normal way of using the cwltool [Python API](https://github.com/common-workflow-language/cwltool#import-as-a-module) does not (easily) allow using all of the CLI arguments. (In particular, --cachedir and --provenance)

### DIFF
--- a/cwltool/main.py
+++ b/cwltool/main.py
@@ -1511,10 +1511,9 @@ def run(*args: Any, **kwargs: Any) -> int:
     """Run cwltool."""
     windows_check()
     signal.signal(signal.SIGTERM, _signal_handler)
+    retval = 1
     try:
         retval = main(*args, **kwargs)
-    except:
-        retval = 1
     finally:
         _terminate_processes()
     return retval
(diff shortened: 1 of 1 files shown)


kinow (approved):

mr-c:
Thank you @jfennick ; I see that you are the main author of https://github.com/jfennick/workflow_inference_compiler ; will you be submitting a talk to the 2023 CWL Conference about that tool? I'd love to see it!

https://cwl.discourse.group/t/deadline-for-talk-proposals-is-monday-january-16th-2023/720

jfennick:
Glad you like it! FYI development is now moved to https://github.com/PolusAI/workflow-inference-compiler

Pending paperworks and approvals from my management (likely), then yeah I'll be happy to give a talk.

PR #1779, opened 2022-12-25
cwl v1.3.0-dev1 native loop support

GlassOfWhiskey:
This commit converts the logic introduced to handle the `cwltool:Loop` requirement extension to handle the new `loop` construct introduced in the CWL v1.3 spec.
A proper method in the `upgrade.py` module handles the automatic conversion from the old requirement to the new format.
The existing `test_loop.py` file is still useful to verify the correctness of the conversion process.

See also https://github.com/common-workflow-language/cwl-v1.3/pull/5


cwltool/update.py
```diff
@@ -20,6 +20,69 @@
 from .utils import CWLObjectType, CWLOutputType, aslist, visit_class, visit_field
 
 
+def v1_2to1_3(
```
mr-c, line 23: There is no CWL v1.3 yet, only `v1.3.0-dev1`

cwltool/update.py
```diff
@@ -231,13 +304,14 @@ def v1_2_0dev5to1_2(
...
     "v1.2.0-dev5": v1_2_0dev5to1_2,
+    "v1.3.0-dev1": v1_3_0dev1to1_3,
 }  # type: Dict[str, Optional[Callable[[CommentedMap, Loader, str], Tuple[CommentedMap, str]]]]
 
 
 ALLUPDATES = UPDATES.copy()
 ALLUPDATES.update(DEVUPDATES)
 
-INTERNAL_VERSION = "v1.2"
+INTERNAL_VERSION = "v1.3"
```
mr-c, line 314:
```suggestion
INTERNAL_VERSION = "v1.3.0-dev1"
```

See the https://github.com/common-workflow-language/cwltool/blob/f0a844af6f93883fcacb3572a6366124a50d0909/cwltool/update.py#LL222

mr-c (changes requested):
Huzzah!

You'll also need to add a dump of the `v1.3.0-dev1` schema to https://github.com/common-workflow-language/cwltool/tree/cwl-v1.3/cwltool/schemas

cwltool/update.py
```diff
@@ -11,15 +11,61 @@
...
+                    list(cast(MutableSequence[CWLObjectType], s["requirements"]))
+                ):
+                    cls = cast(str, r["class"])
+                    if cls == "http://commonwl.org/cwltool#Loop":
+                        if "when" in s:
+                            raise SourceLine(s, "when", ValidationException).makeError(
+                                "The `cwltool:Loop` clause is not compatible with the `when` directive."
+                            )
+                        if "loopWhen" not in r:
+                            raise SourceLine(
```
mr-c, line 43: Test case for this, please

cwltool/update.py
```diff
@@ -54,34 +100,17 @@ def v1_0to1_1(
...
-                    cls = cast(str, r["class"])
-                    if cls in rewrite:
-                        r["class"] = rewrite[cls]
-                else:
-                    raise ValidationException(
-                        f"hints entries must be dictionaries: {type(r)} {r}."
-                    )
+                cls = cast(str, r["class"])
+                if cls in rewrite:
+                    r["class"] = rewrite[cls]
```
mr-c, line 110: This line has no test coverage.

mr-c (changes requested):

tetron:
With the fix in https://github.com/common-workflow-language/schema_salad/pull/846 this passes unit tests for me.

mr-c:
I didn't think you meant to close this PR @tetron 🙂

tetron:
> I didn't think you meant to close this PR @tetron 🙂

github seems to have done it automatically :frowning_face:

mr-c:
@tetron I think those may be due to recent changes to `cwltest`:  https://github.com/common-workflow-language/cwltest/issues/216

Did you see my email about the cofest?

tetron:
Ok!  So we're just down to the test coverage check.

tetron:
I keep fiddling with this but I think we should just override the codecov check because at least one of the lines it is complaining about appears to be almost impossible to trigger in normal usage because it is an exception handler and "normal" exceptions get routed another way, so it only engages if a very truly unexpected case happens.  Possibly we could mock something but I'm pretty sure at this point the cost/benefit of trying to make codecov happy for these corner cases is very bad.

I haven't yet managed to determine if there's a way to tell codecov to ignore certain lines.

mr-c (approved):

PR #1782, opened 2023-01-06
Add jobname to cached folders.

bdw2292:
Each cached folder is a hex string of characters that does not indicate which job is being run. Since we would like pseudo real time information, we cannot use the –provenance flag to look at the provenance/workflow/primary-output.json content since this is only produced at the very end of the work flow and we would like real time updates. We have explored other options such as capturing the standard output of the logger which outputs the job and cachedir, however this is error prone especially if running many work flows through cwltool in parallel. We are also building features in our own tools that depend on cwltool that will rely upon knowing information in real time about the jobname corresponding to each cached directory. Thus, we think this is the best solution.

### DIFF
--- a/cwltool/process.py
+++ b/cwltool/process.py
@@ -58,7 +58,7 @@ from .mpi import MPIRequirementName
 from .pathmapper import MapperEnt, PathMapper
 from .secrets import SecretStore
 from .stdfsaccess import StdFsAccess
-from .update import INTERNAL_VERSION, ORIGINAL_CWLVERSION
+from .update import INTERNAL_VERSION, ORDERED_VERSIONS, ORIGINAL_CWLVERSION
 from .utils import (
     CWLObjectType,
     CWLOutputAtomType,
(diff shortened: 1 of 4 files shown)


mr-c:
> Each cached folder is a hex string of characters that does not indicate which job is being run. Since we would like pseudo real time information, we cannot use the –provenance flag to look at the provenance/workflow/primary-output.json content since this is only produced at the very end of the work flow and we would like real time updates. We have explored other options such as capturing the standard output of the logger which outputs the job and cachedir, however this is error prone especially if running many work flows through cwltool in parallel. We are also building features in our own tools that depend on cwltool that will rely upon knowing information in real time about the jobname corresponding to each cached directory. Thus, we think this is the best solution.

Hello @bdw2292 , and thanks for opening this.

Adding a file into the cache directory will result in `outputBinding: { glob: "*" }` catching that additional file.

I would suggest directly targeting your core needs with a new reporting stream: either a file or socket to give you the information you really need. Maybe you could join us on the CWL call today to chat about your needs?

The next call is in ~5 hours: <https://www.commonwl.org/community/#cwl-online-community-meetings>

PR #1784, opened 2023-01-12
use device indices for docker --gpus

jfennick:
The logic for supporting cuda seems to mix up the number of devices (the 'count') and the device indices in several places. In particular, when running with docker `--gpus=1` is passed, which doesn't work on a machine with only a single gpu (i.e. device index 0). This might not be noticed when running on a machine with two or more gpus. This PR fixes the bug and otherwise leaves the API unchanged.

More generally, I think it would be better to expose the device indices directly, i.e. rename `cudaDeviceCountMin` and `cudaDeviceCountMax` to `cudaDeviceIndexMin` and `cudaDeviceIndexMax`. This would be very useful when combining scattering on a machine with multiple gpus, where it can be faster to restrict each calculation to a single gpu and cycle through them in a round robin fashion.

### DIFF
--- a/cwltool/docker.py
+++ b/cwltool/docker.py
@@ -399,7 +399,9 @@ class DockerCommandLineJob(ContainerCommandLineJob):
             runtime.append("--rm")
 
         if self.builder.resources.get("cudaDeviceCount"):
-            runtime.append("--gpus=" + str(self.builder.resources["cudaDeviceCount"]))
+            count = int(self.builder.resources["cudaDeviceCount"])
+            indices = ",".join([str(x) for x in range(count)])
+            runtime.append("--gpus=" + indices)
 
         cidfile_path = None  # type: Optional[str]
         # add parameters to docker to write a container ID file
--- a/cwltool/process.py
+++ b/cwltool/process.py
@@ -991,8 +991,8 @@ hints:
 
         cudaReq, _ = self.get_requirement("http://commonwl.org/cwltool#CUDARequirement")
         if cudaReq:
-            request["cudaDeviceCountMin"] = 1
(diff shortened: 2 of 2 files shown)


tetron:
I don't think this is right.  See here:

https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/user-guide.html#gpu-enumeration

Confusingly, it starts by describing NVIDIA_VISIBLE_DEVICES which is not actually the same as the syntax of `--gpus`

However the 3rd example demonstrates requesting 2 GPUS with `--gpus 2` without specifying specific devices, I believe that if you just request a number of GPUs, docker/nvidia daemon manage the resource allocation for you.  In my experience Docker also remaps the host devices inside the container, so `/dev/nvidia3` on the host shows up as `/dev/nvidia0` inside the container.

The 4th example shows requesting specific devices with the syntax `--gpus 'devices=1,2'`.  This is what you are suggesting, but your pull request isn't using it.

However, in general, the CWL philosophy is to avoid specific resources within the workflow to maintain portability.  If you specify that a workflow must use driver devices 5,6,7 then it won't be portable to an instance that only has 4 devices even if that node otherwise would have been able to run the job.

If you tried to use the existing GPU support (without the changes in this PR) and it didn't work or didn't fit your use case, you should file a bug and we should talk about it there.

jfennick:
I thought I had read that documentation before carefully, but it seems you are correct. I also take your point on CWL portability, even if it affects some use cases. In general, I try to be very careful to test locally before submitting PRs. I thought for sure I had a minimal example demonstrating a bug, but after more carefully checking and double checking again, my example seems to work with both --gpus 0 and 1.  Sorry for the noise, closing.

PR #1786, opened 2023-01-19
Keep track of which files have already been parsed

tetron:
This is to avoid re-parsing when the workflow refers to multiple fragments in the same file.


tetron:
Oops this needs https://github.com/common-workflow-language/cwl-utils/pull/188 to be merged and released first

mr-c (approved):

PR #1787, opened 2023-01-24
[doc] Add autolabels so intersphinx can link to anchors like index#install

kinow:
Related to https://github.com/common-workflow-language/user_guide/issues/334

The HTML build of Sphinx creates an inverted list of links (like a hashmap? or a dict), which can be used by extensions like `intersphinx` for linking between Sphinx sites.

In the issue linked, it is not possible to link to the cwltool installation anchor section, as it's not included in the inverted objects list (test with `python -m sphinx.ext.intersphinx https://cwltool.readthedocs.io/en/latest/objects.inv`).

With this change, the anchor items must be unique (I believe they are all unique at the moment, no errors after this change), and they are included in the `.inv` file during the HTML build, and the linked PR can successfully use it as a link.

Cheers
Bruno

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,6 @@ value
 
 # Folder created when using make
 cwltool_deps
+docs/_build/
+docs/autoapi/
+
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -38,6 +38,7 @@ author = "Peter Amstutz and Common Workflow Language Project contributors"
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
+    "sphinx.ext.autosectionlabel",
     "sphinx.ext.intersphinx",
     "sphinx.ext.inheritance_diagram",
     "autoapi.extension",
@@ -45,6 +46,8 @@ extensions = [
     "sphinx_rtd_theme",
     "sphinxcontrib.autoprogram",
 ]
+
+autosectionlabel_prefix_document = True
 
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3", None),


.gitignore
```diff
@@ -55,3 +55,6 @@ value
 
 # Folder created when using make
 cwltool_deps
+docs/_build/
+docs/autoapi/
```
kinow, line 59: Annoyed by these files that appeared in my `git status` while testing this PR. Happy to drop this if it's better :+1:

mr-c (commented):
This adds over 100 warnings about duplicate labels; maybe it needs to be configured to not touch the autoapi pages?

https://readthedocs.org/projects/cwltool/builds/19256779/

kinow:
> This adds over 100 warnings about duplicate labels; maybe it needs to be configured to not touch the autoapi pages?
> 
> https://readthedocs.org/projects/cwltool/builds/19256779/

Ah, spoke too fast, sorry @mr-c. Was looking at the user_guide & cwltool logs and I think I didn't see these warnings. Let me see what to do about them...

kinow:
Phew... heaps of warnings (cc @Mackenzie-OO7 see?! I told that happens a lot with me :laughing: ). 

>maybe it needs to be configured to not touch the autoapi pages?

Thanks for the suggestion @mr-c, sounds like the simplest approach (or find another way to create intersphinx links without the autolabel).

docs/conf.py
```diff
@@ -46,6 +47,8 @@
     "sphinxcontrib.autoprogram",
 ]
 
+autosectionlabel_prefix_document = True
```
kinow, line 50: This adds prefixes to each label, using the document name, reducing risks of conflicts.

https://www.sphinx-doc.org/en/master/usage/extensions/autosectionlabel.html

mr-c (approved):

PR #1798, opened 2023-02-02
CWLProv: add directory basename to the provenance metadata

simleo:
Changes the code to add a "cwlprov:basename" entry for directories (currently it is only added for files).

This is needed to support reproducibility when directories need to follow a naming pattern (e.g., from `secondaryFiles`).

### DIFF
--- a/cwltool/provenance_profile.py
+++ b/cwltool/provenance_profile.py
@@ -386,6 +386,10 @@ class ProvenanceProfile:
                 (PROV_TYPE, RO["Folder"]),
             ],
         )
+
+        if "basename" in value:
+            coll.add_attributes({CWLPROV["basename"]: cast(str, value["basename"])})
+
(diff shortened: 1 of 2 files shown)


mr-c (approved):

mr-c:
Thank you @simleo ! What is the urgency level to get this released?

simleo:
In theory the RO bundle could be regenerated with cwltool installed from the main branch, but it would be better to be able to point to a release. I see there's been a release right before this PR. Would another release in a week or so be ok?

simleo:
@mr-c In theory the RO bundle could be regenerated with cwltool installed from the main branch, but it would be better to be able to point to a release. I see there's been a release right before this PR. Would another release in a week or so be ok?

mr-c:
I'll release again next week, no problem. Have a great weekend!

PR #1807, opened 2023-02-14
ga: colorful output

mr-c:


kinow (approved):
:rocket: nice! Confirmed output of jobs in this PR have colors, while the same jobs in previous PR's did not.

image

image

mr-c:
@kinow Thanks! I wish this was the default everywhere. We have so many repos!

kinow:
> @kinow Thanks! I wish this was the default everywhere. We have so many repos!

Hmmm, GH Actions at organization levels seem to focus more on permissions/security at the moment… however, what about [required workflows](https://docs.github.com/en/actions/using-workflows/required-workflows), @mr-c? Wouldn't a required workflow run for every repository as a check? And in that case, creating an `env` in that required workflow... would it apply variables to every other workflow?

Issue #1806, opened 2023-02-14
Bad URN-scheme for `urn:hash::sha1:`


MBueschelberger:
The uri for aggregates in the provenance graph is produced by [this](https://github.com/common-workflow-language/cwltool/blob/1969a654b7d1ea4f21f6ee7b45b611206c1e2ab3/cwltool/provenance.py#L520) hardcoded prefix.

According to the [generic uri-scheme](https://www.rfc-editor.org/rfc/rfc2141), double colons are not compliant to the urn-syntax.

Is there any reason for the double colon? If this was coded by accident, I would like to request a change to `urn:hash:sha1:` as prefix.

* cwltool version: 
[3.1.20230213100550](https://github.com/common-workflow-language/cwltool/releases/tag/3.1.20230213100550)

mr-c:
Tagging @stain

https://github.com/common-workflow-language/cwltool/commit/973cff2168ef1d8f868f66c550e61e9f5bd93f54

PR #1808, opened 2023-02-14
Update namespace in urn

MBueschelberger:
Related to #1806, the URN-double colon is updated to a single colon in order to be fully compliant to the [urn-scheme](https://www.rfc-editor.org/rfc/rfc2141)

### DIFF
--- a/cwltool/provenance.py
+++ b/cwltool/provenance.py
@@ -517,7 +517,7 @@ class ResearchObject:
             # TODO: Change to nih:sha-256; hashes
             #  https://tools.ietf.org/html/rfc6920#section-7
             aggregate_dict = {
-                "uri": "urn:hash::sha1:" + filename,
+                "uri": "urn:hash:sha1:" + filename,
                 "bundledAs": {
                     # The arcp URI is suitable ORE proxy; local to this Research Object.
                     # (as long as we don't also aggregate it by relative path!)


kinow (commented):
Hi @MBueschelberger 

Approved the GH Actions workflow to run. I don't know much about the urn spec or syntax, but searching the `common-workflow-language`, there are many other occurrences (55 “code“ occurrences):

https://github.com/search?q=org%3Acommon-workflow-language+urn%3Ahash%3A%3Asha1%3A&type=code

Also in this IETF doc: https://datatracker.ietf.org/doc/id/draft-thiemann-hash-urn-00.txt

Reading the document you linked in the issue linked here, I would expect `:sha1`, so no idea why we have the `::sha1`, nor why this document above also uses the same two colons. Strange.

image

stain:
If we're using the `urn:hash` scheme then we should follow its [recommendations](https://datatracker.ietf.org/doc/id/draft-thiemann-hash-urn-01.txt), which in this case has a `::` as the media type is unknown. Unfortunately his is [stuck](https://datatracker.ietf.org/doc/draft-thiemann-hash-urn/) in 'draft' stage (now expired).

It is as far as I can tell however compatible with the older [RFC2141](https://www.rfc-editor.org/rfc/rfc2141#section-2.2) you linked to.

```
<URN> ::= "urn:" <NID> ":" <NSS>
<NID>         ::= <let-num> [ 1,31<let-num-hyp> ]
<NSS>         ::= 1*<URN chars>
```

Substituting the example `urn:hash::sha1:LBPI666ED2QSWVD3VSO5BG5R54TE22QL` we get:

```
NID = "hash"
NSS = ":sha1:LBPI666ED2QSWVD3VSO5BG5R54TE22QL"
```

Now the question is if `":"` is permitted in `<URN chars>`:

```
<URN chars>   ::= <trans> | "%" <hex> <hex>
<trans>       ::= <upper> | <lower> | <number> | <other> | <reserved>
<other>       ::= "(" | ")" | "+" | "," | "-" | "." |
                     ":" | "=" | "@" | ";" | "$" |
                     "_" | "!" | "*" | "'"
```

And there it is in second line of `<other>`.

Now in newer [RFC8141](https://datatracker.ietf.org/doc/html/rfc8141) the rules are different:

```
assigned-name = "urn" ":" NID ":" NSS
NID           = (alphanum) 0*30(ldh) (alphanum)
ldh           = alphanum / "-"
NSS           = pchar *(pchar / "/")
```

We assign `NID` and `NSS` the same, and `hash` is still valid NID.  Now the question is if the NSS can start with `:` from `pchar`?

```
      pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
```

So `":"` is declared directly, so it should also be valid.  

Therefore I don't see the need for this pull request as it is breaking the namespace which was already valid before, if a bit odd-looking (but then so are all the URNs!  :-) ) 

The real fix should be to get rid of insecure SHA1 all together and use sha256 or better, which then can use the newer [Naming things with hashes](https://www.rfc-editor.org/rfc/rfc6920.html) scheme (RFC 6920), which deliberately do not support SHA1 and has a registry of modern checksum algorithms.

stain:
> [](https://datatracker.ietf.org/doc/id/draft-thiemann-hash-urn-00.txt)
> 
> Reading the document you linked in the issue linked here, I would expect `:sha1`, so no idea why we have the `::sha1`, nor why this document above also uses the same two colons. Strange.

You are thinking of the separate `urn:sha1` scheme which is not well defined.

mr-c:
@MBueschelberger Can you help us by sharing your motivation for this change?

MBueschelberger:
Hi @mr-c,

I actually requested this change since I was not able to parse the graph with any of the available RDFLib-versions. Removing the second colon actually solved the problem for me. 

This is why I actually initiated this discussion here about the origin of this `::`.

Best regards

mr-c:
@simleo Do you have experience using rdflib to parse CWL prov aggregates that contain `urn:hash::sha1` (note the double colon)?

https://github.com/common-workflow-language/cwltool/pull/1808/files#diff-2bb87b99f4b0d10faa69f7bcec12f404bc4a41ac1aae63f8a14d6460fa5798a5L520

simleo:
@mr-c no, sorry. In runcrate, I use [cwlprov-py](https://github.com/common-workflow-language/cwlprov-py) to read CWLProv ROs, but I don't use rdflib directly.

PR #1822, opened 2023-03-24
guard `import pwd` for windows

jfennick:
Running CWL on windows currently requires Windows Subsystem for Linux, but due to this import statement, windows users who have not yet installed WSL (or cannot install WSL due to administrator privileges) cannot even import most modules.

Since pwd is only used in provenance.py on one line https://github.com/common-workflow-language/cwltool/blob/main/cwltool/provenance.py#L80, guarding the `import pwd` will allow all other use cases that don't involve running with provenance.

### DIFF
--- a/cwltool/builder.py
+++ b/cwltool/builder.py
@@ -21,12 +21,13 @@ from cwl_utils import expression
 from cwl_utils.file_formats import check_format
 from mypy_extensions import mypyc_attr
 from rdflib import Graph
-from ruamel.yaml.comments import CommentedMap
 from schema_salad.avro.schema import Names, Schema, make_avsc_object
 from schema_salad.exceptions import ValidationException
 from schema_salad.sourceline import SourceLine
 from schema_salad.utils import convert_to_dict, json_dumps
 from schema_salad.validate import validate
+
+from ruamel.yaml.comments import CommentedMap
 
 from .errors import WorkflowException
 from .loghandler import _logger
--- a/cwltool/command_line_tool.py
+++ b/cwltool/command_line_tool.py
@@ -33,13 +33,14 @@ from typing import (
 
 import shellescape
 from mypy_extensions import mypyc_attr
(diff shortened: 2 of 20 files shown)


jfennick:
Looks like the CI didn't re-run when I force-pushed a formatting fix.

mr-c:
Hey @jfennick . We explicitly stopped supporting running `cwltool` on Microsoft Windows systems outside of WSL several years ago.

Even with this change, you should see a warning about that

jfennick:
Yes, I'm aware of that. This PR is not necessarily about running on windows without WSL, it's just about being able to import on windows without WSL. For example, if I `pip install cwltool` in a brand new conda environment, I get this error before the windows_check() function can be called.

```
(test) C:\Users\fennickjr\>cwltool
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\fennickjr\mambaforge-pypy3\envs\test\Scripts\cwltool.exe\__main__.py", line 4, in <module>
  File "C:\Users\fennickjr\mambaforge-pypy3\envs\test\Lib\site-packages\cwltool\main.py", line 87, in <module>
    from .provenance import ResearchObject, WritableBagFile
  File "C:\Users\fennickjr\mambaforge-pypy3\envs\test\Lib\site-packages\cwltool\provenance.py", line 11, in <module>
    import pwd
ModuleNotFoundError: No module named 'pwd'
```

With this PR, the warning is correctly printed to stdout.

jfennick:
That said, I'm tentatively suggesting I may be willing to restore windows support without WSL. I'm not saying you need to officially support it, but if you're willing to accept PRs for unofficial support, that would be nice. WSL is fine, but not everyone can install it, and even if you can it makes writing robust install scripts a bit harder.

mr-c (approved):

Issue #1883, opened 2023-08-12
Deadlock with podman (at least 4.5.0) and simple workflow


jmfernandez:
A simple workflow which works both in docker and in singularity mode (example)

```bash
cwltool --singularity https://raw.githubusercontent.com/inab/hello-workflows/227d4b7b0c925b3baf8241fecd3dd29b1d586a41/cwl/hello-workflow.cwl https://raw.githubusercontent.com/inab/hello-workflows/227d4b7b0c925b3baf8241fecd3dd29b1d586a41/cwl/hello.yml
```

 hangs forever in podman mode

```bash
cwltool --podman https://raw.githubusercontent.com/inab/hello-workflows/227d4b7b0c925b3baf8241fecd3dd29b1d586a41/cwl/hello-workflow.cwl https://raw.githubusercontent.com/inab/hello-workflows/227d4b7b0c925b3baf8241fecd3dd29b1d586a41/cwl/hello.yml
```

## Expected Behavior

The workflow, when it is run in podman mode, should finish almost immediately. 

## Actual Behavior

The execution hangs forever, so it has to be killed through Ctrl+C or other means.

## Workflow Code

It is available at https://github.com/inab/hello-workflows/tree/227d4b7b0c925b3baf8241fecd3dd29b1d586a41/cwl

## Full Traceback

This is what appears running in debug mode and having to kill `cwltool` through Ctrl+C
```pytb
INFO /home/jmfernandez/projects/hello-workflows/cwl/.cwlenv/bin/cwltool 3.1.20230719185429
DEBUG [workflow ] initialized from https://raw.githubusercontent.com/inab/hello-workflows/227d4b7b0c925b3baf8241fecd3dd29b1d586a41/cwl/hello-workflow.cwl#cosifer_workflow
INFO [workflow ] start
DEBUG [workflow ] inputs {
    "an_input": {
        "class": "File",
        "location": "https://raw.githubusercontent.com/etc/passwd",
        "basename": "passwd",
        "nameroot": "passwd",
        "nameext": ""
    }
}
INFO [workflow ] starting step hello
DEBUG [step hello] job input {
    "https://raw.githubusercontent.com/inab/hello-workflows/227d4b7b0c925b3baf8241fecd3dd29b1d586a41/cwl/hello-workflow.cwl#cosifer_workflow/hello/an_input": {
        "class": "File",
        "location": "https://raw.githubusercontent.com/etc/passwd",
        "basename": "passwd",
[...]

PR #1890, opened 2023-08-21
Fixed race condition which happens when a job runs "too fast",

jmfernandez:
When a podman process finishes even before reaching the monitoring method, a deadlock happens, as no one is updating `process.returncode` and spawned process is in zombie state (so, no signal is sent).

This fix adds a `process.poll()` call, so it gives the chance to fill in `process.returncode`.

Fixes: #1883


mr-c:
Thank you @jmfernandez !

Issue #1884, opened 2023-08-12
cwltool does not properly detect available podman containers


jmfernandez:
Next workflow https://github.com/inab/hello-workflows/tree/227d4b7b0c925b3baf8241fecd3dd29b1d586a41/cwl depends on container [bash:5.2.15](https://hub.docker.com/_/bash/tags). The execution of it in podman mode with `--disable-pull` tag does not work as expected when the container is already in the local cache.

For instance, if you run
```bash
podman pull bash:5.2.15
```
it either fetches or updates the container with that tag, having it cached for its later usage. But it is also true that podman lists the available containers with a fully qualified tag:

```bash
podman images
```
```
REPOSITORY                             TAG            IMAGE ID      CREATED       SIZE
docker.io/library/bash                 5.2.15         3c04497fad88  4 days ago    14.4 MB
```

so internal `cwltool` code is failing to find the container as it relies on the output of `podman images --no-trunc --all`

https://github.com/common-workflow-language/cwltool/blob/2fedbd61e3bbb18a6288c71d8b273201049c31af/cwltool/docker.py#L116-L145

Maybe you could use a more targeted method, like using `podman inspect bash:5.2.15`, in order to learn whether the tag resolves to something in podman cache or not.

## Expected Behavior

Once the container is in the local podman cache, running next command, where pulls are disabled, should work:

```bash
cwltool --debug --disable-pull --podman https://raw.githubusercontent.com/inab/hello-workflows/227d4b7b0c925b3baf8241fecd3dd29b1d586a41/cwl/hello-workflow.cwl https://raw.githubusercontent.com/inab/hello-workflows/227d4b7b0c925b3baf8241fecd3dd29b1d586a41/cwl/hello.yml
```

## Actual Behavior

`cwltool` complies about not finding the container image (see dump below)

## Workflow Code

Workflow is available at https://github.com/inab/hello-workflows/tree/227d4b7b0c925b3baf8241fecd3dd29b1d586a41/cwl

## Full Traceback
```pytb
cwltool --debug --disable-pull --podman https://raw.githubusercontent.com/inab/hello-workflows/227d4b7b0c925b3baf8241fecd3dd29b1d586a41/cwl/hello-workflow.cwl https://raw.githubusercontent.com/inab/hello-workflows/227d4b7b0c925b3baf8241fecd3dd29b1d586a41/cwl/hello.yml 
[...]

PR #1891, opened 2023-08-21
Propery locate podman images

jmfernandez:
Use docker/podman/udocker inspect to check container images are already in the local registry.

Additionally, removed superfluous udocker/podman pull, which was done even when `--disable-pull` parameter is passed.

This should fix issue #1884

### DIFF
--- a/cwltool/docker.py
+++ b/cwltool/docker.py
@@ -2,6 +2,7 @@
 
 import csv
 import datetime
+import json
 import math
 import os
 import re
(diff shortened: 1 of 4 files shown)


cwltool/docker.py
```diff
@@ -113,35 +114,18 @@ def get_image(
...
-            except ValueError:
+                manifest = json.loads(
+                    subprocess.check_output(
+                        [self.docker_exec, "inspect", docker_image_id]
+                    ).decode(  # nosec
+                        "utf-8"
+                    )
+                )
+                found = True
+            except Exception as e:
```
mr-c, line 128: Can a narrower Exception be used here?

jmfernandez: Good question. I'm sure that if the command is not found, it raises a FileNotFoundError and a CalledProcessError when the exit value is not 0 (i.e. when docker/podman does not find the tag in the local registry). I have to study whether there are other possible error cases.

jmfernandez: I have just added a new commit which narrows the exceptions, rebased the branch to gather latest commits from `main`, and pushed it to my GitHub branch.

jmfernandez:
I have just seen at https://github.com/common-workflow-language/cwltool/actions/runs/5937318373/job/16099520096?pr=1891 that there are several tests which are not passed now, due the error messages are apparently different. I have tried to reproduce it from command line:

```bash
```

mr-c:
@jmfernandez 
See https://github.com/common-workflow-language/cwltool/actions/runs/5937318373/job/16099520096?pr=1891#step:8:96

The test is looking for the exact phrase `or quaquioN is missing or broken` (the old behavior) but now we get `'quaquioN' not found: [Errno 2] No such file or directory: 'quaquioN'` (which is more useful, thanks!)

jmfernandez:
I still don't understand the reason of the "irreproducibility" of the test failure. Along today I got the test failure at least twice when all the tests were run, but not always. And never when only the failing test is run (through `tox -- "-k test_bad_userspace_runtime"`).

In any case, I have updated the failing test case, but I'm afraid it is not the real workaround / solution.

jmfernandez:
The failing test shows that there seems to be some kind of unpredictable slowness when `udocker` is used. As you can see at https://github.com/common-workflow-language/cwltool/actions/runs/5962544324/job/16174584481?pr=1891 a timeout is raised, but it doesn't happen at https://github.com/common-workflow-language/cwltool/actions/runs/5962544324/job/16174585196?pr=1891

When I was debugging the source of this condition I was tracing the source. Next line calls cwltool code:

https://github.com/common-workflow-language/cwltool/blob/bd89c5694685bff46bf56fb32316c8f6fe0d799d/tests/util.py#L108

which ends leading to next code snippet at `_job_popen`:

https://github.com/common-workflow-language/cwltool/blob/bd89c5694685bff46bf56fb32316c8f6fe0d799d/cwltool/job.py#L967-L985

I detected it was remaining stuck at line 984, whose monitoring body for this use case is at

https://github.com/common-workflow-language/cwltool/blob/bd89c5694685bff46bf56fb32316c8f6fe0d799d/cwltool/job.py#L507-L537

and then it is waiting just at line 528 until the timeout happens

https://github.com/common-workflow-language/cwltool/blob/bd89c5694685bff46bf56fb32316c8f6fe0d799d/cwltool/job.py#L528

I did a test where I removed the process termination, and it finished properly later than the timelimit. So, I raised the `timelimit` at:

https://github.com/common-workflow-language/cwltool/blob/bd89c5694685bff46bf56fb32316c8f6fe0d799d/tests/wf/timelimit.cwl#L14

to 20, and it started working!

jmfernandez:
Also, I realized that the max memory usage monitoring being done was happening only once, after the first second of the process running. And that both the `timelimit` and monitoring features were not implemented for the cases where a shelled popen was used.

mr-c (approved):
Thank you @jmfernandez !

PR #1914, opened 2023-09-27
guard `import tcntl` for windows

jfennick:
This PR is essentially identical to #1822 It simply guards the `import fcntl` on windows so that the proper error message can be displayed.


jfennick:
FYI build_test_container is failing due to an (apparently) unrelated error with schema-salad:
`schema-salad ... has inconsistent version: expected '8.4.20230426093816', but metadata has '8.4'`
https://github.com/common-workflow-language/cwltool/actions/runs/6328208603/job/17185819443?pr=1914#step:3:263

mr-c (approved):

PR #1938, opened 2023-11-07
add singularityFile

misterbrandonwalker:
Add ability to use `dockerFile` with singularity. 

1. Convert dockerFile into singularity recipe via sypthon.
2. Then use singularity build to build image


mr-c (commented):
Thank you @misterbrandonwalker !

Can you add a test?

cwltool/singularity.py
```diff
@@ -153,6 +157,42 @@ def get_image(
...
+                with open(dockerfile_path, "w") as dfile:
+                    dfile.write(dockerRequirement["dockerFile"])
+                cmd = [
+                    "spython",
+                    "recipe",
+                    dockerfile_path,
+                    singularityfile_path,
+                ]
+                _logger.info(str(cmd))
+                check_call(cmd, stdout=sys.stderr)  # nosec
```
mr-c, line 186: Can you use the `spython` API instead of the command line? That would be safer
https://singularityhub.github.io/singularity-cli/recipes#parsers

mr-c (changes requested):

cwltool/singularity.py
misterbrandonwalker: Thank you! Just switched and pushed :)

cwltool/singularity.py
```diff
@@ -153,6 +158,42 @@ def get_image(
...
+                    file.write(result)
+
+                cmd = [
+                    "singularity",
+                    "build",
+                    image_path,
+                    singularityfile_path,
+                ]
+                _logger.info(str(cmd))
+                check_call(cmd, stdout=sys.stderr, env=custom_env)
```
mr-c, line 195: Thanks for switching to the API for the conversion. Can you also use the API for building?

https://singularityhub.github.io/singularity-cli/commands-images#build

misterbrandonwalker: Hey, so the issue is that the Client.build requires sudo access to run on our cluster :/ . The singuliarty used in cli is installed system wide (not local copy etc) and so its able to build.

cwltool/singularity.py
```diff
@@ -185,6 +186,8 @@ def get_image(
                 with open(singularityfile_path, 'w') as file:
                     file.write(result)
 
+                # Client.build(recipe=singularityfile_path, build_folder=absolute_path, options=["--fakeroot"])
```
mr-c, line 189: Did you try with `sudo=False`?

cwltool/singularity.py
mr-c: huh, `spython` bundles its own client?

cwltool/singularity.py
misterbrandonwalker: Excellent, that works great!

cwltool/singularity.py
misterbrandonwalker: Yes sir

misterbrandonwalker:
The test passes for me with tox I think the issue now is with how I am using current working directory rather than /tmp I see permissions error on the runner
https://github.com/common-workflow-language/cwltool/actions/runs/6932721232/job/18861573891#step:8:1696

However I added to comment here if I try to use /tmp to execute singularity it will crash
https://github.com/common-workflow-language/cwltool/compare/main...misterbrandonwalker:cwltool:singularityFile#diff-2cbd89e23f6e4c3f1f7df787bc1ebf9f32408c48b13125b70c6c31111a119b1aR173

cwltool/singularity.py
```diff
@@ -143,6 +146,9 @@ def get_image(
         """
         found = False
 
+        if "dockerImageId" not in dockerRequirement and "dockerPull" in dockerRequirement:
+            dockerRequirement["dockerImageId"] = dockerRequirement["dockerPull"]
+
```
mr-c, line 151: This is breaking `dockerPull`; if you need this only for `dockerFile`, then set that there

https://github.com/common-workflow-language/cwltool/actions/runs/6947078679/job/19231705078?pr=1938#step:8:165

(missing the `docker://` prefix set at https://github.com/common-workflow-language/cwltool/pull/1938/files#diff-dc5ddffad347b39baf2c837e6f60869433318cc3023bb81bc90e7b480da6e34fR198 because `dockerImageId` was set here)

mr-c (approved):
Thank you @misterbrandonwalker !

PR #1955, opened 2023-12-08
remove galaxy-tool-util from build dependency

jfennick:
Some of the build dependencies appear to be regular dependencies, and do not appear to be required at build time.
This may cause issues in certain specific situations. See this Dockerfile for a minimal reproducible example.
For now, this PR simply removes galaxy-tool-util from pyproject.toml (only).

```
FROM condaforge/miniforge3

# This works
# RUN conda install "python<3.10"

# Have to downgrade python to install pypy
RUN conda install "python<3.10" pypy

# Install binary system dependencies through conda (so we don't have to install compilers)
RUN conda install psutil

# This works
# RUN pip install cwltool

RUN conda install git
# Due to listing galaxy-tool-util as a build dependency,
# pip attempts to install conda-package-streaming and then zstandard from source and fails with:
# No such file or directory: 'cc'
RUN pip install cwltool@git+https://github.com/common-workflow-language/cwltool.git
```

### DIFF
--- a/cwltool.Dockerfile
+++ b/cwltool.Dockerfile
@@ -4,7 +4,7 @@ RUN apk add --no-cache git gcc python3-dev libxml2-dev libxslt-dev libc-dev linu
 
 WORKDIR /cwltool
 COPY . .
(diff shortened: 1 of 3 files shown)


mr-c (approved):
Sure. It was required for typing reasons, but the cast seem to be sufficient.

PR #1956, opened 2023-12-11
rename singularity image

misterbrandonwalker:
If user adds `dockerImageId` along with `dockerFile` , then rename the output image from spython Client.build accordingly

### DIFF
--- a/cwltool/singularity.py
+++ b/cwltool/singularity.py
@@ -3,6 +3,7 @@
 import logging
 import os
 import os.path
+from pathlib import Path
 import re
 import shutil
 import sys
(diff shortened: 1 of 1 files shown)


cwltool/singularity.py
```diff
@@ -171,29 +172,43 @@ def get_image(
...
+                Client.build(
+                    recipe=singularityfile_path,
+                    build_folder=absolute_path,
+                    sudo=False,
+                    options=singularity_options,
+                )
+                if "dockerImageId" in dockerRequirement:
+                    image_name = dockerRequirement["dockerImageId"]
+                    children = sorted(Path(absolute_path).glob("*.sif"))
+                    image_path = children[0]
```
mr-c, line 206: This presumes that there are no `.sif` files in the cache already; why not pass the `image_name` to the `image` parameter of `Client.build` to avoid the need for a rename?

https://singularityhub.github.io/singularity-cli/api/source/spython.main.base.html#spython.main.base.Client.build

mr-c (approved):

mr-c:
Thank you @misterbrandonwalker !

Issue #1961, opened 2023-12-29
cwltool hangs running fast jobs in Podman


kuanyili:
#1883 and #1890 tried to address this issue, but it is still not solved.

## Cause Analysis

The behavior of how Podman handles cidfile changed since 4.4.0.

https://github.com/containers/podman/commit/3fee351c355788aed0d5f9e069f8b14ed7eb109a

Container ID file is now removed along with the container by Podman.

https://docs.podman.io/en/latest/markdown/podman-run.1.html#cidfile-file

So on a fast machine with Podman >= 4.4.0, there is a chance that we'll run into a race condition.

If we are relatively lucky, the job runs long enough, `cwltool` captures the cidfile, removes it, and Podman THROWS OUT A WARNING (cidfile not found).

And if we are unlucky, race condition occurs, `cwltool` HANGS INDEFINITELY in the loop

https://github.com/common-workflow-language/cwltool/blob/3.1.20231207110929/cwltool/job.py#L860

waiting for the cidfile which has already been created AND REMOVED by Podman.

Moving `time.sleep(1)` to the end of the loop might help a bit, but this is still no guarantee to avoid race condition. Currently I'm out of ideas for solving this issue in a correct way.

## Full Traceback

```pytb
Traceback (most recent call last):
  File "/home/user/miniforge3/envs/cwltool/bin/cwltool", line 11, in <module>
    sys.exit(run())
             ^^^^^
  File "cwltool/main.py", line 1457, in run
  File "cwltool/main.py", line 1301, in main
  File "cwltool/executors.py", line 62, in __call__
  File "cwltool/executors.py", line 145, in execute
  File "cwltool/executors.py", line 253, in run_jobs
  File "cwltool/job.py", line 843, in run
  File "cwltool/job.py", line 336, in _execute
  File "cwltool/job.py", line 986, in _job_popen
  File "cwltool/job.py", line 861, in docker_monitor
KeyboardInterrupt
```

## Your Environment

* cwltool version: 3.1.20231207110929

mr-c:
Hello @kuanyili ; thank you for your persistence with podman!

Two things: 

1) I think the return statement at https://github.com/common-workflow-language/cwltool/blob/29af2be2861058fb721afe48a8da271b17a7846f/cwltool/job.py#L872 needs to de-indented one level.

[...]

PR #1962, opened 2023-12-31
Fix fast container job hangs

kuanyili:
Starting from Podman 4.4.0, cidfile is removed along with the container.

Always return if container is terminated so we don't stuck in the loop waiting for a file which has already been removed.

Fixes: #1961


mr-c (approved):

kuanyili:
Hmm.... It's strange that such a small change triggers this many CI errors.

Some udocker jobs are failing but the change is in `docker_monitor` which is a function NOT used by udocker jobs. On the other hand, this patch runs well on my machine and fixed the aforementioned issue beautifully.

The CI error pattern is very similar to that in the most recent PR #1960. Maybe they're related?

mr-c:
Thanks for alerting me to that @kuanyili ; I think I fixed the udocker problem

PR #1965, opened 2024-01-12
Added new test on loop

LanderOtto:
This PR adds a new unit test using scatter and loop features in the workflow. In particular, it is tested if a scatter step works well inside a loop. Before this PR, only the "loop inside scatter" case was checked.


mr-c (approved):
Thank you @LanderOtto !

PR #1972, opened 2024-01-29
add the --shm-size parameter at runtime

### LINKED ISSUE #1821: How to add the -- shm-size parameter at runtime
Shm errors in running deep learning models within a docker container

Error: Unexpected bus error accounted in worker This might be caused by insufficient shared memory (shm).

Reason: The main reason is that the shm size is too small

When creating an analysis container using cwltool, it is expected to add the parameter – shm-size 4g through configuration

docker run  --shm-size 4G  

Can you add this configuration? Thank you

lihl2014:
Fixes #1821  
https://github.com/common-workflow-language/cwltool/issues/1821#issuecomment-1481347082

### DIFF
--- a/cwltool/cwlprov/provenance_profile.py
+++ b/cwltool/cwlprov/provenance_profile.py
@@ -498,7 +498,7 @@ class ProvenanceProfile:
             self.research_object.add_uri(entity.identifier.uri)
             return entity
 
-        if isinstance(value, (str, str)):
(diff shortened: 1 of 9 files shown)


mr-c:
Thank you for this @lihl2014 ; do you know what the equivalent options for podman and singularity/apptainer are?

mr-c (approved):

PR #1974, opened 2024-02-04
Adding new choice to --on-error

AlexTate:
### Summary
This pull request introduces a the new choice `kill` to the `--on-error` parameter.

### Motivation
There currently isn't a way to have cwltool immediately stop parallel jobs when one of them fails. One might expect `--on-error stop` to accomplish this, but the help string is specific and accurate: "do not submit any more steps". Since scatter and subworkflow are treated as single "steps" within the parent workflow, this means cwltool is not wrong to wait for the rest of the scatter jobs to finish when `--on-error stop`. However, sometimes individual scatter jobs take a long time to complete, so if one of them fails early on, cwltool might wait great lengths of time for the other scatter jobs to complete before terminating the workflow. With `--on-error kill`, all running jobs are quickly notified and self-terminate upon one job's failure.

### Forum Post
https://cwl.discourse.group/t/how-to-fail-fast-during-parallel-scatter/868

### DIFF
--- a/cwltool/argparser.py
+++ b/cwltool/argparser.py
@@ -607,10 +607,11 @@ def arg_parser() -> argparse.ArgumentParser:
     parser.add_argument(
         "--on-error",
         help="Desired workflow behavior when a step fails.  One of 'stop' (do "
-        "not submit any more steps) or 'continue' (may submit other steps that "
-        "are not downstream from the error). Default is 'stop'.",
+        "not submit any more steps), 'continue' (may submit other steps that "
+        "are not downstream from the error), or kill (same as stop, but also "
+        "terminates running jobs in the active step(s)). Default is 'stop'.",
         default="stop",
-        choices=("stop", "continue"),
+        choices=("stop", "continue", "kill"),
     )
 
     checkgroup = parser.add_mutually_exclusive_group()
--- a/cwltool/job.py
+++ b/cwltool/job.py
@@ -341,6 +341,7 @@ class JobBase(HasReqsHints, metaclass=ABCMeta):
                 env=env,
                 cwd=self.outdir,
                 make_job_dir=lambda: runtimeContext.create_outdir(),
+                kill_switch=runtimeContext.kill_switch,
                 job_script_contents=job_script_contents,
                 timelimit=self.timelimit,
                 name=self.name,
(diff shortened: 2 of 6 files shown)


cwltool/job.py
```diff
@@ -534,10 +543,27 @@ def get_tree_mem_usage(memory_usage: MutableSequence[Optional[int]]) -> None:
                 if mem_tm is not None:
                     mem_tm.cancel()
 
+        def monitor_kill_switch() -> None:
+            nonlocal ks_tm
+            if kill_switch.is_set():
+                _logger.error("[job %s] terminating by kill switch", self.name)
+                if sproc.stdin: sproc.stdin.close()
```
mr-c, line 550: This needs to be two lines, run `make dev cleanup` (or just `make cleanup` if you already run `make dev`) to fix that automatically

mr-c (changes requested):
Thank you, again,  @AlexTate for your PR!

`tests/test_parallel.py::test_on_error_kill` is unfortunately failing.

cwltool/argparser.py
```diff
@@ -596,10 +596,11 @@ def arg_parser() -> argparse.ArgumentParser:
     parser.add_argument(
         "--on-error",
         help="Desired workflow behavior when a step fails.  One of 'stop' (do "
-        "not submit any more steps) or 'continue' (may submit other steps that "
-        "are not downstream from the error). Default is 'stop'.",
+        "not submit any more steps), 'continue' (may submit other steps that "
+        "are not downstream from the error), or kill (same as stop, but also "
```
mr-c, line 600:
```suggestion
        "are not downstream from the error), or 'kill' (same as 'stop', but also "
```

AlexTate: @mr-c thank you, suggested changes added in 3d16022

mr-c:
Thank you again for this contribution, @AlexTate ! Alas, the test is sometimes hanging in CI: https://github.com/common-workflow-language/cwltool/actions/runs/11496251742/job/31997507449?pr=1974#step:8:1983

cwltool/job.py
```diff
@@ -347,14 +355,20 @@ def stderr_stdout_log_path(
                 processStatus = "permanentFail"
 
             if processStatus != "success":
-                if rcode < 0:
+                if runtimeContext.kill_switch.is_set():
+                    processStatus = "killed"
+                    return
```
tetron, line 360: Instead of `return` here and then putting a huge amount of code in a `finally` block, how about turning this into `raise WorkflowKillSwitch()` or `raise WorkflowException("killed")` and then revert the cleanup code back the way it was?  You're trying to exit early but I think once the kill switch is set, all the other processes are going to shut down in a (more or less) orderly fashion so you should just let the existing code do its thing, unless there's a specific bad behavior you need to avoid.

tetron (changes requested):
I think this is a good idea and the implementation just needs to be tweaked a bit

PR #1977, opened 2024-02-22
Append "cwltool" to HTTP User-Agent string

svonworl:
Hi, my name is Steve, and I'm an engineer on the Dockstore team.  This PR appends the string "cwltool" to the `User-Agent` string of all cwltool HTTP requests issued via the `requests` package, so it looks like:
```
python-requests/2.31.0 cwltool
```

This PR would allow us to differentiate cwltool traffic from other Dockstore traffic that originates from the `requests` package.  After this change, we'd be able to track cwltool usage and handle its requests more appropriately.  For example, it'd be more feasible for us to filter a DOS attack and not affect cwltool traffic, or to tweak our webservice with a cwltool-specific accommodation.


mr-c:
Thanks for your PR @svonworl . There are several other CWL runners that reuse this code. Would you prefer they use their own program name, or should we stick to `cwltool` only, or also append their name?

svonworl:
> Thanks for your PR @svonworl . There are several other CWL runners that reuse this code. Would you prefer they use their own program name, or should we stick to `cwltool` only, or also append their name?

Hi @mr-c, it'd be great if `cwltool` was always in the string, after which the other CWL runners could append their name (ex: `python-requests/2.31 cwltool other-runner`.  That way, it'd be apparent that the queries are cwltool-based and we'd also be able to differentiate the various runners.

svonworl:
Looks like one of the checks failed, could it be a test flake?  I can't see anything in the PR that would've broken it...

PR #1998, opened 2024-05-03
Replace .decode method with str(bytes, "utf-8")

tetron:
This is because some bytes-like objects (such as memoryview) don't have a "decode()" method.  It appears that the preferred way to get a decoded string is to simply call the constructor with the bytes-like object and the desired encoding.


tetron:
I guess what it is telling me is that because I added tests, but the tests are not exhaustive of all lines/branches of the newly tested functions, the % total coverage actually went down.  Argh.

mr-c (approved):

PR #2005, opened 2024-05-23
bump prov to 2.0.0

jfennick:
Bump the version of the `prov` library for compatibility with the following PR:

https://github.com/trungdong/prov/pull/158

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ ruamel.yaml>=0.16.0,<0.19
 rdflib>=4.2.2,<7.1
 shellescape>=3.4.1,<3.9
 schema-salad>=8.4.20230426093816,<9
-prov==1.5.1
+prov==2.0.0
 mypy-extensions
 psutil>=5.6.6
 importlib_resources>=1.4;python_version<'3.9'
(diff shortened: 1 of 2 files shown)


jfennick:
`prov` just released 2.0.1 with the deepcopy fix

mr-c:
Thank you @jfennick , can you fix the test errors?

tests/test_provenance.py
```diff
@@ -490,11 +490,6 @@ def check_prov(
     assert len(engines) == 1, "Found too many WorkflowEngines: %s" % engines
     engine = engines.pop()
 
-    assert (
-        main_run,
-        PROV.wasAssociatedWith,
-        engine,
-    ) in g, "Wf run not associated with wf engine"
```
mr-c, line 497: Why does this need to be removed? Maybe do a side-by-side diff of running `cwltool --provenance some_folder tests/wf/sec-wf-out.cwl` with this PR and without it.

mr-c (changes requested):

tests/test_provenance.py
jfennick: The short answer is that I couldn't get the CI to pass without removing it. Yes, I tried doing side by side diffs and/or other ways of debugging it, but I'm all out of energy. I don't think I'm going to be able to fix this particular assertion.

I suspect the change is due to one of the following two commits from the prov repo, but IDK.
https://github.com/trungdong/prov/pull/117/commits/a8aba637759da7ac6592a33fe6f5d5439d5b460a
https://github.com/trungdong/prov/commit/2e19dcc12950c1886916e7e6be4e1aecac8c9158

PR #2010, opened 2024-06-27
Fix 2d output arrays with format tag

vjaganat90:
The issue is when we print 2d_output arrays which have "format" tag

`touch filename.txt`
`cwltool --cachedir foo output_2D_file_format.cwl`
``` ERROR Workflow error, try again with --debug for more information:
("Error collecting output for parameter 'output_array': output_2D_file_format.cwl:14:3: list indices must
                                                                                        be integers or
                                                                                        slices, not str",
                                                                                        {})
```

The crux of the issue is aslist function, only promotes scalars to 1d arrays. Arbitrary JS can return an arbitrary json, format handling code needs to handle arbitrary json (not just scalars or 1d arrays which is done through aslist function).

With this fix the output should be something like this

```
jaganathv2$ touch filename.txt
jaganathv2$ cwltool --cachedir foo output_2D_file_format.cwl
INFO /Users/jaganathv2/mambaforge/envs/cwt/bin/cwltool 3.1.20220628170239.dev281+g57619ffd
INFO Resolved 'output_2D_file_format.cwl' to 'file:///Users/jaganathv2/Code/WIC/cwltool/tests/output_2D_file_format.cwl'
INFO [job output_2D_file_format.cwl] Output of job will be cached in /Users/jaganathv2/Code/WIC/cwltool/tests/foo/585a1a734523fb1e13030642dc26ad4a
INFO [job output_2D_file_format.cwl] /Users/jaganathv2/Code/WIC/cwltool/tests/foo/585a1a734523fb1e13030642dc26ad4a$ true
INFO [job output_2D_file_format.cwl] completed success
{
    "output_array": [
        [
            {
                "class": "File",
                "location": "file:///Users/jaganathv2/Code/WIC/cwltool/tests/filename.txt",
                "format": "file:///Users/jaganathv2/Code/WIC/cwltool/tests/output_2D_file_format.cwl#some_format",
                "basename": "filename.txt",
                "size": 0,
                "checksum": "sha1$da39a3ee5e6b4b0d3255bfef95601890afd80709",
                "path": "/Users/jaganathv2/Code/WIC/cwltool/tests/filename.txt"
            },
            {
                "class": "File",
                "location": "file:///Users/jaganathv2/Code/WIC/cwltool/tests/filename.txt",
                "format": "file:///Users/jaganathv2/Code/WIC/cwltool/tests/output_2D_file_format.cwl#some_format",
                "basename": "filename.txt",
                "size": 0,
                "checksum": "sha1$da39a3ee5e6b4b0d3255bfef95601890afd80709",
                "path": "/Users/jaganathv2/Code/WIC/cwltool/tests/filename.txt"
            }
        ],
        [
            {
                "class": "File",
                "location": "file:///Users/jaganathv2/Code/WIC/cwltool/tests/filename.txt",
                "format": "file:///Users/jaganathv2/Code/WIC/cwltool/tests/output_2D_file_format.cwl#some_format",
                "basename": "filename.txt",
                "size": 0,
                "checksum": "sha1$da39a3ee5e6b4b0d3255bfef95601890afd80709",
                "path": "/Users/jaganathv2/Code/WIC/cwltool/tests/filename.txt"
            },
            {
                "class": "File",
                "location": "file:///Users/jaganathv2/Code/WIC/cwltool/tests/filename.txt",
                "format": "file:///Users/jaganathv2/Code/WIC/cwltool/tests/output_2D_file_format.cwl#some_format",
                "basename": "filename.txt",
                "size": 0,
                "checksum": "sha1$da39a3ee5e6b4b0d3255bfef95601890afd80709",
                "path": "/Users/jaganathv2/Code/WIC/cwltool/tests/filename.txt"
            }
        ]
    ]
}INFO Final process status is success
```

### DIFF
--- a/cwltool/command_line_tool.py
+++ b/cwltool/command_line_tool.py
@@ -1497,14 +1497,16 @@ class CommandLineTool(Process):
                 else:
 
                     def recursively_insert(j_dict: Any, key: Any, val: Any) -> Any:
-                        """Recursively inserts a value into any dictionaries"""
+                        """Recursively insert a value into any dictionary."""
                         if isinstance(j_dict, List):
                             return [recursively_insert(x, key, val) for x in j_dict]
                         if isinstance(j_dict, Dict):
                             if j_dict.get("class") == "File":
                                 j_dict[key] = val
                             else:
-                                return {x: recursively_insert(y, key, val) for x, y in j_dict.items()}
+                                return {
+                                    x: recursively_insert(y, key, val) for x, y in j_dict.items()
+                                }
                         return j_dict
 
                     result = recursively_insert(result, "format", format_field)
--- a/cwltool/command_line_tool.py
+++ b/cwltool/command_line_tool.py
@@ -1498,9 +1498,9 @@ class CommandLineTool(Process):
 
                     def recursively_insert(j_dict: Any, key: Any, val: Any) -> Any:
                         """Recursively insert a value into any dictionary."""
-                        if isinstance(j_dict, List):
+                        if isinstance(j_dict, MutableSequence):
                             return [recursively_insert(x, key, val) for x in j_dict]
-                        if isinstance(j_dict, Dict):
+                        if isinstance(j_dict, MutableMapping):
                             if j_dict.get("class") == "File":
                                 j_dict[key] = val
                             else:
--- a/cwltool/command_line_tool.py
+++ b/cwltool/command_line_tool.py
@@ -1495,8 +1495,19 @@ class CommandLineTool(Process):
                             )
                         primary["format"] = format_eval
                 else:
-                    for primary in aslist(result):
-                        primary["format"] = format_field
+
+                    def recursively_insert(j_dict: Any, key: Any, val: Any) -> Any:
+                        """Recursively inserts a value into any dictionaries"""
+                        if isinstance(j_dict, List):
+                            return [recursively_insert(x, key, val) for x in j_dict]
+                        if isinstance(j_dict, Dict):
+                            if j_dict.get("class") == "File":
+                                j_dict[key] = val
+                            else:
+                                return {x: recursively_insert(y, key, val) for x, y in j_dict.items()}
+                        return j_dict
+
+                    result = recursively_insert(result, "format", format_field)
             # Ensure files point to local references outside of the run environment
             adjustFileObjs(result, revmap)
 
--- a/tests/test_2D.py
+++ b/tests/test_2D.py
@@ -1,0 +1,17 @@
+import subprocess
+import sys
+
+from .util import get_data
+
+
+def test_output_2D_file_format() -> None:
+    """Test format tag for 2D output arrays."""
+
+    params = [
+        sys.executable,
+        "-m",
+        "cwltool",
+        get_data("tests/output_2D_file_format.cwl"),
+    ]
+
+    assert subprocess.check_call(params) == 0
--- a/tests/test_2D.py
+++ b/tests/test_2D.py
@@ -1,16 +1,20 @@
 import subprocess
 import sys
+from pathlib import Path
 
 from .util import get_data
 
 
 def test_output_2D_file_format() -> None:
-    """Test format tag for 2D output arrays."""
+    """A simple test for format tag fix for 2D output arrays."""
 
+    Path("filename.txt").touch()
     params = [
         sys.executable,
         "-m",
         "cwltool",
+        "--cachedir",  # just so that the relative path of file works out
+        "foo",
         get_data("tests/output_2D_file_format.cwl"),
     ]
 
--- a/tests/test_2D.py
+++ b/tests/test_2D.py
@@ -1,21 +1,19 @@
-import subprocess
-import sys
 from pathlib import Path
+import pytest
+from .util import get_data, get_main_output
 
-from .util import get_data
-
-
-def test_output_2D_file_format() -> None:
+@pytest.fixture(scope="session")
+def test_output_2d_file_format(tmp_path_factory: pytest.TempPathFactory) -> None:
     """A simple test for format tag fix for 2D output arrays."""
 
-    Path("filename.txt").touch()
-    params = [
-        sys.executable,
-        "-m",
-        "cwltool",
-        "--cachedir",  # just so that the relative path of file works out
-        "foo",
-        get_data("tests/output_2D_file_format.cwl"),
-    ]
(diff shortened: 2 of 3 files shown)


cwltool/command_line_tool.py
```diff
@@ -1495,8 +1495,22 @@ def collect_output(
                             )
                         primary["format"] = format_eval
                 else:
-                    for primary in aslist(result):
-                        primary["format"] = format_field
+
+                    def recursively_insert(j_dict: Any, key: Any, val: Any) -> Any:
+                        """This is a helper function that recursively inserts
+                            a value into any dictionary."""
```
mr-c, line 1501:
```suggestion
                    def recursively_insert(j_dict: Any, key: Any, val: Any) -> Any:
                        """Recursively insert a value into any dictionary."""
```

vjaganat90: I exactly had the previous description. The CI complained that it needs to be a proper sentence.


Can you also help me with creating a an empty file as part of the test : 


[https://github.com/common-workflow-language/cwltool/actions/runs/9699693820/job/26770124937#step:8:48](permission denied in CI environment)

tests/test_2D.py
```diff
@@ -0,0 +1,21 @@
...
+import sys
+from pathlib import Path
+
+from .util import get_data
+
+
+def test_output_2D_file_format() -> None:
+    """A simple test for format tag fix for 2D output arrays."""
+
+    Path("filename.txt").touch()
```
mr-c, line 11: That creates a file in the current working directory. If you need to create new files, please use the `tmp_path` Pytest fixture

https://github.com/common-workflow-language/cwltool/blob/main/tests/test_examples.py#L1263-L1268

tests/test_2D.py
```diff
@@ -0,0 +1,21 @@
...
+    params = [
+        sys.executable,
```
mr-c, line 13: It would be better to use `get_main_output()` from https://github.com/common-workflow-language/cwltool/blob/46e0405320014d3d9e74c1047337e81ee2adc89c/tests/util.py#L85C5-L85C16

cwltool/command_line_tool.py
mr-c: Hey, sorry for not replying earlier. When you edit a comment, others don't get an updated notification :-)

I left some comments on your PR with hints on how to fix the permission denied issue for the "No leftovers" test

cwltool/command_line_tool.py
```diff
@@ -1495,8 +1495,21 @@ def collect_output(
...
                         primary["format"] = format_eval
                 else:
-                    for primary in aslist(result):
-                        primary["format"] = format_field
+
+                    def recursively_insert(j_dict: Any, key: Any, val: Any) -> Any:
+                        """Recursively insert a value into any dictionary."""
+                        if isinstance(j_dict, List):
+                            return [recursively_insert(x, key, val) for x in j_dict]
+                        if isinstance(j_dict, Dict):
```
mr-c, line 1503:
```suggestion
                        if isinstance(j_dict, MutableSequence):
                            return [recursively_insert(x, key, val) for x in j_dict]
                        if isinstance(j_dict, MutableMapping):
```

https://docs.python.org/3.8/library/typing.html#typing.Dict

> To annotate arguments it is preferred to use an abstract collection type such as [`Mapping`](https://docs.python.org/3.8/library/typing.html#typing.Mapping).

https://docs.python.org/3.8/library/typing.html#typing.List

> To annotate arguments it is preferred to use an abstract collection type such as [`Sequence`](https://docs.python.org/3.8/library/typing.html#typing.Sequence) or [`Iterable`](https://docs.python.org/3.8/library/typing.html#typing.Iterable).

tests/test_2D.py
```diff
@@ -0,0 +1,21 @@
...
+        "-m",
+        "cwltool",
+        "--cachedir",  # just so that the relative path of file works out
+        "foo",
```
mr-c, line 17:
```suggestion
        "--cachedir",  # just so that the relative path of file works out
        str(tmp_path),
```
combined with using the `tmp_path` Pytest fixture as previously mentioned: https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

mr-c:
Thank you @vjaganat90 ! I've made a note to turn your example into a new conformance test for CWL: https://github.com/common-workflow-language/cwl-v1.3/issues/31

cwltool/command_line_tool.py
vjaganat90: Thanks, I will keep that in mind in future. Fixed.

tests/test_2D.py
vjaganat90: @mr-c  I think, I understood and followed the fixture pattern. I am not sure it is exactly what is needed. I hope the test passes :)

vjaganat90:
@mr-c if the test passes, please have another look.

mr-c:
Okay, the conformance test errors are due to https://github.com/common-workflow-language/cwltest/issues/216

mr-c (approved):
This looks good to me; and we should add a version of this as a CWL v1.3 conformance test. Any thoughts @tetron ?

tetron:
> This looks good to me; and we should add a version of this as a CWL v1.3 conformance test. Any thoughts @tetron ?

Yes, I agree, and I wonder if the spec should have say anything about this behavior as well.

mr-c:
> > This looks good to me; and we should add a version of this as a CWL v1.3 conformance test. Any thoughts @tetron ?
> 
> Yes, I agree, and I wonder if the spec should have say anything about this behavior as well.

Ah, yes, now that I re-read https://www.commonwl.org/v1.2/CommandLineTool.html#CommandInputParameter , it says

> `format`
> optional
> [string](https://www.commonwl.org/v1.2/CommandLineTool.html#CWLType) | array<[string](https://www.commonwl.org/v1.2/CommandLineTool.html#CWLType)> | [Expression](https://www.commonwl.org/v1.2/CommandLineTool.html#Expression)
> Only valid when `type: File` or is an array of `items: File`.

(The quoted text is the same for [CWL v1.1](https://www.commonwl.org/v1.1/CommandLineTool.html#CommandInputParameter) and [v1.0](https://www.commonwl.org/v1.0/CommandLineTool.html#CommandInputParameter))

And yet, the example from this PR is an [array of arrays of `File`s](https://github.com/common-workflow-language/cwltool/pull/2010/files#diff-64d51c6cc41237bdada4f174a90886d411606de9c7c92ff6a1a26b79d0a2ec81R15). So maybe this behavior change has to wait for CWL version 1.3?

PR #2038, opened 2024-09-06
use max_cores in taskQueue instead of system cores

kmavrommatis:
The class `TascQueue` accepts an argument `thread_count` that is used for the max size of a queue.
In the MultithreadedJobExecutor class there is a variable defined (`max_cores`) that is getting its value from the available cores of the machine. Further in the same class when TaskQueue is called instead of using the max_cores it is using `psutil.cpu_count()`. 
I propose to use the `max_cores` variable when initializing the TaskQueue.

Use case. one can override the max_cores in the MutlithreadedJobExecutor after initialization and force cwltool to use only up to a certain number of cores.


kmavrommatis:
Hello, is there any update on this? 
Does this modification work - to your knowledge - with the overall architecture of the tool?

kmavrommatis:
Hi @tetron 
my use case requires to use more cores than the ones that are physically available on the system (because the tasks running in the pipeline use the CPU very sparingly while waiting for network signals). 
I have modified cwltool to allow to specify the number of max cores (as you suggest) in both places in the TaskQueue and MultithreadedJobExecutor to a number that is higher than the actual cpu count (in my typical case it is 256 on a 32 core instance.
Adding the arguments you suggest would work - provided that it is allowed to specify higher number of cores than the ones physically available to the instance :)
Thanks

tetron:
@kmavrommatis just to confirm, are you requesting a fractional core in your CWL file?  e.g. `minCores: 0.125` ?  Then it would just need to no longer limit the number of processes it launches to `cpu_count()`.

mr-c:
Hey @kmavrommatis , thank you for opening this PR.

Why do you want to run so many CWL Processes at the same time? Are they very short processes and you want to launch many to avoid delays from the setup and tear-down? Or are they IO-bound, so by over-subscribing the system it is still faster to be processing other jobs while some are waiting for IO?

Ah, I see that you are working with `cwl-tes`. I think you want to issue any many jobs as possible, because all of them are executed elsewhere.

Okay, this PR makes more sense to me now. It should produce identical behavior for everyone, except those that override `‎MultithreadedJobExecutor.max_cores` to some other value.

kmavrommatis:
> Hey @kmavrommatis , thank you for opening this PR.
> 
> Why do you want to run so many CWL Processes at the same time? Are they very short processes and you want to launch many to avoid delays from the setup and tear-down? Or are they IO-bound, so by over-subscribing the system it is still faster to be processing other jobs while some are waiting for IO?
> 
> Ah, I see that you are working with `cwl-tes`. I think you want to issue any many jobs as possible, because all of them are executed elsewhere.
> 
> Okay, this PR makes more sense to me now. It should produce identical behavior for everyone, except those that override `‎MultithreadedJobExecutor.max_cores` to some other value.

Exactly right, `cwl-tes` is waiting for network events and it can handle many tasks at the same time, so it makes sense to make it assume that it can use many more cores than the host provides.
Thanks

mr-c (approved):

PR #2065, opened 2024-11-10
Use "run" with singularity/apptainer instead of "exec", when possible

sameeul:
Context for the PR:
Currently if the user uses a `docker` image in a workflow and depending on the choice of the container engine (`docker` vs `singularity`), the execution can be different, if the `docker` image has an entrypoint. 

I understand that CWL community encourages docker images w/o entrypoints, however there are a good number of images produced by various scientific communities that comes with entrypoints. When this images are used in HPC environment, in most cases, due to security policies (not having access to `root`), `singularity` is chosen as the container engine. In these cases, `docker` images with entrypoint fails to execute. 

I also know that creating `CommandLineTools` with a `baseCommand` can bypass the entrypoint issue when running with `singularity` but that also makes those `CommandLineTools` container engine dependent, in turn making the CWL workflows themselves container engine dependent. 

With those observations, if there is no other strong reason for choosing `singularity exec` in the initial implementation, I would propose to change it to `singularity run`.

Here is also a link of relevant discussion: https://cwl.discourse.group/t/singularity-exec-vs-docker-run/956/4

### DIFF
--- a/cwltool/singularity.py
+++ b/cwltool/singularity.py
@@ -75,6 +75,14 @@ def is_apptainer_1_or_newer() -> bool:
     return v[0][0] >= 1
 
 
+def is_apptainer_1_1_or_newer() -> bool:
+    """Check if apptainer singularity distribution is version 1.1 or higher."""
+    v = get_version()
+    if v[1] != "apptainer":
+        return False
+    return v[0][0] >= 2 or (v[0][0] >= 1 and v[0][1] >= 1)
+
+
 def is_version_2_6() -> bool:
     """
     Check if this singularity version is exactly version 2.6.
@@ -117,6 +125,12 @@ def is_version_3_9_or_newer() -> bool:
         return True  # this is equivalent to singularity-ce > 3.9.5
     v = get_version()
     return v[0][0] >= 4 or (v[0][0] == 3 and v[0][1] >= 9)
+
+
+def is_version_3_10_or_newer() -> bool:
+    """Detect if Singularity v3.10+ is available."""
+    v = get_version()
+    return v[0][0] >= 4 or (v[0][0] == 3 and v[0][1] >= 10)
 
 
 def _normalize_image_id(string: str) -> str:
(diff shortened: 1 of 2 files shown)


mr-c:
6 conformance tests fail when running with Singularity: https://github.com/common-workflow-language/cwltool/actions/runs/11769747600/job/32791926095?pr=2065#step:7:905

cwltool/singularity.py
```diff
@@ -467,7 +467,7 @@ def create_runtime(
         runtime = [
             "singularity",
             "--quiet",
-            "exec",
+            "run",
```
mr-c, line 470:
```suggestion
            "run",
            "--no-eval",
```

but only for Singularity versions 3.10.0+ and apptainer versions 1.1.0+

cwltool/singularity.py
```diff
@@ -74,6 +74,14 @@ def is_apptainer_1_or_newer() -> bool:
...
     return v[0][0] >= 1
 
+def is_apptainer_1_1_or_newer() -> bool:
+    """
+    Check if apptainer singularity distribution is version 1.1 or higher.
+    """
+    v = get_version()
+    if v[1] != "apptainer":
+        return False
+    return v[0][0] >= 1 and v[0][1] >= 1
```
mr-c, line 84:
```suggestion
    return v[0][0] >= 2 or (v[0][0] >= 1 and v[0][1] >= 1)
```
future proofing

cwltool/singularity.py
```diff
@@ -467,11 +479,17 @@ def create_runtime(
         runtime = [
             "singularity",
             "--quiet",
-            "exec",
             "--contain",
             "--ipc",
             "--cleanenv",
         ]
```
mr-c, line 485: This three flags are valid for `singularity exec` or `singularity run`, but not plain `singularity`, so they have to be added after the `exec` or the `run`

sameeul:
Thanks for the feedback. I ran the CI on my fork and looks like the singularity tests are passing now.

mr-c (approved):

PR #2081, opened 2024-12-04
Fix `cwltool:Loop` extension definition

GlassOfWhiskey:
This commit fixes two small errors in the `cwltool:Loop` definition that were breaking the Schema SALAD codegen procedure.


mr-c:
> This commit fixes two small errors in the `cwltool:Loop` definition that were breaking the Schema SALAD codegen procedure.

What's the 2nd error?

PR #2085, opened 2024-12-09
Append "_latest" to image id if no tag is present

sameeul:
Based on the discussion here: https://github.com/common-workflow-language/cwl-utils/pull/342


sameeul:
@mr-c : Hey, can you please take a look at this?

mr-c (approved):

PR #2092, opened 2025-01-09
Change caching pathmapper to respect the pathmapper factory

stxue1:
This changes the PathMapper construction in caching to support the pathmapper factory by calling `self.make_path_mapper` instead of directly constructing `PathMapper`. This should allow Toil to hook its own pathmapper in when caching is enabled and allow https://github.com/DataBiosphere/toil/pull/5187 to work.


mr-c (approved):

mr-c:
Thank you @stxue1 ! I'll make a new release of cwltool including this fix

PR #2101, opened 2025-02-19
Add environmental variable caching

stxue1:
PR to support environmental variable caching as mentioned here: https://github.com/DataBiosphere/toil/pull/5187#issuecomment-2610653243

The structure of the preserve_environment code is roughly copied from a related function: https://github.com/common-workflow-language/cwltool/blob/1b5633876aabd4cb57ef3f1fe91c853f3ee82e46/cwltool/job.py#L486-L497


tests/cache_environment_tool.cwl
```diff
@@ -0,0 +1,8 @@
+class: CommandLineTool
+cwlVersion: v1.2
+inputs:
+  []
+outputs:
+  []
+
+baseCommand: ["echo"]
```
mr-c, line 8:
```suggestion
baseCommand: [ echo, "$TEST_ENV"]
```
Might as well echo the changing environment, for manual testing

mr-c (changes requested):
Thanks for the extra tests! I see that they motivated some code changes :-)

PR #2166, opened 2025-10-24
feat: handle singularity/apptainer sandbox images

### LINKED ISSUE #2165: singularity/apptainer sandbox images are not handled
I want to use Singularity/Apptainer sandbox images that are already present locally.

While using  `--singularity` and `--disable-pull` options it doesn't find the image because it looks for `.sif` extension which doesn't exists in my case because the image is a sandbox (e.g. unpacked).

Here is an example:
```
cwlVersion: v1.2
class: CommandLineTool
inputs:
  local_script:
    type: File
    inputBinding:
      position: 1
outputs:
  local_output:
    type: File
    outputBinding:
      glob: "output.txt"
baseCommand: ["python"]
hints:
  DockerRequirement:
    dockerPull: /path/to/containers/python/3.12-slim
```

```
cwltool --debug --disable-pull --singularity hello_world_docker.cwl inputs.yaml
[...]
DEBUG Singularity error
  File "cwltool/job.py", line 759, in run
  File "cwltool/singularity.py", line 361, in get_from_requirements
cwl_utils.errors.WorkflowException: Container image _path_to_containers_python_3.12-slim_latest.sif not found
ERROR Workflow error:
Singularity is not available for this tool, try --no-container to disable Singularity, or install a user space Docker replacement like uDocker with --user-space-docker-cmd.: Container image _cvmfs_sw.cta-observatory.org_software_containers_python_3.12-slim_latest.sif not found
[...]
```

natthan-pigoux:
Running `cwltool` with `dockerRequirements` (with pull disabled) using a local Singularity/Apptainer sandbox image is not currently handle. `cwltool` is only searching for files (e.g. sif images) and not path as it should for sandbox images.

Here I provide:
- a function using `singularity inspect` to verify that a path is a singularity image
- searching for the sandbox image in the `get_image` method
- a test which build an alpine sandbox image and use it in `dockerPull`

Some questions I have:
- if the sandbox image is found I decided to skip the `candidates` search, I don't know if we wish that
- I'm not sure the `singularity inspect` is strong enough to ensure the presence of the image

It is the first time I contribute here so I might have miss some things :)

closes #2165

### DIFF
--- a/tests/test_tmpdir.py
+++ b/tests/test_tmpdir.py
@@ -463,6 +463,8 @@ def test_singularity_image_base_path(
     )
     assert res_get_req == str(image_path)
 
+    with _IMAGES_LOCK:
+        _IMAGES.clear()
     requirements3: CWLObjectType = {"class": "DockerRequirement", "dockerPull": "alpine"}
     # get requirements from without sandbox image path
     # should return an error
--- a/cwltool/singularity.py
+++ b/cwltool/singularity.py
@@ -166,17 +166,9 @@ def _normalize_sif_id(string: str) -> str:
     return string.replace("/", "_") + ".sif"
 
 
-def _normalize_sandbox_id(string: str) -> str:
-    if ":" not in string:
-        string += "_latest"
-    return string.replace("/", "_")
-
-
 @mypyc_attr(allow_interpreted_subclasses=True)
 def _inspect_singularity_sandbox_image(path: str) -> bool:
     """Inspect singularity sandbox image to be sure it is not an empty directory."""
-    if not os.path.isdir(path):
-        return False
     cmd = [
         "singularity",
         "inspect",
@@ -314,10 +306,10 @@ class SingularityCommandLineJob(ContainerCommandLineJob):
                 found = True
         elif "dockerImageId" not in docker_req and "dockerPull" in docker_req:
             # looking for local singularity sandbox image and handle it as a local image
-            sandbox_image_path = os.path.join(
-                sandbox_base_path, _normalize_sandbox_id(dockerRequirement["dockerPull"])
-            )
-            if _inspect_singularity_sandbox_image(sandbox_image_path):
+            sandbox_image_path = os.path.join(sandbox_base_path, dockerRequirement["dockerPull"])
+            if os.path.isdir(sandbox_image_path) and _inspect_singularity_sandbox_image(
+                sandbox_image_path
+            ):
                 docker_req["dockerImageId"] = sandbox_image_path
                 _logger.info(
                     "Using local Singularity sandbox image found in %s",
@@ -339,7 +331,9 @@ class SingularityCommandLineJob(ContainerCommandLineJob):
         elif "dockerImageId" in docker_req:
             sandbox_image_path = os.path.join(sandbox_base_path, dockerRequirement["dockerImageId"])
             # handling local singularity sandbox image
-            if _inspect_singularity_sandbox_image(sandbox_image_path):
+            if os.path.isdir(sandbox_image_path) and _inspect_singularity_sandbox_image(
+                sandbox_image_path
+            ):
                 _logger.info(
                     "Using local Singularity sandbox image found in %s",
                     sandbox_image_path,


mr-c:
Thank you @natthan-pigoux ; can you run `make cleanup` and push again?

natthan-pigoux:
@mr-c , thank you to already have a look. I will be away next week, so I will probably not answer before 2nd November.

mr-c:
@natthan-pigoux Code coverage of the tests looks great, thank you! I'll investigate the mypy/mypyc issue

mr-c:
> The image path can be specified within `dockerPull` and should not necessarily stored in the current working directory. But you are right that in tests I refer only to relative path, I will add another one with an absolute path.

No absolute paths and no relative paths that go "up"; we need to check for both and not allow them. They are bad for portability of the `CommandLineTool` descriptions; and possible security issues.

We need to decide on a default location for sandbox/unpacked singularity/apptainer images, and way for the user to specify an alternative root folder for those (command line option to `cwltool` and/or a environment variable).

mr-c:
> > Yes, but I'd like to re-use CWL_SINGULARITY_CACHE for this purpose.
> 
> Fine for me, event if it can be a bit confusing with the actual singularity cache purpose.

How about we rename that to `CWL_SINGULARITY_IMAGES` and still support the old name?

> 
> > How about --singularity-sandbox-path ?
> 
> Also OK for me.
> 
> When you said:
> 
> > No absolute paths and no relative paths that go "up"
> 
> Do you mean that even in `CWL_SINGULARITY_CACHE` we should not use absolute path? or only in `dockerPull`?

only in `dockerPull` should there not be any absolute paths or relative paths that go "up" (subdirectories are okay)

`CWL_SINGULARITY_CACHE` / `CWL_SINGULARITY_IMAGES` must be an absolute path.

natthan-pigoux:
Thank you for the precisions. 

> How about we rename that to CWL_SINGULARITY_IMAGES and still support the old name?

Yes we can support `CWL_SINGULARITY_CACHE`, and ok for `CWL_SINGULARITY_IMAGES`

mr-c:
@natthan-pigoux here are the remaining missing commits: https://github.com/natthan-pigoux/cwltool/compare/feat_handle_singularity_sandbox_images...common-workflow-language%3Acwltool%3Amain
`git fetch --all` will ensure your local checkout sees them

natthan-pigoux:
Sorry, I don't how but something went wrong in my git config ...

tests/test_tmpdir.py
```diff
@@ -285,6 +287,215 @@ def test_dockerfile_singularity_build(monkeypatch: pytest.MonkeyPatch, tmp_path:
...
+        req,
+        pull_image=False,
+        tmp_outdir_prefix=str(tmp_outdir_prefix),
+        force_pull=False,
+    )
+    assert req["dockerImageId"] == str(image_path)
+    assert res
+
+@needs_singularity
+def test_singularity_image_base_path(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> None:
```
mr-c:
```suggestion
    assert res


@needs_singularity
def test_singularity_image_base_path(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> None:
```

kmk142789 (approved):

tests/test_singularity.py
```diff
@@ -252,3 +258,150 @@ def test_singularity_dockerfile_with_name_with_cache(
...
+        result_code, _, _ = get_main_output(
+            [
+                "--singularity",
+                "--disable-pull",
+                get_data("tests/sing_local_sandbox_test.cwl"),
+                "--message",
+                "hello",
+            ]
+        )
+        assert result_code == 0
```
mr-c: Please use a monkeypatch to adjust the environ, needed for parallel execution of tests
https://docs.pytest.org/en/stable/how-to/monkeypatch.html#monkeypatching-environment-variables

tests/test_tmpdir.py
```diff
@@ -285,6 +287,216 @@ def test_dockerfile_singularity_build(monkeypatch: pytest.MonkeyPatch, tmp_path:
...
+    requirements = copy.deepcopy(initial_requirements)
+    os.environ["CWL_SINGULARITY_IMAGES"] = str(repo_path.absolute())
+    res_get_image = SingularityCommandLineJob(
+        builder, {}, CommandLineTool.make_path_mapper, [], [], ""
+    ).get_image(
+        requirements,
+        pull_image=False,
+        tmp_outdir_prefix=str(tmp_outdir_prefix),
+        force_pull=False,
+    )
```
mr-c: Please use a monkeypatch to adjust the environ, needed for parallel execution of tests
https://docs.pytest.org/en/stable/how-to/monkeypatch.html#monkeypatching-environment-variables

natthan-pigoux:
@mr-c , can you reproduce the test failure locally?

mr-c:
> @mr-c , can you reproduce the test failure locally?

Not yet, no

natthan-pigoux:
Thanks a lot @mr-c  ! It looks better now yes.

PR #2171, opened 2025-10-29
Fix normalization of image id for singularity pre-pulled images

spinto:
Apptainer will replace the ':' character with an '_' when the image is pulled during pre-pull, not a '/'.

In fact, '_latest' is correctly added if a ':' is not present. I am not sure what is the reason for the '/' replacement, as '/' is not allowed at all in the docker image name+tag.

This pull fixes the replacement of the ':' character.


mr-c:
What does singularity do in recent versions? If it is something different, then we'll need an apptainer specific fix

mr-c:
Okay, now we need to figure out which versions of singularity & apptainer have which behavior

PR #2177, opened 2025-11-10
Discard undeclared fields from `cwl.output.json`

GlassOfWhiskey:
This commit discards all fields that are present in a `cwl.output.json` file but have not been declared in the `outputs` section of the related `CommandLineTool`, implementing common-workflow-language/cwl-v1.3#80.
Conformance tests for this update have been implemented in common-workflow-language/cwl-v1.3#81.

### DIFF
--- a/tests/test-cwl-out-v1.3.cwl
+++ b/tests/test-cwl-out-v1.3.cwl
@@ -1,0 +1,15 @@
+class: CommandLineTool
+cwlVersion: v1.3.0-dev1
+
+inputs: []
+
+baseCommand: sh
+
+arguments:
+   - -c
+   - |
+     echo '{"foo": 5 }'
+
+stdout: cwl.output.json
+
+outputs: {}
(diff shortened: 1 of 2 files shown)


cwltool/command_line_tool.py
```diff
@@ -1257,7 +1257,17 @@ def collect_output_ports(
 
                 if compute_checksum:
                     adjustFileObjs(ret, partial(compute_checksums, fs_access))
-            expected_schema = cast(Schema, self.names.get_name("outputs_record_schema", None))
+            expected_schema = cast(RecordSchema, self.names.get_name("outputs_record_schema", None))
+            for k in list(ret.keys()):
```
mr-c:
```suggestion
            for k in ret:
```

mr-c (changes requested):
- [ ] Add something like `if ORDERED_VERSIONS.index(self.metadata[ORIGINAL_CWLVERSION]) >= ORDERED_VERSIONS.index("v1.3.0-dev1"):`  to only change behavior for CWL v1.3.0+
- [ ] add tests here for the previous behavior using `cwlVersion: v1.0`, `cwlVersion: v1.1`, and `cwlVersion: v1.2`

PR #2197, opened 2025-12-19
Suppress erroneous warnings for `pickValue`

LanderOtto:
Handle `pickValue` in `check_types` to prevent erroneous warnings messages


mr-c (approved):
This is great, thank you!

PR #2216, opened 2026-03-06
Detect MPI with Singularity

kinow:
Related to #2208 (not sure if this is enough to close that issue)

@mr-c, I started some tests yesterday, and completed them today after work. Once I managed to run the workflow on my machine reproducing the error we have on HPCs, debugging it showed that the changes appeared simple to be made.

I managed to keep the `--contain`, but in order to do that I had to include `/dev/shm`. I couldn't find any files on `/tmp` or on the temporary directory created by cwltool. And actually ,`cwltool` worked without having to remove the temporary directory created by cwltool (i.e. I was wrong, and the process manager wasn't using the temp dir to sync with processes, but apparently shared memory).

I kept the `--cleanenv` for when the job isn't using the cwltool `MPIRequirement`. When that requirement is a requirement (not a hint) and is used in the job, then I do not set the `--cleanenv`.

I tried keeping it, and even tested the MPI config file using the variables from the [MPICH documentation](https://github.com/pmodels/mpich/blob/6c90f711c3a7c117aa3dc298b66ddd1f4f5fa0f3/doc/wiki/design/Environment_Variables.md).

```yaml
runner: mpirun.mpich
extra_flags: [
# "likwid-perfctr",
# "-C", "L:N:0",
# "-g", "FLOPS_DP",
# "-o", "/output/path/likwid_%j_%h_%r.json"
]
nproc_flag: -n
env_pass_regex: [
  "MPIEXEC_.*",
  "HYDRA_.*",
  "SMPD_.*",
  "MPICH_.*",
  "DCMF_.*",
  "MPIO_.*",
  "RMS_.*",
  "MPITEST_.*",
  "PMI_*",
  "MPI_*"
  #  "SLURM_.*"
]
```

(The MPI config file was needed on my laptop as I have OpenMPI & MPICH, so I need to override the runner as mpirun defaults to OpenMPI on my machine, and the container I'm using has MPICH -- I also confirmed with the debugger that the MPI config file values were added)

The issue with the environment variables, @mr-c, is that when `cwltool` runs, I don't have any environment variables yet.

When we create the CWL job, the runtime config, and even when we pass through the env vars from [`mpi.py`](https://github.com/common-workflow-language/cwltool/blob/b363bd8f93971e9b1eccbbdb6e408ee84b5140c7/cwltool/mpi.py#L66-L80), there are no MPI variables loaded yet.

The Edinburgh paper discusses how to use some variables, but I believe those are only the variable already available. For instance, when you have a Slurm allocation and you have `SLURM_JOB_ID`, `SLURM_MEM_PER_CPU`, etc.

When the CWL job is executed, already with the runtime (command + args) created, and then Python launches `mpirun.mpich`, that's when the MPICH (or Hydra) program will create the environment variables.

```bash
(venv) kinow@ranma:~/Development/python/workspace/cwltool$ env | grep MPI
(venv) kinow@ranma:~/Development/python/workspace/cwltool$ mpirun.mpich -n 1 env | grep MPI
MPIR_CVAR_CH3_INTERFACE_HOSTNAME=ranma
MPI_LOCALNRANKS=1
MPI_LOCALRANKID=0
(venv) kinow@ranma:~/Development/python/workspace/cwltool$ mpirun.mpich -n 1 env | grep PMI
PMI_RANK=0
PMI_FD=9
PMI_SIZE=1
```

I tried setting these variables in the command line of singularity without success. This is what I did:

```
$ mpirun.mpich -n 1 env | sort > /tmp/mpirun.txt
$ env | sort > /tmp/run.txt
$ meld /tmp/run.txt /tmp/mpirun.txt
```

Got the list of variables that appeared in the diff. There were 8 extra variables:

- `GFORTRAN_UNBUFFERED_PRECONNECTED=y`
- `HYDI_CONTROL_FD=7`
- `MPIR_CVAR_CH3_INTERFACE_HOSTNAME=ranma`
- `MPI_LOCALNRANKS=1`
- `MPI_LOCALRANKID=0`
- `PMI_FD=9`
- `PMI_RANK=0`
- `PMI_SIZE=1`

I added them to the Singularity command with just their names to see if it would pass the host env vars to the container (as it wouldn't make sense trying to define random file descriptor, ranks, etc.).

```diff
diff --git a/cwltool/singularity.py b/cwltool/singularity.py
index 8e391d64..128df6e2 100644
--- a/cwltool/singularity.py
+++ b/cwltool/singularity.py
@@ -601,6 +601,12 @@ class SingularityCommandLineJob(ContainerCommandLineJob):
         if not mpi_req or not is_req:
             runtime.append("--cleanenv")
         else:
+            runtime.append("--cleanenv")
+            for var in ["PMI_RANK", "PMI_FD", "PMI_SIZE",
+                        "MPI_LOCALNRANKS", "MPI_LOCALRANKID", "MPIR_CVAR_CH3_INTERFACE_HOSTNAME",
+                        "HYDI_CONTROL_FD", "GFORTRAN_UNBUFFERED_PRECONNECTED"]:
+                runtime.append("--env")
+                runtime.append(var)
             self.append_volume(
                 runtime,
                 runtime_context.create_tmpdir(),
```

The resulting command has the `--cleanenv` and all the environment variables from the diff:

```bash
INFO [job run] /tmp/hib1sp8z$ mpirun.mpich \
    -n \
    2 \
    singularity \
    --quiet \
    run \
    --ipc \
    --contain \
    --cleanenv \
    --env \
    PMI_RANK \
    --env \
    PMI_FD \
    --env \
    PMI_SIZE \
    --env \
    MPI_LOCALNRANKS \
    --env \
    MPI_LOCALRANKID \
    --env \
    MPIR_CVAR_CH3_INTERFACE_HOSTNAME \
    --env \
    HYDI_CONTROL_FD \
    --env \
    GFORTRAN_UNBUFFERED_PRECONNECTED \
    --mount=type=bind,source=/tmp/g7lo1zku,target=/dev/shm \
    --no-eval \
    --userns \
    --home \
    /tmp/hib1sp8z:/PCIKJh \
    --mount=type=bind,source=/tmp/ub89clkk,target=/tmp \
    --mount=type=bind,source=/tmp/h8j2gy1j/a.out,target=/var/lib/cwl/stgfb9c3552-0c10-424c-82ba-284742e69287/a.out,readonly \
    --pwd \
    /PCIKJh \
    --net \
    --network \
    none \
    /home/kinow/Development/python/workspace/cwl-mpi/examples/mpich-sr/cwltool/mfisherman_mpich:4.3.2.sif \
    /var/lib/cwl/stgfb9c3552-0c10-424c-82ba-284742e69287/a.out \
    0 \
    1 > /tmp/hib1sp8z/sr.out 2> /tmp/hib1sp8z/sr.err
WARNING [job run] exited with status: 1
WARNING [job run] completed permanentFail
WARNING [step run] completed permanentFail
```

That failed again. And running a smaller test looks like only key=value is valid.

```bash
$ ARUBA=111 mpirun.mpich -n 2 singularity run --cleanenv --env ARUBA mfisherman_mpich\:4.3.2.sif env | grep ARUBA
Error for command "run": invalid argument "ARUBA" for "--env" flag: ARUBA must be formatted as key=value

Error for command "run": invalid argument "ARUBA" for "--env" flag: ARUBA must be formatted as key=value
```

But since I don't know the file descriptor, rank, etc., I can't see a way to define PMI_RANK, PMI_SIZE, PMI_FD, and other variables for MPICH.

@mr-c I did not write tests as I wanted to check with you first if this is going in the right direction. I'll run a couple of tests on MN5 and CESGA FT3. My main concern is `--contain` and InfiniBand. Even without `--cleanenv`, I think `--contain` may force the container to use ethernet instead of InfiniBand.

### DIFF
--- a/cwltool/mpi.py
+++ b/cwltool/mpi.py
@@ -23,6 +23,8 @@ class MpiConfig:
         env_pass: list[str] | None = None,
         env_pass_regex: list[str] | None = None,
         env_set: Mapping[str, str] | None = None,
+        shm_enabled: bool = True,
+        shm_dir: str = "/dev/shm",  # nosec B108 - required for MPI/shared memory in containers
     ) -> None:
         """
         Initialize from the argument mapping.
@@ -35,6 +37,8 @@ class MpiConfig:
         env_pass: []
         env_pass_regex: []
         env_set: {}
+        shm_enabled: True
+        shm_dir: "/dev/shm
 
         Any unknown keys will result in an exception.
         """
@@ -45,6 +49,11 @@ class MpiConfig:
         self.env_pass = env_pass or []
         self.env_pass_regex = env_pass_regex or []
         self.env_set = env_set or {}
+        self.shm_enabled = shm_enabled
+        # POSIX only contains functions to handle shared memory, but it does not
+        # specify the directory to be used, nor if a directory needs to be used
+        # at all -- ref: https://pubs.opengroup.org/onlinepubs/9699919799/
+        self.shm_dir = shm_dir
 
     @classmethod
     def load(cls: type[MpiConfigT], config_file_name: str) -> MpiConfigT:
--- a/cwltool/singularity.py
+++ b/cwltool/singularity.py
@@ -1,5 +1,6 @@
 """Support for executing Docker format containers using Singularity {2,3}.x or Apptainer 1.x."""
 
+import atexit
 import copy
 import hashlib
 import json
@@ -11,7 +12,10 @@ import shutil
 import sys
 import threading
 from collections.abc import Callable, MutableMapping, MutableSequence
+from contextlib import suppress
+from importlib.resources import files as resource_files
 from subprocess import check_call, check_output, run  # nosec
+from tempfile import NamedTemporaryFile
 from typing import cast
 
 from cwl_utils.types import CWLDirectoryType, CWLFileType, CWLObjectType
@@ -29,6 +33,7 @@ from .docker import DockerCommandLineJob
 from .errors import WorkflowException
 from .job import ContainerCommandLineJob
 from .loghandler import _logger
+from .mpi import MPIRequirementName
 from .pathmapper import MapperEnt, PathMapper
 from .singularity_utils import singularity_supports_userns
 from .utils import create_tmp_dir, ensure_non_writable, ensure_writable
@@ -203,7 +208,7 @@ class SingularityCommandLineJob(ContainerCommandLineJob):
         hints: list[CWLObjectType],
         name: str,
     ) -> None:
-        """Builder for invoking the Singularty software container engine."""
+        """Builder for invoking the Singularity software container engine."""
         super().__init__(builder, joborder, make_path_mapper, requirements, hints, name)
 
     @staticmethod
@@ -592,14 +597,55 @@ class SingularityCommandLineJob(ContainerCommandLineJob):
         """Return the Singularity runtime list of commands and options."""
         any_path_okay = self.builder.get_requirement("DockerRequirement")[1] or False
 
-        runtime = [
-            "singularity",
-            "--quiet",
-            "run" if (is_apptainer_1_1_or_newer() or is_version_3_10_or_newer()) else "exec",
-            "--contain",
-            "--ipc",
-            "--cleanenv",
-        ]
+        mpi_req, is_req = self.builder.get_requirement(MPIRequirementName)
+        mpi_enabled = mpi_req and is_req
+        mpi_config = runtime_context.mpi_config
+        mpi_env_vars_reference_file_name: str | None = None
+        runtime: list[str] = []
+        if mpi_enabled:
+            # Save current environment variables. The ``singularity_wrapper.sh`` will
+            # diff it against the env vars produced by mpirun/srun/etc., and use the new
+            # env vars as SINGULARITYENV_... for Singularity.
+            with NamedTemporaryFile(mode="w+", delete=False) as f:
+                for k, v in os.environ.items():
+                    f.write(f"{k}={v}\n")
+                mpi_env_vars_reference_file_name = f.name
+
+            def delete_mpi_baseline_env() -> None:
+                """Clean up the MPI baseline environment variables file at exit."""
+                with suppress(FileNotFoundError):  # pragma: no cover
+                    os.remove(mpi_env_vars_reference_file_name)  # pragma: no cover
+
+            atexit.register(delete_mpi_baseline_env)
+
+            runtime.extend(
+                [
+                    str(resource_files("cwltool") / "singularity_wrapper.sh"),
+                    mpi_env_vars_reference_file_name,
+                    "singularity",
+                ]
+            )
+        else:
+            runtime.append("singularity")
+
+        runtime.extend(
+            [
+                "--quiet",
+                "run" if (is_apptainer_1_1_or_newer() or is_version_3_10_or_newer()) else "exec",
+                "--contain",
+                "--ipc",
+                "--cleanenv",
+            ]
+        )
+        if mpi_enabled and mpi_config.shm_enabled:
+            # MPI implementations like OpenMPI and MPICH use shared memory.
+            self.append_volume(
+                runtime,
+                runtime_context.create_tmpdir(),
+                mpi_config.shm_dir,
+                writable=True,
+            )
+
         if is_apptainer_1_1_or_newer() or is_version_3_10_or_newer():
             runtime.append("--no-eval")
 
@@ -665,4 +711,4 @@ class SingularityCommandLineJob(ContainerCommandLineJob):
         if container_HOME:
             # Restore HOME if we removed it above.
             self.environment["HOME"] = container_HOME
-        return (runtime, None)
+        return runtime, None
--- a/tests/test_singularity_wrapper.py
+++ b/tests/test_singularity_wrapper.py
@@ -1,0 +1,85 @@
+"""Tests for the Shell wrapper of the Singularity command.
+
+This script tests a Shell script. This script does not contribute to the
+project test coverage (although kcov, or bats+kcov could be used in the
+future).
+"""
+
+import os
+import subprocess
+from importlib.resources import files
+from textwrap import dedent
+from typing import TYPE_CHECKING
+
+import pytest
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+
+@pytest.mark.parametrize(
+    "args,expected_return_code",
+    [(["--help"], 1), ([""], 1), (["singularity"], 1)],
+    ids=[
+        "Print usage because user passed --help",
+        "Print usage because of missing args",
+        "Print usage because not all args provided",
+    ],
+)
+def test_wrapper_usage(args: list[str], expected_return_code: int) -> None:
+    """Test the usage of the Singularity wrapper is printed."""
+    wrapper = str(files("cwltool") / "singularity_wrapper.sh")
+    command: list[str] = [wrapper] + args
+    result = subprocess.run(command, capture_output=True, text=True)
+
+    assert result.returncode == expected_return_code
+    assert "Wrapper around Singularity/Apptainer for CWL + MPI + Singularity" in result.stderr
+
+
+def test_wrapper_invalid_baseline_env_file() -> None:
+    """Test the script fails if the given file is not valid."""
+    wrapper = str(files("cwltool") / "singularity_wrapper.sh")
+    command: list[str] = [wrapper, "parangaricutirimicuaro.dat", "foo"]
+    result = subprocess.run(command, capture_output=True, text=True)
+
+    assert result.returncode == 2
+    assert "file not found" in result.stderr
+
+
+def test_wrapper_env_vars(tmp_path: "Path") -> None:
+    """Test that the wrapper script adds the new environment variables."""
+    fake_singularity = tmp_path / "fake_singularity"
+    fake_singularity.write_text(dedent("""\
+    #!/bin/bash
+    echo "Fake Singularity script"
+    env
+    """))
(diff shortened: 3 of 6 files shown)


cwltool/singularity.py
```diff
@@ -593,10 +594,19 @@ def create_runtime(
...
         ]
+        mpi_req, is_req = self.builder.get_requirement(MPIRequirementName)
+        if not mpi_req or not is_req:
+            runtime.append("--cleanenv")
+        else:
+            self.append_volume(
+                runtime,
+                runtime_context.create_tmpdir(),
+                "/dev/shm",
+                writable=True,
```
mr-c: Neat trick; is this universal for all known MPI implementations? Or do some not use the shared memory device for communication between MPI processes?

kinow: That's not defined in the MPI specification. OpenMPI [uses shared memory](https://docs.open-mpi.org/en/main/tuning-apps/networking/shared-memory.html). MPICH also [uses shared memory](https://github.com/pmodels/mpich/blob/badff5b825ce1ea6e548dcadb99350d4956bffb9/src/util/mpir_nodemap.c#L19-L31). Cray and Intel are [derivatives from MPICH](https://en.wikipedia.org/wiki/MPICH#MPICH_derivatives) so they likely use it by default too. It appears there are some environment variables to disable it, but I haven't seen any case where shared memory was disabled for MPI. If you think it's useful I can add a new option to disable it in cwltool, or we can do it later if a user requests it?

cwltool/singularity.py
```diff
@@ -592,14 +596,50 @@ def create_runtime(
         """Return the Singularity runtime list of commands and options."""
         any_path_okay = self.builder.get_requirement("DockerRequirement")[1] or False
 
-        runtime = [
-            "singularity",
+        mpi_req, is_req = self.builder.get_requirement(MPIRequirementName)
+        mpi_enabled = mpi_req and is_req
```
kinow, line 601: I think the `is_req` excludes if MPIRequirement is used as a hint. I'm considering that MPI is used/enabled only if it's a requirement, and not a hint.

cwltool/singularity.py
```diff
@@ -593,10 +594,19 @@ def create_runtime(
...
             "--quiet",
             "run" if (is_apptainer_1_1_or_newer() or is_version_3_10_or_newer()) else "exec",
-            "--contain",
             "--ipc",
-            "--cleanenv",
+            "--contain",
         ]
+        mpi_req, is_req = self.builder.get_requirement(MPIRequirementName)
+        if not mpi_req or not is_req:
+            runtime.append("--cleanenv")
```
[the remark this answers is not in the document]
kinow: I've added --cleanenv back, and added a wrapper script. I will need some time now to test it, but in the meantime feel free to have a look at the code here and at the new Shell script to see if we are going in the right direction :) 👍

cwltool/mpi.py
```diff
@@ -45,6 +49,11 @@ def __init__(
         self.env_pass = env_pass or []
         self.env_pass_regex = env_pass_regex or []
         self.env_set = env_set or {}
+        self.shm_enabled = shm_enabled
+        # POSIX only contains functions to handle shared memory, but it does not
+        # specify the directory to be used, nor if a directory needs to be used
+        # at all -- ref: https://pubs.opengroup.org/onlinepubs/9699919799/
```
kinow, line 55: Looks like FreeBSD uses objects instead of files as in Linux: 

https://man.freebsd.org/cgi/man.cgi?shm_open#:~:text=The%20shm%5Fopen%28%29%20and%20shm%5Funlink%28%29%20functions%20first%20appeared%20in%20FreeBSD%204%2E3%2E%20The%20functions%20were%20reimplemented%20as%20system%20calls%20using%20shared%20memory%20objects%20directly%20rather%20than%20files%20in%20FreeBSD%208%2E0%2E

These two, shm_enabled and shm_dir, are used to control whether the shared memory will be used and the directory to be used, respect/.

And MPICH allows users to disable the shared memory completely, or change its size. https://github.com/pmodels/mpich/blob/main/doc/wiki/faq/Frequently_Asked_Questions.md#:~:text=The%20work,stack

So with these settings I think users are able to match the behavior of that implementation at least.

tests/test_singularity_wrapper.py
```diff
@@ -0,0 +1,108 @@
+"""Tests for the Shell wrapper of the Singularity command.
+
+This script tests a Shell script. This script does not contribute to the
+project test coverage (although kcov, or bats+kcov could be used in the
+future).
```
kinow, line 5: I use `kcov` with `bats` to test and track coverage of Shell code in a project. It can be combined with Python and pytest reports too, but it takes some time to set everything up. So here I used a pytest that, from my own debugging, appears to cover all the branches and lines in the Bash code.

kinow:
Tested the current version, and it's passing OK on MN5 :tada: 

```
}INFO Final process status is success
```

I moved the shared memory settings to `MpiConfig` from `MPIRequirement`. There are two new settings there, `shm_dir=/dev/shm` and `shm_enabled=True`.

Also added an initial test for the wrapper Bash script.

@mr-c, I think what's missing now are the unit tests to improve coverage of the code changed, and to pass the macos tests (I may have used some linux command that's not compatible, or some arg).

It shouldn't be too hard, I think I can get the build passing and all tests green this week.

kinow:
@mr-c. I just tested this again on MareNostrum5, and it worked without issues :relieved: 

The version of the Shell I had did not work on MacOS. First, it failed because a function `is_apptainer_1_1_or_newer()` fails when executed on MacOS (I think this was not called before in the tests). You can probably still see the log here: https://github.com/common-workflow-language/cwltool/actions/runs/25075581707/job/73467209090?pr=2216

To fix that I added a fake `singularity` binary in the `$PATH` for one of the tests, returning the fake script. That script is not used in Linux tests, as subprocess appears to find the real singularity first (it's probably searching the binary directly through another way, and if not found searching the path), but after doing that change the tests on macos-latest failed with a different error.

The macos tests now failed because the fake singularity was printing a random string, and the `is_apptainer_....` functions called `singularity --version`, so I had to return a valid Singularity version. The failure log for `get_version` should still be available here, https://github.com/common-workflow-language/cwltool/actions/runs/25075581707/job/73467209090?pr=2216.

There's an atexit function to clean up the temporary file with the baseline env vars (used to diff with the mpirun envs). That function is not easy to test, and it has one statement wrapped in a `contextlib.suppress()`, so I marked that as `# pragma: nocover` to ignore that in coverage reports. Let me know if you prefer that to be marked as not covered instead (right now it's not red nor green in the codecov report; it appears as white like comments/etc.). The coverage CI check is still failing as there's a -0.17% diff (lines added/removed), but I think the test coverage should be OK.

Once, there was a failure in a conformance 1.3 test. Re-running with another commit (that was fixing linting errors), the test now passed.

There were also failures in the build_test_container job that I am not sure how to fix.

Oh, and about the singularity wrapper. Initially the code was using `--env MISSING_VAR=SOME_VALUE`, but I realized that the command line was *really, really, really* long on MareNostrum5, and we could have issues with commands being truncated or failing to run in some cases, I think (or terrible entries in logs). So I switched to `SINGULARITYENV_$MISSING_VAR=SOMEVALUE`. :+1: 

This should now be ready for review :relieved: 

Cheers

kinow:
Thanks for rebasing it, Michael. I noticed the test failures, but I couldn't really link them directly to the changes I did, so I just re-launched the failed GH Actions jobs (a few times).

kinow:
Note to self: check the README later/next week. There is a table of parameters supported, and if we add one for singularity & shared memory, we need to update that table. Noticed that now while reading the docs & writing thesis.

kinow:
@mr-c, I edited the commit to add this to `README.rst`

```
+----------------+------------------+------------+------------------------------+
| shm_enabled    | bool             | True       | Flag to control whether      |
|                |                  |            | shared memory is used or not.|
+----------------+------------------+------------+------------------------------+
| shm_dir        | str              | "/dev/shm" | Location to use for shared   |
|                |                  |            | memory.                      |
+----------------+------------------+------------+------------------------------+
```

kinow:
I am testing this pull request with the getit-workflows FALL3D modified workflow. Original: https://gitlab.geo3bcn.csic.es/fall3d/getit-workflows. Fork I am testing: https://github.com/kinow/getit-workflows/pull/1/

Marked it as a draft while I test. If my tests succeed, then I will mark it again as ready for review :+1:

kinow:
Interesting. It failed on LUMI because the only MPI available is CrayMPI. And FALL3D is compiled in the container with OpenMPI! A case I didn't think about. Will have to test it again later on MN5 one more time. But I think if this is merged as-is, it should work but we need to improve the documentation a bit more.

kinow:
Note: this PR worked correctly on LUMI today. What I was missing was to add `--mpi=pmi2` in extra_flags.

kinow:
I'm doing the very last test for the thesis:

- Using Hetzner cloud 2 nodes with OpenMPI
- Launch the FALL3D application compiled with OpenMPI
- Use this branch, and provide a MPI Config file with the env vars needed (found via `env` and the `mpirun -np1 env`, to pass to the first level / MPIRequirement)
- The Singularity Wrapper of this PR doesn't work

I have a few possibilities,

- The wrapper and the exported vars are in conflict (although this didn't happen in other cases, and I tried without anything in the `env_pass_regex`).
- OpenMPI + PRRTE is a bit special, and has something that's not supported when using an intermediary shell (my guess)
- Something else

I tested a simple `singularity exec openmpi.sif mpicc sr.c -o sr`, then `mpirun -np 2 singularity exec ./sr`, and that worked. This indicates that Singularity & OpenMPI are able to work in tandem in the env (VM). But that with the wrapper script something appears to fail to submit the MPI application.

So far all my development here, and most other tests were performed on my laptop (MPICH) and on LUMI (Cray MPICH).

PR #2286, opened 2026-06-12
Fix: Honor EnvVarRequirement on workflow steps with --preserve-entire…

mmancini-skao:
When running a workflow with --preserve-entire-environment, EnvVarRequirements set on individual workflow steps were being ignored. This was because the prepare_environment() method in job.py was applying the required environment variables (HOME, TMPDIR, PATH) after applying the EnvVarRequirement, causing the requirement to be overwritten.

Root Cause:
The prepare_environment() method was calling _required_env() after extract_environment(). Since dict.update() overwrites existing keys, the values from _required_env() (which includes TMPDIR and HOME) were overwriting any values set by EnvVarRequirement.

Solution:
Reorder the operations so that _required_env() is called first, then extract_environment() is called, allowing EnvVarRequirement to take precedence.

This ensures that:
1. Default required environment variables (HOME, TMPDIR, PATH) are set
2. Preserved environment variables are applied
3. EnvVarRequirement values override defaults (highest priority)

Test:
Added test_preserve_entire_environment_honors_step_env_requirements() to verify that EnvVarRequirement values are respected when --preserve-entire-environment is used. The test uses TMPDIR as an example, setting it to /custom/tmpdir in the workflow step's EnvVarRequirement and verifying it is not overwritten by the --tmpdir-prefix value.

Files Changed:
- cwltool/job.py: Reordered prepare_environment() operations
- tests/test_environment.py: Added test case
- tests/env-preserve-step-wf.cwl: Test workflow with EnvVarRequirement
- tests/env-preserve-step-tool.cwl: Test tool for environment inspection

### DIFF
--- a/cwltool/job.py
+++ b/cwltool/job.py
@@ -511,10 +511,10 @@ class JobBase(HasReqsHints, metaclass=ABCMeta):
         elif runtimeContext.preserve_environment:
             self._preserve_environment_on_containers_warning(runtimeContext.preserve_environment)
 
-        env.update(self.extract_environment(runtimeContext, envVarReq))
-
         # Set required env vars
         env.update(self._required_env())
+
+        env.update(self.extract_environment(runtimeContext, envVarReq))
 
         # Set on ourselves
         self.environment = env
--- a/cwltool/job.py
+++ b/cwltool/job.py
@@ -512,9 +512,14 @@ class JobBase(HasReqsHints, metaclass=ABCMeta):
             self._preserve_environment_on_containers_warning(runtimeContext.preserve_environment)
 
         # Set required env vars
+        # Preserve host variables first.
+        env.update(self.extract_environment(runtimeContext, {}))
+
+        # Keep required CWL runtime vars deterministic.
         env.update(self._required_env())
 
-        env.update(self.extract_environment(runtimeContext, envVarReq))
+        # Allow explicit EnvVarRequirement to override defaults.
+        env.update(envVarReq)
 
         # Set on ourselves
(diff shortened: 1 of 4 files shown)


mmancini-skao:
@mr-c Hello Michael, we ran into this issue. 
I am not sure if it is supposed to be like this or it is an actual issue. 
Let me know.

mr-c:
Thank you for this PR. Yes, this is probably a correct fix, but I will need to review it in more detail later. Let's see if the conformance tests still pass.

mmancini-skao:
Hi @mr-c, sorry I am not familiar with the CI setup here, but I think some approval has to be made to make the CI run.

mr-c:
Reviewing this PR and https://github.com/LOFAR-VLBI/pilot/pull/127 ; it feels like you are trying to work-around an issue your in your CWL runner, toil. The "correct" thing to do is to get Toil to use shorter paths, either by configuration or by a fix there.

I'm uncomfortable with encouraging overrides of TMPDIR, HOME, etc.. to work-around runner misconfiguration/bugs..

mmancini-skao:
> LOFAR-VLBI/pilot#127

I understand your concern and it is valid. The problem however is not actually toil because any path should be ok the problem is python multiprocess that is creating a socket in TMPDIR and that is not allowed. 

I do not think the problem is specifically toil or cwltool but more like a simple step software which is exactly I gather the reason why EnvRequirement could override the default for a step. Also it used to be working before. 
But maybe I am overlooking something, is the standard specifically making an exception for TMPDIR?

PR #2290, opened 2026-06-18
Report program name as cwltool when invoked as cwl-runner (#1535)

### LINKED ISSUE #1535: report prog as "cwltool" if invoked as "cwl-runner"

arimu1:
## Summary

Fixes #1535.

`cwltool` is installed with a `cwl-runner` console-script entry point. Because `argparse` defaults `prog` to `os.path.basename(sys.argv[0])`, invoking it through that alias made the help and error output identify the program as `cwl-runner`:

```console
$ cwl-runner --help
usage: cwl-runner [-h] [--basedir BASEDIR] ...
```

## Change

Set `prog="cwltool"` explicitly on the top-level `ArgumentParser` in `cwltool/argparser.py`, so the reference implementation is always identified consistently no matter which entry point launched it:

```console
$ cwl-runner --help
usage: cwl-runner --help     # before
usage: cwltool [-h] ...      # after
```

This also applies to the usage line printed on argument errors.

## Tests

Added `test_argparser_prog_is_cwltool` to `tests/test_toolargparse.py`, which patches `sys.argv[0]` to `cwl-runner` and asserts `arg_parser().prog == "cwltool"`. Full `test_toolargparse.py` passes (18); `black`/`flake8`/`isort` clean.

---

This change was produced with the assistance of **Claude Code** (model: Claude Opus). It was reviewed by a human before submission.

### DIFF
--- a/cwltool/argparser.py
+++ b/cwltool/argparser.py
@@ -89,6 +89,7 @@ def arg_parser() -> argparse.ArgumentParser:
     )
 
     parser = argparse.ArgumentParser(
+        prog="cwltool",
         formatter_class=RichHelpFormatter,
         description="Reference executor for Common Workflow Language standards. "
         "Not for production use.",
--- a/cwltool/argparser.py
+++ b/cwltool/argparser.py
@@ -89,7 +89,12 @@ def arg_parser() -> argparse.ArgumentParser:
     )
 
     parser = argparse.ArgumentParser(
-        prog="cwltool",
+        # Force the program name to "cwltool" only when invoked through the
+        # generic ``cwl-runner`` alias (see #1535). Leaving it as ``None`` for
+        # any other invocation preserves argparse's basename default, so
+        # downstream tools that reuse this parser (e.g. Calrissian) keep their
+        # own program name.
+        prog="cwltool" if os.path.basename(sys.argv[0]) == "cwl-runner" else None,
         formatter_class=RichHelpFormatter,
         description="Reference executor for Common Workflow Language standards. "
         "Not for production use.",
--- a/tests/test_toolargparse.py
+++ b/tests/test_toolargparse.py
@@ -1,4 +1,5 @@
 import argparse
+import sys
 from collections.abc import Callable
 from io import StringIO
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


cwltool/argparser.py
```diff
@@ -89,6 +89,7 @@ def arg_parser() -> argparse.ArgumentParser:
     )
 
     parser = argparse.ArgumentParser(
+        prog="cwltool",
```
mr-c: How will this affect downstream users of cwltool, like Toil, Calrissian, Arvados, and others?

arimu1:
Good question — I checked the three:

- **Toil** and **Arvados** are unaffected: Toil builds its own `ArgumentParser` and Arvados defines its own `arg_parser()`, only borrowing cwltool's default *values* via `get_default_args()`. Neither reuses cwltool's parser object, and `cwltool.main()` skips `arg_parser()` entirely when given a pre-parsed `args`.
- **Calrissian** *would* be affected: it uses `cwltool.argparser.arg_parser()` as its base parser and never sets `prog`, so today it shows `calrissian`; an unconditional `prog="cwltool"` would change that to `cwltool`.

To avoid that, I scoped the override to the actual entry-point name: `prog="cwltool" if os.path.basename(sys.argv[0]) == "cwl-runner" else None`. That still fixes the `cwl-runner` → `cwltool` case from #1535 while leaving Calrissian (and any other downstream reuser) on its own program name. Added a regression test for the downstream case. Pushed.

arimu1:
Checking back in — the pushed change should address the Toil/Calrissian/Arvados question (scoped the override to `cwl-runner` invocations only, with a regression test covering the downstream-reuse case). Let me know if anything else is needed for review.

mr-c (approved):

mr-c:
@arimu1
[ci: retrigger CI (previous run had cancelled jobs)](https://github.com/common-workflow-language/cwltool/pull/2290/commits/abaaecf778237661d51e75ebeeac4efccfec0ac0)

Please don't do that after a maintainer has approved a PR; you should let them handle it from there.

mr-c:
@arimu1 I can re-run just the failed CI; but if you push a new commit then all of the CI has to run again, which is a waste of time and resources

Issue #607, opened 2018-01-12
when `format` is erroneously specified for `type: File` inputs, error out


mr-c:
Example: https://github.com/EOSC-LOFAR/prefactor-cwl/pull/15/commits/c94b7d2fb1b6df64ae2c3acbdce85e2d63627eb1
```
make no-singularity
make run
```
Here execution continues until a confusing message during the first scatter, instead of complaining upfront

PR #2291, opened 2026-06-18
Warn when 'format' is used on a non-File parameter (#1616, #607)

### LINKED ISSUE #1616: cwltool validate must have a warning when format is used with an invalid type (e.g. Directory)
## Expected Behavior

A warning when format is used with an invalid type.

## Actual Behavior

Validation passes with no warnings/errors.

## Workflow Code

https://cwl.discourse.group/t/potential-defect-in-ontology-format-checking/529

## Full Traceback
NA
## Your Environment
* cwltool version: 
3.1.20220204090313

arimu1:
## Summary

Fixes #1616 and #607.

`format` is only meaningful for `File` parameters. cwltool silently accepted it on `Directory` (and other) inputs/outputs, so `--validate` passed with no indication that the field would be ignored — exactly the "validation passes with no warnings/errors" behavior reported in #1616, and the "complains far too late" symptom in #607.

```console
$ cwltool --validate format-on-directory.cwl
format-on-directory.cwl is valid CWL.    # before — no hint that `format` is bogus
```

## Change

In `Process.__init__` (`cwltool/process.py`), while iterating the tool's `inputs`/`outputs`, emit a warning when a parameter declares `format` but its type does not contain a `File`. A small recursive helper `_type_contains_file` handles plain (`File`), namespaced, array (`File[]`), nullable (`File?`) and record forms.

```console
$ cwltool --validate format-on-directory.cwl
format-on-directory.cwl:10:5: 'format' is only valid for 'File' type parameters, but 'indir' is not a File; the 'format' field will be ignored.
format-on-directory.cwl is valid CWL.    # after
```

A **warning** (not a hard error as #607 literally suggested) was chosen so existing documents carrying a harmless stray `format` keep validating; this also matches the direction in the newer #1616. The warning respects `--no-warnings`/`--quiet` like other validation warnings.

## Tests

Added `tests/wf/format-on-directory.cwl` and `tests/wf/format-on-file.cwl` fixtures, plus two tests in `tests/test_validate.py`:
- `test_validate_warns_on_format_for_non_file` — asserts the warning is emitted (captured via a custom log handler, as `test_validate_custom_logger` does).
- `test_validate_no_format_warning_for_file` — asserts `format` on a `File` does **not** warn.

I verified the warn-test fails against the pre-fix code and passes after, and that the File case never warns. Full `test_validate.py` passes (8); `black`/`flake8`/`isort` clean.

---

This change was produced with the assistance of **Claude Code** (model: Claude Opus). The diff, analysis, and tests were reviewed by a human before submission.

### DIFF
--- a/cwltool/process.py
+++ b/cwltool/process.py
@@ -467,6 +467,26 @@ def avroize_type(
     return field_type
 
 
+def _type_contains_file(
+    field_type: CWLObjectType | MutableSequence[Any] | CWLOutputType | None,
+) -> bool:
+    """Return True if a (possibly compound/nullable) parameter type includes a ``File``."""
+    if isinstance(field_type, str):
+        # handle plain ("File"), namespaced ("...#File") and avro ("...cwl.File") forms
+        return field_type.replace("#", ".").split(".")[-1] == "File"
+    if isinstance(field_type, MutableSequence):
+        return any(_type_contains_file(entry) for entry in field_type)
+    if isinstance(field_type, MutableMapping):
+        if field_type.get("type") == "array":
+            return _type_contains_file(cast(CWLOutputType, field_type.get("items")))
+        if field_type.get("type") == "record":
+            return any(
+                _type_contains_file(cast(CWLObjectType, fld).get("type"))
+                for fld in cast(MutableSequence[Any], field_type.get("fields", []))
+            )
+    return False
+
+
 def get_overrides(overrides: MutableSequence[CWLObjectType], toolid: str) -> CWLObjectType:
     """Combine overrides for the target tool ID."""
     req: CWLObjectType = {}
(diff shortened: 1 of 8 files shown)


cwltool/process.py
```diff
@@ -467,6 +467,23 @@ def avroize_type(
...
+    match field_type:
+        case str() as type_str:
+            # handle plain ("File"), namespaced ("...#File") and avro ("...cwl.File") forms
+            return type_str.replace("#", ".").split(".")[-1] == "File"
+        case MutableSequence() as field_seq:
+            return any(_type_contains_file(entry) for entry in field_seq)
+        case {"type": "array", "items": list() as items}:
+            return _type_contains_file(cast(CWLOutputType, items))
+        case {"type": "record", "fields": list() as fields}:
+            return any(_type_contains_file(cast(CWLObjectType, fld).get("type")) for fld in fields)
```
mr-c: These two cases need test coverage

mr-c (changes requested):
Thank you @arimu1 ; please add tests to complete the test coverage

arimu1:
Thanks @mr-c — added coverage for the array and record arms:

- `File[]` / `Directory[]` (array recursion; string `items` after load)
- record with a File field / record with only string fields

Also adjusted the array match to accept string `items` (not only lists). After schema salad load, `type: File[]` is `{"type": "array", "items": "File"}`, so the `list()`-only pattern was incorrectly warning on valid `File[]` + `format`.

`tests/test_validate.py` — 12 passed locally.

mr-c (approved):

PR #2301, opened 2026-07-07
Welling/check tmp dir exists

jswelling:
There appears to be an order-of-operations problem when using the --provenance flag which causes cwltool to attempt to create a scratch directory before the parent directory cwl-tmp has been created.  This mod verifies that the parent directory exists before a scratch directory is created.


jswelling:
The small mod for this PR is getting tangled with the git rebase.  I'll try again.

PR #2304, opened 2026-07-23
Handle list-valued doc when rendering --tool-help

### LINKED ISSUE #2100: cwltool --tool-help fails if `doc` is a list
example_tool.cwl:
```yaml
cwlVersion: v1.2
class: CommandLineTool
$namespaces:
  s: https://schema.org/

label: A short, human-readable label
doc:
- A longer description of the tool or workflow.
- "Ideally with examples, for instance:"
- |
      cwltool --tool-help tool_help_doclist.cwl

inputs:
  message:
    default: hello world
    type: string

outputs:
  main_output: stdout

baseCommand: echo
arguments: [ $(inputs.message) ]

stdout: message.txt
```

Running `cwltool --tool-help example_tool.cwl` (both with version 3.1.20250110105449 and 3.1.20241217163858) prints the following error:
```
ERROR Unhandled error, try again with --debug for more information:
  Unable to get render width for ['A longer description of the tool.', 'Ideally with examples, e.g.:', 'cwltool --tool-help tool_help_doclist.cwl\n']; a str, Segment, or object with __rich_console__ method is required
```

Expected output:
```
usage: example_tool.cwl [-h] [--message MESSAGE] [job_order]

A longer description of the tool or workflow. Ideally with examples,
for instance: cwltool --tool-help example_tool.cwl

positional arguments:
  job_order          Job input json file

options:
  -h, --help         show this help message and exit
  --message MESSAGE
```

Or, (maybe) even better:
```
usage: example_tool.cwl [-h] [--message MESSAGE] [job_order]

A short, human-readable label.

A longer description of the tool or workflow. Ideally with examples,
for instance:

  cwltool --tool-help example_tool.cwl


positional arguments:
  job_order          Job input json file

options:
  -h, --help         show this help message and exit
  --message MESSAGE
```

arimu1:
## Problem

Fixes #2100.

The CWL spec allows `doc` (and `label`) to be either a single string or an array of strings, but the `--tool-help` renderer assumed `doc` was always a string. A tool whose top-level `doc:` is a YAML list crashes:

```console
$ cwltool --tool-help doclist.cwl
ERROR Unhandled error:
  Unable to get render width for ['line one', 'line two']; a str, Segment, or object with __rich_console__ method is required
```

And a list-valued `doc` on an *input* crashes earlier, with a different symptom:

```console
$ cwltool --tool-help doclist.cwl
ERROR Unhandled error, try again with --debug for more information:
  expected string or bytes-like object, got 'CommentedSeq'
```

Both come from the same root cause in `cwltool/argparser.py`:
- `generate_parser` assigns `tool.tool.get("doc", ...)` directly to `toolparser.description`, which `rich_argparse`'s `RichHelpFormatter` then tries to render — a list isn't a valid rich renderable.
- `add_argument` passes the same raw value into `rich.markup.escape(description)`, which requires a `str`.

## Fix

Added a small `doc_to_str` helper that normalizes `doc`/`label` to a string (list entries are joined with newlines), and used it at both call sites. The existing string path is unchanged; `None` (no doc/label) still produces `toolparser.description = None` as before.

## Tests

- `tests/test_toolargparse.py::test_argparser_with_doc_as_list` — `generate_parser` on a tool whose `doc` is a list produces a non-`None` description containing all list entries.
- `tests/test_misc_cli.py::test_tool_help_with_doc_list` — end-to-end `--tool-help` (the actual crash path via `print_help`) exits 0 and prints all doc lines, for both the top-level tool doc and an input's doc.
- New fixture: `tests/with_doc_list.cwl` (list-valued `doc` on the tool and on an input), alongside the existing `tests/with_doc.cwl`.

Both new tests fail against the unpatched code (`TypeError: expected string or bytes-like object, got 'CommentedSeq'` / `rich.errors.NotRenderableError: Unable to get render width for [...]`) and pass with the fix. Full run of `tests/test_toolargparse.py` and `tests/test_misc_cli.py`: 31 passed. `black`, `flake8`, `isort`, and `mypy` are clean on the changed files.

---

This change was produced with the assistance of Claude Code. The diff and tests were reviewed by a human before submission.

### DIFF
--- a/cwltool/argparser.py
+++ b/cwltool/argparser.py
@@ -966,7 +966,7 @@ def add_argument(
     urljoin: Callable[[str, str], str] = urllib.parse.urljoin,
     base_uri: str = "",
 ) -> None:
-    description = rich.markup.escape(doc_to_str(description))
+    description = rich.markup.escape(description)
     if len(name) == 1:
         flag = "-"
     else:
@@ -1075,7 +1075,7 @@ def generate_parser(
         name = shortname(inp["id"])
         namemap[name.replace("-", "_")] = name
         inptype = inp["type"]
-        description = inp.get("doc", inp.get("label", ""))
+        description = doc_to_str(inp.get("doc", inp.get("label", "")))
         default = inp.get("default", None)
         add_argument(
             toolparser,
--- a/cwltool/argparser.py
+++ b/cwltool/argparser.py
@@ -941,6 +941,20 @@ class AppendAction(argparse.Action):
         setattr(namespace, self.dest, g)
 
 
+def doc_to_str(doc: Any) -> str:
+    """Normalize a CWL ``doc``/``label`` field to a single string.
+
+    The CWL spec allows ``doc`` to be either a single string or an array
+    of strings (one entry per line). Join list entries with newlines so
+    downstream consumers (argparse, rich) always receive a plain string.
+    """
+    if isinstance(doc, MutableSequence):
+        return "\n".join(str(entry) for entry in doc)
+    if doc is None:
+        return ""
+    return str(doc)
+
+
 def add_argument(
     toolparser: argparse.ArgumentParser,
     name: str,
@@ -952,7 +966,7 @@ def add_argument(
     urljoin: Callable[[str, str], str] = urllib.parse.urljoin,
     base_uri: str = "",
 ) -> None:
-    description = rich.markup.escape(description)
+    description = rich.markup.escape(doc_to_str(description))
     if len(name) == 1:
         flag = "-"
     else:
@@ -1052,7 +1066,8 @@ def generate_parser(
     base_uri: str = "",
 ) -> argparse.ArgumentParser:
     """Generate an ArgumentParser for the given CWL Process."""
-    toolparser.description = tool.tool.get("doc", tool.tool.get("label", None))
+    doc = tool.tool.get("doc", tool.tool.get("label", None))
+    toolparser.description = doc_to_str(doc) if doc is not None else None
     toolparser.add_argument("job_order", nargs="?", help="Job input json file")
     namemap["job_order"] = "job_order"
 
--- a/cwltool/argparser.py
+++ b/cwltool/argparser.py
@@ -1066,8 +1066,7 @@ def generate_parser(
     base_uri: str = "",
 ) -> argparse.ArgumentParser:
     """Generate an ArgumentParser for the given CWL Process."""
-    doc = tool.tool.get("doc", tool.tool.get("label", None))
-    toolparser.description = doc_to_str(doc) if doc is not None else None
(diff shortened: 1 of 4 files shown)


cwltool/argparser.py
```diff
@@ -941,6 +941,20 @@ def __call__(
...
 
+def doc_to_str(doc: Any) -> str:
+    """Normalize a CWL ``doc``/``label`` field to a single string.
+
+    The CWL spec allows ``doc`` to be either a single string or an array
+    of strings (one entry per line). Join list entries with newlines so
+    downstream consumers (argparse, rich) always receive a plain string.
+    """
+    if isinstance(doc, MutableSequence):
+        return "\n".join(str(entry) for entry in doc)
```
mr-c: Why not `return "\n".join(doc)` for all cases? Works on plain strings, and no need to repack the collection.

mr-c (changes requested):

cwltool/argparser.py
arimu1: Tried it, and it doesn't work on plain strings — `str` is itself an iterable of characters, so `join` splits it:

```pycon
>>> "\n".join("A tool that does things.")
'A\n \nt\no\no\nl\n \nt\nh\na\nt\n \nd\no\ne\ns\n \nt\nh\ni\nn\ng\ns\n.'
```

and it raises on the `None` case that `generate_parser` relies on:

```pycon
>>> "\n".join(None)
TypeError: can only join an iterable
```

So the `isinstance` check is doing real work — it's what keeps the single-string case (the overwhelmingly common one) from being shredded a character per line in `--tool-help`.

The other half of your comment does apply though: the `str(entry)` repack is only needed if a `doc` array can contain non-strings. If you'd rather trust schema-salad to have validated `array<string>` by this point, I'm happy to drop it to a plain `"\n".join(doc)` inside the branch:

```python
if isinstance(doc, MutableSequence):
    return "\n".join(doc)
if doc is None:
    return ""
return str(doc)
```

I left the `str()` in because #2100 was itself a crash from a `doc` that wasn't the shape the code assumed, and `"\n".join([1, 2])` would be a `TypeError` in the same place. Your call — say the word and I'll push it either way.

mr-c: plain `"\n".join(doc)` inside the branch, please. Thanks!

arimu1:
Updated: plain `"\n".join(doc)` inside the `MutableSequence` branch as requested — still guarded so a plain `str` is not character-split, and `None` still becomes `""`.

mr-c (approved):

arimu1:
@mr-c Done — `doc_to_str` now uses plain `"\n".join(doc)` inside the `MutableSequence` branch (see tip). Also rebased onto current `main` so CI re-runs; previous reds looked like podman/singularity runner flakes (crun/python:2-slim), not the doc fix itself.

PR #2325, opened 2026-08-04
Fix resolve_local returning CWD-anchored URI for commonwl share lookups

Dev9269:
## Summary
`resolve_local` searched the XDG_DATA_HOME/commonwl share directories for a tool, found it, but returned `Path(uri).as_uri()` — a URI anchored to the **current working directory** instead of the found file. For relative URIs this raised `ValueError: relative path can't be expressed as a file URI`; for absolute URIs it pointed at the wrong (nonexistent) file. It also silently dropped `#fragment` references, and the raw uri (fragments included) was embedded in the share path so fragment lookups never matched.

## Fix
- Return the URI of the actually-found share file, using `is_file()` instead of `os.path.exists()`.
- Defrag the uri before building share paths, then append `#fragment` to the result like the direct-file branch does.
- Added the missing `.cwl`-extension branch's fragment handling.

## Tests
New `tests/test_resolver.py` with 3 tests (share lookup, fragment preservation, implicit `.cwl` extension). All crash/return-wrong-value on the old code.

### DIFF
--- a/cwltool/resolver.py
+++ b/cwltool/resolver.py
@@ -28,15 +28,21 @@ def resolve_local(document_loader: Loader | None, uri: str) -> str | None:
         os.environ.get("XDG_DATA_HOME", os.path.join(os.path.expanduser("~"), ".local", "share"))
     ]
     sharepaths.extend(os.environ.get("XDG_DATA_DIRS", "/usr/local/share/:/usr/share/").split(":"))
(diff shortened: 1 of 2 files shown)


mr-c:
Thank you @Dev9269 for this PR; can you share a bit about how you are using CWL? 

I'm glad for this feature to get attention, but I think you are the first to try it :-P

PR #2338, opened 2026-08-19
Warn when a baseCommand cannot be found on PATH or in a container image

### LINKED ISSUE #2240: `--validate` should check for `baseCommand` presence on PATH or in containers
would catch `baseCommand` that contains a space, instead of comma separated list

- [ ] 1. throw a warning about a space
- [ ] 2. actually check PATH or containers

kyoai-zhao:
Closes #2240 (second checklist item).

During `--validate`, emit a warning (never an error) when a literal `baseCommand` cannot be resolved on the PATH, or cannot be found inside the declared container image when that image is already available locally.

Design points:
- Lists take the first element (the program); strings with a `/` (paths) and expressions are skipped.
- Container probing only happens when the image is already available locally (`docker image inspect`), so `--validate` never triggers a large image pull; when the image is not local the check is skipped with a DEBUG message.
- Images without a shell (e.g. distroless) are distinguished from images that genuinely lack the command, and are skipped rather than misreported.
- Supports both `docker` and `podman`.

Tests: 6 new cases in `tests/test_validate.py` (PATH missing, list-form missing, present command silent, absolute path silent, container probe unit tests, container-mode warning), new CWL fixtures under `tests/wf/2240-*`. `tests/test_validate.py` and `tests/test_examples.py` pass locally; flake8/black/isort clean.

### DIFF
--- a/cwltool/command_line_tool.py
+++ b/cwltool/command_line_tool.py
@@ -9,7 +9,7 @@ import os
 import re
 import shlex
 import shutil
-import subprocess
+import subprocess  # nosec
 import threading
 import urllib
 import urllib.parse
@@ -422,7 +422,7 @@ def default_make_path_mapper(
 def _container_has_command(docker: str, image: str, command: str) -> Optional[bool]:
     """Return True/False if *command* is (not) present in *image*, or None if
     the image is not available locally and cannot be inspected."""
-    inspect = subprocess.run(
+    inspect = subprocess.run(  # nosec
         [docker, "image", "inspect", image],
         capture_output=True,
         text=True,
@@ -430,7 +430,7 @@ def _container_has_command(docker: str, image: str, command: str) -> Optional[bo
     )
     if inspect.returncode != 0:
         return None
-    probe = subprocess.run(
+    probe = subprocess.run(  # nosec
         [docker, "run", "--rm", "--entrypoint", "sh", image, "-c", shlex.quote(command)],
         capture_output=True,
         text=True,
@@ -440,7 +440,7 @@ def _container_has_command(docker: str, image: str, command: str) -> Optional[bo
         return True
     # The probe can also fail when the image has no shell at all (e.g.
     # distroless images); distinguish that from a missing command.
-    shell = subprocess.run(
+    shell = subprocess.run(  # nosec
         [docker, "run", "--rm", "--entrypoint", "sh", image, "true"],
         capture_output=True,
         text=True,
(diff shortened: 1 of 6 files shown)


mr-c:
Hello @kyoai-zhao and thank you for your PR; can you tell me more about your experience with CWL?

Is this a LLM or coding-agent assisted contribution?

kyoai-zhao:
Thanks for the quick response! Happy to answer both questions directly.

**On CWL experience:** I have used CWL (writing and running workflows), but this is my first contribution to the CWL ecosystem — no prior contributions to cwltool or other CWL tooling.

**On LLM/coding-agent assistance:** Yes — this PR was made with the help of a coding agent running in my local workspace, in the following workflow: I selected the issue; the agent verified it was reproducible on current main and that no competing PR existed, then implemented it; I reviewed the diff and ran the full relevant test suite and lint locally before submitting. The approach decisions (warning rather than error, never pulling images, only probing images that are already available locally) were made jointly and reviewed by me.

Happy to follow any policy your project prefers for agent-assisted contributions — for example flagging the contribution as AI-assisted, adjusting scope, or amending the implementation — just let me know.

kyoai-zhao:
Pushed a21cb30 fixing the two failing checks:
- **bandit**: added `# nosec` markers on the new `import subprocess` and the three `subprocess.run` calls, following the existing convention in docker.py/job.py/utils.py (Low-severity B404/B603); local `bandit -r cwltool` is clean.
- **mypy (3.11)**: errors were all in the new test code (missing annotations on the mock helpers under `strict = True`, implicit re-exports `clt.subprocess`/`clt.shutil`/`clt._container_has_command`). Fixed with explicit `-> object` annotations and `getattr` access; local `make mypy` equivalent is clean except the pre-existing local-stub issue in test_http_input.py (missing httpretty stub in my env, unrelated — CI has the full dependency set).

`build_test_container` still fails, but I cannot see its logs from the fork (no admin access). It builds the module container and runs the full suite inside it; locally the full suite passes (546 passed / 159 skipped; remaining errors are env-only: node.js sandbox, httpretty stub). The unit tox jobs for all Python versions are green, so this looks like a container-build/infra issue rather than a code issue — could you re-run it?

mr-c:
@kyoai-zhao Sorry, we can't accept LLM/agent contributions as they are not redistributable under our license. 

And personally I find them very verbose, and not a good use of my time to review.

If you, a human, would like to make contributions yourself; then I would be happy to review (and mentor you, as needed).

kyoai-zhao:
@mr-c Thanks for your reply. I will try to do some contributions without code-agent.

